PR #20, opened 2022-10-08
[Meta]: Add CodeQL scanning

ad3154:
Adding security scanning.
Not sure how this will work, so creating a PR to allow the workflow to run.


ad3154:
Neat! This actually found something! I will push a commit to clean up the 1 issue it found in ETP

PR #48, opened 2022-11-12
[Meta]: Add linting checks to CI

ad3154:
## Added clang-format verification to all PRs

Adding these checks to CI will prevent merging code to `main` with non-compliant formatting and will provide quick feedback on your PRs regarding their compliance with the repo's style.

You can easily format your code to this repo's standard by running clang-format locally before committing.

Make sure you have clang-format installed:

```
sudo apt install clang-format
```

Then run this command to ensure the code is linted.

```
find . -iname *.hpp -o -iname *.cpp | xargs clang-format -i
```


ad3154:
Looks like doing it for all pushes causes fun duplicate jobs to run. Maybe it makes more sense to only run on PRs. I think I will edit this to run on PR only...

PR #56, opened 2022-11-15
CMake install fixes

### LINKED ISSUE #55: CMake install into locall install directory fails because prefix path is not obeyed
I'm trying to install the library not in system headers, but a local folder like so:
```
$ cmake -S . -B build && cmake --build build && cmake --install build --prefix install
CMake Warning (dev) at hardware_integration/CMakeLists.txt:14 (message):
  No CAN driver specified, choosing Linux socket CAN by default.  Use
  -DCAN_DRIVER=socketCAN to avoid this, or choose another driver and set it
  explicitly.
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Configuring done
-- Generating done
...
CMake Error at build/isobus/cmake_install.cmake:77 (file):
  file cannot create directory: /usr/local/ISOBUS/include.  Maybe need
  administrative privileges.
Call Stack (most recent call first):
  build/cmake_install.cmake:52 (include)
  ```
  
  It should not be installing anything in `/usr/local/ISOBUS/include`; the [prefix](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake--install-0) is not obeyed.
  
  Installing locally first before polluting system headers is a good idea when developing on a host OS.

ad3154:
* Fix the supporting libraries were not in any export set
* Fixed install prefix was being overwritten for isobus library
* Moved cpack directives up to the root so that the supporting libraries also get included in the deb

@RFRIEDM-Trimble Can you review this for me to ensure it will resolve your issue?

Closes #55


RFRIEDM-Trimble:
Approved! The install is successful. There may be more requirements that come up to support `find_package` as feature additions.

Issue #61, opened 2022-12-03
Compiling the Isobus lib as a static lib causes some protocols to be optimized out


ad3154:
This is an unintended side effect of #58. Because `TransportProtocolManager` and `ExtendedTransportProtocolManager` are not directly called anywhere (only indirectly through the list of all protocols) the linker optimizes out those classes and their static singletons do not ever get instantiated.

ad3154:
It looks like one viable solution to this is to add `-Wl,--whole-archive` to the `target_link_libraries` step for _only the isobus::Isobus_ library.  I will perform some testing with this and if it is viable as a solution I will add it to all tutorials, readme, and examples. Compiling and linking will still work without it, but it will omit the transport layer completely without it, limiting your app to 8 byte messages.

PR #62, opened 2022-12-03
[#61]: Help prevent TP and ETP from being optimized out when using static libs

ad3154:
Also added `PRIVATE` to all example `target_link_libraries` calls

Basically here's what's happening... in all the examples when using static libraries the `TransportProtocolManager` and `ExtendedTransportProtocolManager` were not getting linked in from the library into the executable. This is because linkers do not need to do that if they think that translation unit is not referenced anywhere in the final executable. This was causing all transport layer functions to fail, as the transport layer relies on those static objects being instantiated. Using shared libraries fixes this, but because static is the default now, I've added some logic to CMake to force the whole isobus::Isobus archive to be linked.

@RFRIEDM-Trimble I would be interested in your opinion on this if you have one, as it feels a bit strange. This fix seems to work but I am not sure if there's a better way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,10 +123,12 @@ find_package(Threads)
 
 add_subdirectory(<path to this submodule>)
 
-target_link_libraries(<your executable name> isobus::Isobus isobus::HardwareIntegration isobus::SocketCANInterface)
+target_link_libraries(<your executable name> PRIVATE -Wl,--whole-archive isobus::Isobus -Wl,--no-whole-archive isobus::HardwareIntegration isobus::SocketCANInterface)
 ```
 
 A full example CMakeLists.txt file can be found on the tutorial website.
+
+`-Wl,--whole-archive` is required to ensure that some static singleton objects, such as the `TransportProtocolManager` don't get optimized out of your executable by your linker.
 
 ## Documentation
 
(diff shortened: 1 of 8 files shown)


RFRIEDM-Trimble:
> Also added `PRIVATE` to all example `target_link_libraries` calls
> 
> Basically here's what's happening... in all the examples when using static libraries the `TransportProtocolManager` and `ExtendedTransportProtocolManager` were not getting linked in from the library into the executable. This is because linkers do not need to do that if they think that translation unit is not referenced anywhere in the final executable. This was causing all transport layer functions to fail, as the transport layer relies on those static objects being instantiated. Using shared libraries fixes this, but because static is the default now, I've added some logic to CMake to force the whole isobus::Isobus archive to be linked.
> 
> @RFRIEDM-Trimble I would be interested in your opinion on this if you have one, as it feels a bit strange. This fix seems to work but I am not sure if there's a better way.

Sorry, I don't have much experience with using Singletons in shared libs. I did take a look at the usage of `whole-archive` and believe this reference could provide a better option rather than bloating targets with everything. 
https://stackoverflow.com/questions/47186406/is-there-a-way-to-force-c-compiler-to-not-optimize-out-specific-static-objects

PR #71, opened 2022-12-12
[Core]: Fix SocketCAN interface unwanted behaviour

GwnDaan:
SocketCAN interface now closes correctly when interface is down. It also logs a warning for this to CANStackLogger. And resolved an issue where the hardware interface keeps reopening the CANHardwarePlugin.

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,9 @@
+version = 1
+
+[[analyzers]]
+name = "shell"
+enabled = true
+
+[[analyzers]]
+name = "cxx"
+enabled = true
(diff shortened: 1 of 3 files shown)


ad3154 (dismissed):

ad3154:
Looks good, just needs the linter to tidy it up.
If you have clang-format installed (`sudo apt install clang-format`) then you can have it auto-format the files to pass this check by running this command from the root of the repo: `find . -iname *.hpp -o -iname *.cpp | xargs clang-format -i`

ad3154 (approved):

PR #76, opened 2022-12-15
[#75] Stop buffering messages when hardware layer is in an invalid state

### LINKED ISSUE #75: N2K FP should not be successfully sending messages with interface down
image

I would've liked to find the root cause of this issue myself, but I find it a bit tedious to put debug logs everywhere till I find the root of the issue. Surely there has to be a better way to debug this by being able to use breakpoints or something?

ad3154:
Basically the stack was always buffering messages if the specified CAN interface existed and the hardware interface threads were started. This means that address claiming could "succeed" even if the interface was down. Now we'll return false for transmits instead if the underlying hardware is not connected. This, along with an internal control function validity check in fast packet, should close issue 75 and prevent some unwanted behavior where we would spam the bus if you disconnected physically your CAN adapter from the bus, then plug it back in.

Closes #75

### DIFF
--- a/hardware_integration/src/can_hardware_interface.cpp
+++ b/hardware_integration/src/can_hardware_interface.cpp
@@ -8,7 +8,9 @@
 /// @copyright 2022 Adrian Del Grosso
 //================================================================================================
 #include "isobus/hardware_integration/can_hardware_interface.hpp"
+#include <isobus/isobus/can_warning_logger.hpp>
 #include "isobus/utility/system_timing.hpp"
+#include "isobus/utility/to_string.hpp"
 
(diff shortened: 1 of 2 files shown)


ad3154:
Slight amendment, we may still spam a bit if all you do is unplug and replug the interface, because the socket would still be in a valid state (and there's not a super great way to check if the socket buffer is full) but we at least won't falsely succeed anymore if the interface is truly in the down state. Will have to investigate further if we can detect BUS_OFF and a full buffer, but that can be out of scope for this PR.

ad3154:
Testing results: 
image

PR #82, opened 2022-12-19
[VT] Add remaining activation message callbacks

### LINKED ISSUE #81: Add remaining activation messages
I saw that only the first 4 activation messages were implemented right now (ButtonActivation, SoftKeyActivation, PointingEvent and SelectInputObject). Since there are 11 in total at this time, that means that there are still 7 missing. For the sake of convenience, do we want to implement these as well? I don't mind spending some time on this myself to get this done.

GwnDaan:
Adds support for the 7 VT activation messages in the form of callbacks which were still missing.

Closes #81

### DIFF
--- a/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
+++ b/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
@@ -162,6 +162,13 @@ namespace isobus
 			ButtonPressAborted = 3 ///< Press was aborted (user navigated away from the button and did not release it)
 		};
 
+		enum class ESCMessageErrorCode : std::uint8_t
+		{
+			NoError = 0, ///< No error occurred
+			NoInputFieldOpen = 1, ///< No input field is open
+			OtherError = 5, ///< Error is not one of the above
+		};
+
 		/// @brief The internal state machine state of the VT client
 		enum class StateMachineState : std::uint8_t
 		{
@@ -256,9 +263,40 @@ namespace isobus
 		/// @brief A typedef for a generic key event for convenience
 		typedef void (*VTKeyEventCallback)(KeyActivationCode keyEvent, std::uint8_t keyNumber, std::uint16_t objectID, std::uint16_t parentObjectID, VirtualTerminalClient *parentPointer);
 		/// @brief A typedef for a generic pointing event, for convenience
-		typedef void (*VTPointingEventCallback)(KeyActivationCode keyEvent, std::uint16_t xPosition, std::uint16_t yPosition, VirtualTerminalClient *parentPointer);
+		typedef void (*VTPointingEventCallback)(KeyActivationCode keyEvent,
+		                                        std::uint16_t xPosition,
+		                                        std::uint16_t yPosition,
+		                                        std::uint16_t parentMaskObjectID,
+		                                        VirtualTerminalClient *parentPointer);
 		/// @brief A typedef for a generic VT input object selection callback for convenience
 		typedef void (*VTSelectInputObjectCallback)(std::uint16_t objectID, bool objectSelected, bool objectOpenForInput, VirtualTerminalClient *parentPointer);
+		/// @brief A typedef for a generic VT ESC message callback for convenience
+		typedef void (*VTESCMessageCallback)(std::uint16_t objectID, ESCMessageErrorCode errorCode, VirtualTerminalClient *parentPointer);
+		/// @brief A typedef for a generic VT change numeric value callback for convenience
+		typedef void (*VTChangeNumericValueCallback)(std::uint16_t objectID, std::uint32_t value, VirtualTerminalClient *parentPointer);
+		/// @brief A typedef for a generic VT change active mask callback for convenience
+		typedef void (*VTChangeActiveMaskCallback)(std::uint16_t maskObjectID,
+		                                           std::uint16_t errorObjectID,
+		                                           std::uint16_t parentObjectID,
+		                                           bool missingObjects,
+		                                           bool maskOrChildHasErrors,
+		                                           bool anyOtherEror,
+		                                           bool poolDeleted,
+		                                           VirtualTerminalClient *parentPointer);
+		/// @brief A typedef for a generic VT change soft key mask callback for convenience
+		typedef void (*VTChangeSoftKeyMaskCallback)(std::uint16_t dataOrAlarmMaskObjectID,
+		                                            std::uint16_t softKeyMaskObjectID,
+		                                            bool missingObjects,
+		                                            bool maskOrChildHasErrors,
+		                                            bool anyOtherEror,
+		                                            bool poolDeleted,
+		                                            VirtualTerminalClient *parentPointer);
+		/// @brief A typedef for a generic VT change string value callback for convenience
+		typedef void (*VTChangeStringValueCallback)(std::uint16_t objectID, std::string value, VirtualTerminalClient *parentPointer);
+		/// @brief A typedef for a generic VT on user-layout hide/show callback for convenience
+		typedef void (*VTUserLayoutHideShowCallback)(std::uint16_t objectID, bool isHidden, VirtualTerminalClient *parentPointer);
+		/// @brief A typedef for a generic VT control audio signal termination callback for convenience
+		typedef void (*VTAudioSignalTerminationCallback)(bool isTerminated, VirtualTerminalClient *parentPointer);
 
 		// Callbacks for events that happen on the VT
 		/// @brief Allows you to register for a callback when a softkey is pressed or released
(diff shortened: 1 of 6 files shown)


isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
```diff
@@ -293,6 +331,65 @@ namespace isobus
...
+		/// @param[in] value The callback to remove
+		void RemoveVTUserLayoutHideShowEventCallback(VTUserLayoutHideShowCallback value);
+
+		/// @brief Allows you to register for a callback when an audio signal is terminated
+		/// @param[in] value The callback to register
+		void RegisterVTControlAudioSignalTerminationEventCallback(VTAudioSignalTerminationCallback value);
+
+		/// @brief Allows you to remove a callback when an audio signal is terminated
+		/// @param[in] value The callback to remove
+		void RemoveVTControlAudioSignalTerminationEventCallback(VTAudioSignalTerminationCallback value);
```
GwnDaan, line 391: Wondering if UpperCamelCase naming is desired for these? I followed the style of the ones that were already present

isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
```diff
@@ -162,6 +162,13 @@ namespace isobus
 			ButtonPressAborted = 3 ///< Press was aborted (user navigated away from the button and did not release it)
 		};
 
+		enum class ESCMessageErrorCode : std::uint8_t
+		{
+			NoError = 0, ///< No error occurred
+			NoInputFieldOpen = 1, ///< No input field is open
+			OtherError = 5, ///< Error is not one of the above
```
ad3154, line 169: Nit: Pedantic compilers don't like trailing commas, I suggest removing this one

isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
ad3154: I've been trying (and apparently not always succeeding) to do `snake_case` for functions. I should really set up clang-tidy checking for this at some point. So I suppose in my perfect world they'd be `snake_case` kind of like other places, like `get_connected_vt_version` etc

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -169,6 +169,111 @@ namespace isobus
...
 	}
 
+	void VirtualTerminalClient::RegisterVTESCMessageEventCallback(VTESCMessageCallback value)
+	{
+		escMessageCallbacks.push_back(value);
+	}
+
+	void VirtualTerminalClient::RemoveVTESCMessageEventCallback(VTESCMessageCallback value)
+	{
+		auto callbackLocation = find(escMessageCallbacks.begin(), escMessageCallbacks.end(), value);
```
ad3154, line 179: Nit: `find` is in `std::` namespace I think, so ideally we should prefix with `std::` in these functions

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -1323,6 +1428,18 @@ namespace isobus
 		return retVal;
 	}
 
+	bool VirtualTerminalClient::get_vt_version_supported(VTVersion minimumVersion)
+	{
+		bool retVal = false;
+
+		if (get_connected_vt_version() != VTVersion::ReservedOrUnknown)
+		{
+			retVal = get_connected_vt_version() >= minimumVersion;
```
ad3154, line 1437: `get_connected_vt_version` is `const` so this function can probably also be `const`. 

`bool VirtualTerminalClient::get_vt_version_supported(VTVersion minimumVersion) const` here and in the header file.

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -2109,6 +2230,109 @@ namespace isobus
 		}
 	}
 
+	void VirtualTerminalClient::process_esc_message_callback(std::uint16_t objectID, ESCMessageErrorCode errorCode, VirtualTerminalClient *parentPointer)
+	{
+		for (uint32_t i = 0; i < parentPointer->escMessageCallbacks.size(); i++)
```
ad3154, line 2235: Nit: `size()` returns `size_t` so we should probably use `std::size_t` in these rather than `uint32_t`. Prefer to have `std::` explicitly

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -2155,16 +2379,17 @@ namespace isobus
 		    (CAN_DATA_LENGTH == message->get_data_length()))
 		{
 			VirtualTerminalClient *parentVT = reinterpret_cast<VirtualTerminalClient *>(parentPointer);
+			auto &data = message->get_data();
```
ad3154, line 2382: I've recently learned that `auto` doesn't deduce references, so this is actually copying the data. I suppose that means we should explicitly say `std::vector<std::uint8_t> &` rather than `auto &`

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -2178,98 +2403,251 @@ namespace isobus
...
+						{
+							bool terminated = (0x01 == (data[1] & 0x01));
+
+							parentVT->process_audio_signal_termination_callback(terminated, parentVT);
+
+							if (parentVT->get_vt_version_supported(VTVersion::Version6))
+							{
+								//! @todo process TAN
+							}
+						}
```
ad3154, line 2629: Missing `break;`

ad3154 (changes requested):

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -2178,98 +2403,253 @@ namespace isobus
...
+
+						case static_cast<std::uint8_t>(Function::VTChangeActiveMaskMessage):
+						{
+							std::uint16_t maskObjectID = (static_cast<std::uint16_t>(data[1]) &
+							                              ((static_cast<std::uint16_t>(data[2])) << 8));
+							std::uint16_t errorObjectID = (static_cast<std::uint16_t>(data[4]) &
+							                               ((static_cast<std::uint16_t>(data[5])) << 8));
+							bool missingObjects = (0x04 == (data[3] & 0x04));
+							bool maskOrChildHasErrors = (0x02 == (data[3] & 0x08));
+							bool anyOtherError = (0x01 == (data[3] & 0x010));
```
ad3154, line 2546: `(0x01 == (data[3] & 0x010)` would never be true, right? You'd have to check `(0 != (data[3] & 0x010))` or `(0x10 == (data[3] & 0x010))`. Similarly, `bool maskOrChildHasErrors = (0x02 == (data[3] & 0x08));` would always be false, because bit 4 (0x08) will never have a value of 2, so maybe double check your bit masking in some of these messages.

ad3154 (changes requested):
I think it's close now, I just had one comment about some of the bit masking.

isobus/src/isobus_virtual_terminal_client.cpp
GwnDaan: Oh dang that doesn't seem right, not sure how I missed those. I'll double check them all

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -2312,15 +2692,15 @@ namespace isobus
...
+								if (data[2] <= static_cast<std::uint8_t>(GraphicMode::TwoHundredFiftySixColor))
 								{
-									parentVT->supportedGraphicsMode = static_cast<GraphicMode>(message->get_data()[2]);
+									parentVT->supportedGraphicsMode = static_cast<GraphicMode>(data[2]);
 								}
-								parentVT->hardwareFeaturesBitfield = message->get_data()[3];
-								parentVT->xPixels = (static_cast<std::uint16_t>(message->get_data()[4]) &
-								                     ((static_cast<std::uint16_t>(message->get_data()[5])) << 8));
-								parentVT->yPixels = (static_cast<std::uint16_t>(message->get_data()[6]) &
-								                     ((static_cast<std::uint16_t>(message->get_data()[7])) << 8));
```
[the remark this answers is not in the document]
ad3154: It should be the bitwise OR operator `|`, correct

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -2178,98 +2403,251 @@ namespace isobus
...
+								{
+									//! @todo process TAN
+								}
+
 								parentVT->process_softkey_event_callback(static_cast<KeyActivationCode>(keyCode),
-								                                         static_cast<std::uint16_t>(message->get_data().at(6)),
-								                                         (static_cast<std::uint16_t>(message->get_data().at(2)) | static_cast<std::uint16_t>(message->get_data().at(3) << 8)),
-								                                         (static_cast<std::uint16_t>(message->get_data().at(4)) | static_cast<std::uint16_t>(message->get_data().at(5) << 8)),
+								                                         static_cast<std::uint16_t>(data.at(6)),
+								                                         (static_cast<std::uint16_t>(data.at(2)) | static_cast<std::uint16_t>(data.at(3) << 8)),
```
ad3154, line 2420: One last note, when we cast to `std::uint16_t` like we do here, the cast should happen before the bit shift to avoid shifting all the bits out of the `std::uin8t_t`. So, it's more correct to do `(static_cast<std::uint16_t>(data.at(3)) << 8)` instead of  ` static_cast<std::uint16_t>(data.at(3) << 8)` Check out the case for `Function::VTStatusMessage` for an example

ad3154 (approved):

PR #91, opened 2022-12-26
PlatformIO support, TWAI driver and MCP2515 driver

GwnDaan:
I made this a draft PR as I'm not sure if all of this should be implemented in the stack. Most of the things below were needed for the project I'm working on, will post an update soon.

What it includes right now:
- `library.json` file for PlatformIO projects
- Two-Wire Automotive Interface (TWAI) driver. Mostly used onboard Espressif boards.
- MCP2515 CAN controller interface/driver. It uses a generic SPI callback so it can be integrated with almost any board, though it is not extensive yet.
- Made SocketCAN driver use check a predefined `__linux__` macro, so that it doesn't give compile errors on other platforms without the need for filtering the src file out.
- Increased the `CANLIB_UPDATE_RATE` from 4ms to 10ms. See https://github.com/ad3154/ISO11783-CAN-Stack/commit/398db3e55a2fd6076db5d08a82a6581a28247c99 for details. This is probably an unwanted change, but at the same time solves unwanted behavior on some platforms.

Any thoughts are welcome, I'm still trying to get the hang of c/c++ 😄


tomstix:
I was actually about to implement an ESP32 driver myself when I just saw this. Great work.
I think it would be useful to check the twai driver funtions for their return codes and log it to CanStackLogger if something goes wrong.

hardware_integration/src/mcp2515_can_interface.cpp
```diff
@@ -0,0 +1,344 @@
...
+			}
+
+			uint8_t len = (rxData[5] & 0x0F);
+			if (len <= CAN_MAX_DLEN)
+			{
+				retVal = true;
+
+				if (read_register(rxb->data, rxData, 8) &&
+				    modify_register(MCPRegister::CANINTF, rxb->intf, 0) &&
+				    0 == (id & CAN_ERR_FLAG))
```
ad3154, line 269: Prefer another set of parenthesis here `(0 ==`

hardware_integration/src/mcp2515_can_interface.cpp
```diff
@@ -0,0 +1,344 @@
...
+						canFrame.isExtendedFrame = true;
+					}
+					else
+					{
+						canFrame.identifier = (id & CAN_SFF_MASK);
+						canFrame.isExtendedFrame = false;
+					}
+					canFrame.dataLength = len;
+					memset(canFrame.data, 0, sizeof(canFrame.data));
+					memcpy(canFrame.data, rxData.data(), canFrame.dataLength);
```
ad3154, line 283: Might want to sanity check `len` or `canFrame.dataLength`  is not more than 8, as it is masked against `0x0F` which could be up to 15.

hardware_integration/src/mcp2515_can_interface.cpp
```diff
@@ -0,0 +1,344 @@
...
+bool MCP2515CANInterface::write_frame(const isobus::HardwareInterfaceCANFrame &canFrame)
+{
+	// The register addresses for the TX buffers
+	constexpr static std::array<TXBRegister, NUM_WRITE_BUFFERS> TXB = { { { MCPRegister::TXB0CTRL, MCPRegister::TXB0SIDH, MCPRegister::TXB0DATA },
+		                                                                    { MCPRegister::TXB1CTRL, MCPRegister::TXB1SIDH, MCPRegister::TXB1DATA },
+		                                                                    { MCPRegister::TXB2CTRL, MCPRegister::TXB2SIDH, MCPRegister::TXB2DATA } } };
+
+	bool retVal = false;
+
+	for (int i = 0; i < NUM_WRITE_BUFFERS; i++)
```
ad3154, line 305: Prefer not to use `int` if `NUM_WRITE_BUFFERS` is well defined, should use a fixed-width type

hardware_integration/src/mcp2515_can_interface.cpp
```diff
@@ -0,0 +1,344 @@
...
+	{
+		const struct TXBRegister *txbuf = &TXB[i];
+		uint8_t ctrl;
```
ad3154, line 308: `std::`

ad3154 (commented):
This is excellent - adding new hardware integrations will be key to broader, easy adoption, and ESP32 and other MCP2515 devices are great target platforms. I made a few comments, mostly style, but we should figure and finalize how we'll select these plugins. I propose in my review that we use CMake to include or not include desired plugins from compilation, so that we can avoid checking preprocessor definitions. This would be a bit closer to the industry standard, as some documents such as Autosar forbid the preprocessor from these kinds of uses. I will try and write up a full suggestion

hardware_integration/include/isobus/hardware_integration/socket_can_interface.hpp
```diff
@@ -9,6 +9,7 @@
 //================================================================================================
 #ifndef SOCKET_CAN_INTERFACE_HPP
 #define SOCKET_CAN_INTERFACE_HPP
+#ifdef __linux__
```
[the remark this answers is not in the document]
tomstix: I also think we should use CMake for driver selection. However, using CMake with microcontrollers is often a hassle.
So a good approach would be to use PlatformioIO to handle source file selection as @GwnDaan already did before [this](https://github.com/GwnDaan/ISO11783-CAN-Stack/commit/1c880ec35deefef9fae7463dca57d837ce357ec3) commit.
I think selecting the source files in the library.json should be the preferred way and would make the #ifdef __linux__ obsolete. For handling other PlatformIO platforms, either ifdefs are needed or we could handle this in a Python script with the [extraScript](https://docs.platformio.org/en/latest/manifests/library-json/fields/build/extrascript.html) option.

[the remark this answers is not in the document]
GwnDaan: Yep, CMake with ESP-IDF does some unwanted behavior sometimes. For example pthread can't be found using `find_package(Threads REQUIRED)` in the main CMakeLists.txt while ptread certainly exists in esp-idf. I agree that selecting the source files in library.json is probably better. I'll change it back.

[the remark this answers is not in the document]
GwnDaan: > I think ideally we'd just select the different driver via CMake so only the needed one(s) get compiled and to avoid checking macro definitions. 

@ad3154 I attempted to implement this in 6c86f037c15ec1214b3445a039281e8c7f8138ba. Is it somewhat like what you had in mind?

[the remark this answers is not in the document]
ad3154: Exactly!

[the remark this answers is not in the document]
ad3154: Oh also, for what it's worth, I was able to compile for ESP32 by changing `find_package(Threads REQUIRED)` to just `find_package(Threads)`. The check then technically fails, but since ESP32 has threads and just doesn't implement this find_package step properly, it still works OK. Perhaps we should change it?

hardware_integration/src/mcp2515_can_interface.cpp
GwnDaan: Pretty sure this sanity check already exists on line 263? Where `CAN_MAX_DLEN` is 8 and will be replaced with `isobus::CAN_DATA_LENGTH`. I'll move `canFrame.dataLength = len;` up to right after that check so it's more clear.

hardware_integration/include/isobus/hardware_integration/socket_can_interface.hpp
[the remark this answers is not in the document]
GwnDaan: We could also add a `if(NOT ESP_PLATFORM)` statement before `find_package(Threads REQUIRED)` to skip it completely? This way building for platforms without threads still fail successfully, and not confuse new users building for ESP thinking they don't have threading support.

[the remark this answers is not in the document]
ad3154: OK, that also sounds good to me!

hardware_integration/include/isobus/hardware_integration/mcp2515_can_interface.hpp
```diff
@@ -0,0 +1,195 @@
...
+#include <thread>
+#include <vector>
+
+#include "isobus/hardware_integration/can_hardware_plugin.hpp"
+#include "isobus/isobus/can_frame.hpp"
+#include "isobus/isobus/can_hardware_abstraction.hpp"
+
+/* special address description flags for the CAN_ID */
+#define CAN_EFF_FLAG 0x80000000UL /* EFF/SFF is set in the MSB */
+#define CAN_RTR_FLAG 0x40000000UL /* remote transmission request */
```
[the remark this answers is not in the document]
GwnDaan: Sure I changed it

hardware_integration/include/isobus/hardware_integration/mcp2515_can_interface.hpp
```diff
@@ -0,0 +1,195 @@
...
+#define CAN_ERR_FLAG 0x20000000UL /* error message frame */
+
+/* valid bits in CAN ID for frame formats */
+#define CAN_SFF_MASK 0x000007FFUL /* standard frame format (SFF) */
+#define CAN_EFF_MASK 0x1FFFFFFFUL /* extended frame format (EFF) */
+#define CAN_ERR_MASK 0x1FFFFFFFUL /* omit EFF, RTR, ERR flags */
+
+/* CAN payload length and DLC definitions according to ISO 11898-1 */
+#define CAN_MAX_DLEN 8
```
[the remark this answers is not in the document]
GwnDaan: Correct, I removed it

GwnDaan:
Added SPITransactionPlugin as a base class for the communication between a CAN driver and SPI. It's used in the MCP2515 interface now instead of the callback. I think this is more neat and allows for better expansion opportunities in the future.

hardware_integration/include/isobus/hardware_integration/can_hardware_interface.hpp
```diff
@@ -145,7 +145,7 @@ class CANHardwareInterface
 	};
 
 	/// @brief A hard-coded update interval for the CAN stack. Mostly arbitrary
-	static const std::uint32_t CANLIB_UPDATE_RATE = 4;
+	static const std::uint32_t CANLIB_UPDATE_RATE = 10;
```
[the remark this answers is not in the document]
GwnDaan: Added `set_can_driver_update_period(std::uint32_t value)` function to `CANHardwareInterface` since building with PlatformIO kind of skips the stack's CMake files. So I think a setter function is preferred aswell.

ad3154:
This PR is still in a draft state - are you still working on it, or do you think it is ready to rebase and re-review?

GwnDaan:
I'm still working on this. The mcp2515 driver together with it's spi communication is not yet finished. I'm debugging weird unwanted behaviour with it and almost reached a point where I'd be blaming the hardware haha. I've refactored the abstract spi communication class a few times now compared to HEAD and not yet sure what would be the optimal one. I hope to get the PR finished within a couple of days

GwnDaan:
It should be ready now. I'll do some more real world testing later tonight to find any missed bugs.

hardware_integration/src/mcp2515_can_interface.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+	return retVal;
+}
+
+bool MCP2515CANInterface::write_frame(const isobus::HardwareInterfaceCANFrame &canFrame)
+{
+	bool retVal = false;
+
+	// Try to write the frame to the first buffer
+	retVal = write_frame(canFrame, MCPRegister::TXB0CTRL, MCPRegister::TXB0SIDH);
```
ad3154, line 370: Super minor thing, you're assigning false to `retVal` before immediately overwriting it, so you can probably simplify it down to:  `bool retVal = write_frame(canFrame, MCPRegister::TXB0CTRL, MCPRegister::TXB0SIDH);`

hardware_integration/src/mcp2515_can_interface.cpp
```diff
@@ -0,0 +1,383 @@
...
+	return retVal;
+}
+
+bool MCP2515CANInterface::get_read_status(std::uint8_t &status)
+{
+	bool retVal = false;
+	if (transactionHandler)
+	{
+		SPIHardwarePlugin::TransactionFrame frame{
+			.txBuffer = { static_cast<std::uint8_t>(MCPInstruction::READ_STATUS), 0x00 },
```
ad3154, line 85: GCC complains about designated initializers here and elsewhere `C++ designated initializers only available with ‘-std=c++2a’ or ‘-std=gnu++2a’`

Perhaps it might make more sense to add a constructor to the struct or make it a class? Either one of those is maybe more portable.

hardware_integration/src/twai_plugin.cpp
```diff
@@ -0,0 +1,129 @@
...
+}
+
+bool TWAIPlugin::read_frame(isobus::HardwareInterfaceCANFrame &canFrame)
+{
+	bool retVal = false;
+
+	//Wait for message to be received
+	twai_message_t message;
+	esp_err_t error = twai_receive(&message, portMAX_DELAY);
+	if (error == ESP_OK)
```
ad3154, line 84:
```suggestion
	if (ESP_OK == error)
```

hardware_integration/src/twai_plugin.cpp
```diff
@@ -0,0 +1,129 @@
...
+	bool retVal = false;
+	twai_message_t message;
+
+	message.identifier = canFrame.identifier;
+	message.extd = canFrame.isExtendedFrame;
+	message.data_length_code = canFrame.dataLength;
+	memcpy(message.data, canFrame.data, canFrame.dataLength);
+
+	esp_err_t error = twai_transmit(&message, portMAX_DELAY);
+	if (error == ESP_OK)
```
ad3154, line 119:
```suggestion
	if (ESP_OK == error)
```

ad3154 (commented):
Looking good, with couple tiny comments. I've had to write SPI drivers before, I feel your pain haha. I really appreciate this addition to the project!

hardware_integration/src/twai_plugin.cpp
GwnDaan: Suggestion applied

hardware_integration/src/twai_plugin.cpp
GwnDaan: Suggestion applied

hardware_integration/src/mcp2515_can_interface.cpp
GwnDaan: Simplified to `bool retVal = write_frame(canFrame, MCPRegister::TXB0CTRL, MCPRegister::TXB0SIDH);`

GwnDaan:
When building and running via PlatformIO, none of the can transport protocols get registered. As in this list remains empty:
```cpp
		static std::vector<CANLibProtocol *> protocolList; ///< A list of all created protocol classes
```
I suspect the static instance of the different transport protocols get optimized out. I'll have to dig deeper into this.

GwnDaan:
Just found out https://github.com/ad3154/ISO11783-CAN-Stack/commit/9a8e389159925b2ffdeb451c686e749cac3f81e4 is a recent commit which probably fixes this.

hardware_integration/src/mcp2515_can_interface.cpp
GwnDaan: Agreed, added a `SPITransactionFrame` class in https://github.com/ad3154/ISO11783-CAN-Stack/pull/91/commits/5cdd9daa43de5360fe677c76b0cd392445602928

hardware_integration/CMakeLists.txt
```diff
@@ -8,6 +8,8 @@ set(HARDWARE_INTEGRATION_SRC_DIR "src")
 set(HARDWARE_INTEGRATION_INCLUDE_DIR "include/isobus/hardware_integration")
 
 if(NOT CAN_DRIVER)
+  message(AUTHOR_WARNING "No CAN driver specified, choosing Linux socket CAN by default. Use -DCAN_DRIVER=\"SocketCAN\" to specify another driver. or -DCAN_DRIVER=\"SocketCan;MCP2515\" to specify multiple drivers.")
```
ad3154, line 11: Duplicate message to line 13, probably just missed it when rebasing.  I would suggest combining the two into:
```
"No CAN driver specified, choosing Linux socket CAN by default. Set the CAN_DRIVER variable with -DCAN_DRIVER=<driver> to specify another driver. or -DCAN_DRIVER=\"SocketCAN;MCP2515;WindowsPCANBasic\" to specify multiple drivers."
```

hardware_integration/include/isobus/hardware_integration/mcp2515_can_interface.hpp
```diff
@@ -0,0 +1,176 @@
...
+	/// @param[in, out] canFrame The frame that was read
+	/// @param[in] ctrl The control register of the buffer to read from
+	/// @param[in] data The data register of the buffer to read from
+	/// @param[in] intf The interrupt flag of the buffer to reset after reading
+	/// @returns If the read was successfull
+	bool read_frame(isobus::HardwareInterfaceCANFrame &canFrame, const MCPRegister ctrl, const MCPRegister data, const std::uint8_t intf);
+
+	/// @brief Write a frame to a buffer on the mcp2515
+	/// @param[in] canFrame The frame to write
+	/// @param[in] ctrl The control register of the buffer to write to
```
ad3154, line 166:
```suggestion
	/// @param[in] ctrlRegister The control register of the buffer to write to
```

hardware_integration/include/isobus/hardware_integration/mcp2515_can_interface.hpp
```diff
@@ -0,0 +1,176 @@
...
+	/// @param[in] sidh The sidh register of the buffer to write to
```
ad3154, line 167:
```suggestion
	/// @param[in] sidhRegister The sidh register of the buffer to write to
```

hardware_integration/include/isobus/hardware_integration/mcp2515_can_interface.hpp
```diff
@@ -0,0 +1,176 @@
...
+
+	/// @brief write a single byte register of the mcp2515
+	/// @param[in] address The address of the register to write
+	/// @param[in] data The data to write to the register
+	/// @returns If the write was successfull
+	bool write_register(const MCPRegister address, const std::uint8_t data);
+
+	/// @brief write multiple byte register of the mcp2515
+	/// @param[in] address The address of the register to write
+	/// @param[in] data The data to write to the register
```
ad3154, line 143:
```suggestion
	/// @param[in] data The data to write to the register
	/// @param[in] length The length of the data to write to the register
```

hardware_integration/include/isobus/hardware_integration/mcp2515_can_interface.hpp
```diff
@@ -0,0 +1,176 @@
...
+
+	/// @brief Write a frame to a buffer on the mcp2515
+	/// @param[in] canFrame The frame to write
+	/// @param[in] ctrl The control register of the buffer to write to
+	/// @param[in] sidh The sidh register of the buffer to write to
+	/// @returns If the write was successfull
+	bool write_frame(const isobus::HardwareInterfaceCANFrame &canFrame, const MCPRegister ctrlRegister, const MCPRegister sidhRegister);
+
+	SPIHardwarePlugin *transactionHandler; ///< The SPI transaction handler
+	const std::uint8_t cfg1, cfg2, cfg3; ///< The configuration values for can and clock speed
```
ad3154, line 172: Doxygen complains about this, can you split them up onto separate lines with individual doxygen comments?

Something like:
```
const std::uint8_t cfg1; ///< Comment
const std::uint8_t cfg2; ///< Comment
const std::uint8_t cfg3; ///< Comment
```

ad3154 (changes requested):
Just requested a few things to clean up the remaining doxygen warnings and to fix one thing that looks like a rebase conflict.

hardware_integration/CMakeLists.txt
GwnDaan: I see, I changed your suggestion a bit to make it even more abstract, let me know what you think:
```cmake
if(NOT CAN_DRIVER)
  if (WIN32)
    set(CAN_DRIVER "WindowsPCANBasic")
  else()
    set(CAN_DRIVER "SocketCAN")
  endif()

  message(
    AUTHOR_WARNING
    "No CAN driver specified, choosing ${CAN_DRIVER} by default. Set the CAN_DRIVER variable with -DCAN_DRIVER=<driver> to specify another driver. or -DCAN_DRIVER=\"<driver1>;<driver2>\" to specify multiple drivers."
  )
endif()
```

GwnDaan: I guess we also need some documentation soon on what drivers are available with the growing number of them

GwnDaan: Added relevant information to `README.md` in commit https://github.com/ad3154/ISO11783-CAN-Stack/pull/91/commits/7e7b48ec7bb22af825ca84740fb4d0caef7d3a2e

ad3154: Looks good to me!

hardware_integration/include/isobus/hardware_integration/mcp2515_can_interface.hpp
ad3154: I'm still seeing it as warnings, did you commit these doxygen changes?
```
mcp2515_can_interface.hpp:173: warning: Member cfg1 (variable) of class MCP2515CANInterface is not documented.
mcp2515_can_interface.hpp:174: warning: Member cfg2 (variable) of class MCP2515CANInterface is not documented.
mcp2515_can_interface.hpp:175: warning: Member cfg3 (variable) of class MCP2515CANInterface is not documented.
```

GwnDaan: Uhm pretty sure a0166b6f3789b2a9f91e1d34f98abefd86dd6c76 includes this fix. I'll try to run doxygen on my end

GwnDaan: Oh I see, forgot to put ///> instead of //

GwnDaan: Should be fixed in https://github.com/ad3154/ISO11783-CAN-Stack/pull/91/commits/6ee216497c6b2752a884cada03438f98ef7a624d

ad3154 (approved):

PR #97, opened 2022-12-29
[Core]: Changed protocol list to be non-static to remove the need for "whole-archive"

ad3154:
Dealing with the "whole-archive" behavior between compilers was very annoying, so I made the affected objects non-static members of the network manager. Now they will not be removed by your linker erroneously. Updated readme and tutorial docs to match this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,12 +61,10 @@ find_package(Threads)
 
 add_subdirectory(<path to this submodule>)
 
-target_link_libraries(<your executable name> PRIVATE -Wl,--whole-archive isobus::Isobus -Wl,--no-whole-archive isobus::HardwareIntegration isobus::SocketCANInterface)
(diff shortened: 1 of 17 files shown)


ad3154:
One minor downside of this change is that TP and ETP cannot simply be compiled out anymore to exclude them, but I think this is an OK sacrifice to make, as most people will have no reason to remove the transport layer, and will now not have to fiddle with proprietary linker commands to ensure TP and ETP work.

ad3154:
Seems to be working OK. Using the diagnostic protocol example, we can see that we get 4 protocols created, which are TP, ETP, DP, and PGN Request, but the linker correctly excludes Fast Packet because it is never used. Previously it would also exclude TP and ETP if `whole-archive` was not set. 
image

Issue #101, opened 2023-01-01
Add Community Health Files


RileyKenyon:
I see you already have the `CODE_OF_CONDUCT` - may be worthwhile adding additional items such as `CONTRIBUTING` and templates for feature requests / bug reports.

For reference:
https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file

ad3154:
I have proposed a contributing document in #104 for your review, as well as @GwnDaan who I think asked for the debugger instructions initially.

PR #104, opened 2023-01-02
[Meta]: Add CONTRIBUTING.md

ad3154:
Added basic contibution guidelines and instructions for how to debug the examples using VS Code.

Part of #101


CONTRIBUTING.md
```diff
@@ -0,0 +1,79 @@
...
+
+You can easily set up a linux PC or virutal machine to develop for this project!
+
+We suggest using Ubuntu or Raspian, but other systems may also work. Windows development is supported with PEAK CAN hardware devices.
+
+### Install The Prerequisites
+
+Linux:
+
+Install git, a working CXX compiler, cmake, and git
```
RileyKenyon, line 31: git is written here twice, could replace with "Install the following tools:" to avoid maintenance of writing dependencies multiple times?

CONTRIBUTING.md
```diff
@@ -0,0 +1,79 @@
...
+
+```
+sudo apt install cmake build-essentials git clang-format graphviz gdb
+```
+
+Install your favorite IDE if you want. We suggest Visual Studio Code because it's easy and we will use it in the remainder of this guide.
+
+VS Code can be installed via the snap store, or by downloading it from Microsoft.
```
RileyKenyon, line 39:
```suggestion
VS Code can be installed via the snap store, or by downloading it from [Microsoft](https://code.visualstudio.com/Download).
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,79 @@
...
+
+![BuildTheProject](docs/images/buildingProjectFromCode.png)
+
+This will compile the static libraries that are the core of this project.
+
+## Building Examples
+
+If you want to build the examples and debug with them, you'll need to tell that to CMake.
+
+Go into the CMake Tools extension settings, and add a new configure-time argument for `-DBUILD_EXAMPLES=true`. You may need to scroll down a ways to get to it.
```
RileyKenyon, line 65: I think the other documentation is `-DBUILD_EXAMPLES=ON` may be good to match

RileyKenyon (changes requested):
Couple comments - thanks for adding it in!

CONTRIBUTING.md
ad3154: Changed to match

CONTRIBUTING.md
ad3154: Applied your suggestion

CONTRIBUTING.md
ad3154: Changed to list of tools, with some marked as optional

CONTRIBUTING.md
```diff
@@ -0,0 +1,84 @@
...
+## What Are The Requirements?
+
+Our style rules and PR reviews are based loosely on Autosar and MISRA suggestions and try to ensure the highest quality possible.
+
+* Contributions must follow the style defined in our `.clang-format` file. You can ensure you pass this check by running `find . -iname *.hpp -o -iname *.cpp | xargs clang-format -i` at the root of the repo.
+* Contributions should follow these additional style requirements, which will be checked in code reviews.
+	* Function names `snake_case`
+	* Variables `camelCase`
+	* Constant values `CAPITALIZED_SNAKE`
+	* Constants on the left in `==` and `!=` checks. Like this: `if (5 == value)` NOT `if (value == 5)`. This is to prevent accidentally omitting an `=` in the operator and creating a runtime bug.
```
GwnDaan, line 14: Could it be useful to add a line about usage of `std::` here?

CONTRIBUTING.md
```diff
@@ -0,0 +1,84 @@
...
+* Builds must pass the compilation github action
+* `NULL` should not be used when `nullptr` can be used
+* Prefer C++ over C, though exceptions may be granted if needed
+* C++11 or earlier is required to help provide the maximum compatibility to all modern compilers, especially for embedded ones that do not support the most up-to-date versions of the standard, such as ARM Compiler 5 and 6.
+
+## Setting up a Development Environment With GUI
+
+You can easily set up a linux PC or virtual machine to develop for this project!
+
+We suggest using Ubuntu, Debian, RHEL, or Raspian, but other systems may also work. Windows development is supported with PEAK CAN hardware devices.
```
GwnDaan, line 27: Raspbian* instead of Raspian

GwnDaan (commented):
Had two minor things, other than that it looks good! Thanks

CONTRIBUTING.md
```diff
@@ -0,0 +1,84 @@
...
+	* Variables `camelCase`
+	* Constant values `CAPITALIZED_SNAKE`
+	* Constants on the left in `==` and `!=` checks. Like this: `if (5 == value)` NOT `if (value == 5)`. This is to prevent accidentally omitting an `=` in the operator and creating a runtime bug.
+* Doxygen should compile with no warnings when run at the root of the project with the command `doxygen doxyfile`
+* Copyright notice must be included in each file
+* Absolutely no code shall be added that is under a more strict license than MIT or which has not had conditions met to be distributed under our license
+* Builds must pass the compilation github action
+* `NULL` should not be used when `nullptr` can be used
+* Prefer C++ over C, though exceptions may be granted if needed
+* C++11 or earlier is required to help provide the maximum compatibility to all modern compilers, especially for embedded ones that do not support the most up-to-date versions of the standard, such as ARM Compiler 5 and 6.
```
jqquanbeck-raven, line 21: ARM Compiler V6 fully supports C++14. May be somewhat freeing to drop support for V5 which doesn't even fully support C++11.

CONTRIBUTING.md
jqquanbeck-raven: Furthermore, this should really be a given for any C++ developer, but a note that `using namespace x` should never be in a header file may be good to have

jqquanbeck-raven (changes requested):

CONTRIBUTING.md
ad3154: Yeah, I can get on-board with that. 5 is technically EOL I think so I am ok with raising it to 14

ad3154:
Added and clarified some style items, such as:

* Explicit namespacing should be used for accessing namespaces outside of our namespace `isobus`, especially for the standard library `std::`
* No `using namespace` directives are allowed in header files
* Constructors with a single parameter shall be marked `explicit`
* Where possible, append `const` to functions that do not modify class member variables
* Where possible, declare variables as `constexpr` if their value can be determined at compile time
* Only those characters specified in the C++ Language Standard basic source character set shall be used in the source code except within the text of a wide string.

Fixed spelling of Raspbian
Updated CXX standard to 14 by default. I had originally set it to 11 to try and support ARMCC (ARM Compiler 5) but to be honest, that compiler is EOL and doesn't support the full C++11 standard anyways so it may not have worked to begin with.

RileyKenyon (approved):
Looks good to me!

GwnDaan (approved):

PR #113, opened 2023-01-07
[#112]: Update VT tutorial with CMake, Fix omissions and a couple outdated function names

### LINKED ISSUE #54: Prefer modern cmake in tutorials
- [x] Use alias targets in the tutorials
- [ ] Use Threads::Threads target vs `CMAKE_THREAD_LIBS_INIT` variable
- [x] Provide a recommendation for consumng an installed version of ISCO11783-CAN-Stack with a call to `find_package(isobus REQUIRED)`
- [x] target_link_libraries in tutorials should specify linkage as PRIVATE.

### LINKED ISSUE #112: Issue while following the Virtual Terminal client tutorial
Hi,
I have come across another issue whilst following the tutorials in the online documentation.
I am trying to run the code copied directly from the Final Result section of the Virtual Terminal page of the tutorial, but am having another issue with linking the header files. When I run <cmake --build build> I get the error below. It would be great if you could let me know how to resolve it. It seems to be unable to find the header file for a script within "utility" from one of the scripts within "isobus", so thought it might be a wider issue than just the tutorial.

Charlie

<[ 96%] Building CXX object CMakeFiles/isobus_vt_test.dir/main.cpp.o
In file included from /home/charlie/Documents/ISObus/test_VT_client/main.cpp:6:
/home/charlie/Documents/ISObus/test_VT_client/ISO11783-CAN-Stack/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp:14:10: fatal error: isobus/utility/processing_flags.hpp: No such file or directory
   14 | #include "isobus/utility/processing_flags.hpp"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/isobus_vt_test.dir/build.make:63: CMakeFiles/isobus_vt_test.dir/main.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:186: CMakeFiles/isobus_vt_test.dir/all] Error 2
make: *** [Makefile:152: all] Error 2
>

ad3154:
* Replaced ${CMAKE_THREAD_LIBS_INIT} with Threads::Threads (See #54)
* Added CMake instructions to VT tutorial
* Fixed `RegisterVTButtonEventCallback` and `RegisterVTSoftKeyEventCallback` should be snake_case
* Clarified use and instructions around `objectPoolObjects.h` and `VT3TestPool.iop`
* Added a section about the other callbacks, like string value events
* Added some more references to the API docs

Closes #112 
Closes #54

### DIFF
--- a/sphinx/source/Tutorials/Virtual Terminal Basics.rst
+++ b/sphinx/source/Tutorials/Virtual Terminal Basics.rst
@@ -537,7 +537,7 @@ We can do that with this little handy bit of CMake:
 		TARGET vt_example
 		POST_BUILD
 		COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/VT3TestPool.iop
-		${CMAKE_CURRENT_BINARY_DIR}/VT3TestPool.iop)
+		$<TARGET_FILE_DIR:vt_example>/VT3TestPool.iop)
 
 Now you should be able to build and run the program!
 
--- a/sphinx/source/Tutorials/Virtual Terminal Basics.rst
+++ b/sphinx/source/Tutorials/Virtual Terminal Basics.rst
@@ -23,7 +23,15 @@ This object will act as your interface for all VT communication.
 The client requires two things to instantiate, a :code:`PartneredControlFunction` and a :code:`InternalControlFunction`. 
 This is so that it can send messages on your behalf needed to maintain the connection and simplify the API over sending raw CAN messages to the VT.
 
-Let's start our program similarly to the other tutorials, complete with the requisite control functions.
+Let's start our program fresh, with a folder containing only the CAN stack.
+
+.. code-block:: bash
+
+	mkdir vt_example
+	cd vt_example
+	git clone https://github.com/ad3154/ISO11783-CAN-Stack.git
+
+Create the file `main.cpp` as shown below inside that folder with the requisite control functions.
 
 .. code-block:: c++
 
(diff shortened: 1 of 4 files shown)


sphinx/source/Tutorials/Virtual Terminal Basics.rst
```diff
@@ -456,6 +491,63 @@ Here's the final code for this example:
...
+We also want to move our IOP file to be in the same folder as the executable after it's built, so that it can locate it.
+We can do that with this little handy bit of CMake:
+
+.. code-block:: text
+
+	add_custom_command(
+		TARGET vt_example
+		POST_BUILD
+		COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/VT3TestPool.iop
+		${CMAKE_CURRENT_BINARY_DIR}/VT3TestPool.iop)
```
GwnDaan, line 540: Building for Windows generates an executable in `./build/debug/VT3ExampleTarget.exe` while the object pool is moved to `./build/VT3TestPool.iop`. What we could use instead to correct this:
```suggestion
		$<TARGET_FILE_DIR:vt_example>/VT3TestPool.iop)
```
From: https://stackoverflow.com/a/56231218

GwnDaan (changes requested):
Good stuff! I noticed the custom cmake command for moving the object pool file didn't work as intended while following this tutorial on windows.

ad3154:
> Good stuff! I noticed the custom cmake command for moving the object pool file didn't work as intended while following this tutorial on windows.

Good catch! Admittedly I only tested it on Linux... On a related note, I wonder how tough it would be to set up a Windows CMake compile github action in addition to the linux one we already have... will have to look into that. Could help catch issues for one OS that might not affect the others.

PR #118, opened 2023-01-08
[VT]: Automatically clean up old versions of object pools

ad3154:
This change will delete stale object pools that the VT has saved to NVM for your ISO NAME. This is mostly a courtesy to other people so you're not hogging space on the VT for an object pool that you're not using actively.


ad3154:
Need opinions on this - should the stack always do this, or should it be an option you can set? If an option, should it be on or off by default?

GwnDaan:
Hmm, I'd say let the stack always deleted unused stored object pools for now. I don't really see a use case for a release state application to consist of different (rotating) object pools. It might be useful in a debugging environment where you want to switch quickly between a few (previously loaded) object pools. Though I don't see myself doing that very often and I guess it would take more time to change the option than to upload the object pools.

ad3154:
OK - yeah I agree with that. I haven't actually tested this yet, so I'll make sure it works then bring it in as just being automatic.

ad3154:
image
Looks OK. Merging 😀

PR #120, opened 2023-01-10
[Core]: Improve destination validation in "any control function" callbacks

ad3154:
This takes the need off the consumer to check that the destination was an internal control function or a broadcast.

### DIFF
--- a/isobus/src/can_network_manager.cpp
+++ b/isobus/src/can_network_manager.cpp
@@ -551,7 +551,9 @@ namespace isobus
 		const std::lock_guard<std::mutex> lock(anyControlFunctionCallbacksMutex);
 		for (auto &currentCallback : anyControlFunctionParameterGroupNumberCallbacks)
 		{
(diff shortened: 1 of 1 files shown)


isobus/src/can_network_manager.cpp
```diff
@@ -551,7 +551,9 @@ namespace isobus
 		const std::lock_guard<std::mutex> lock(anyControlFunctionCallbacksMutex);
 		for (auto &currentCallback : anyControlFunctionParameterGroupNumberCallbacks)
 		{
-			if (currentCallback.get_parameter_group_number() == currentMessage.get_identifier().get_parameter_group_number())
+			if ((currentCallback.get_parameter_group_number() == currentMessage.get_identifier().get_parameter_group_number()) &&
+			    ((ControlFunction::Type::Internal == currentMessage.get_destination_control_function()->get_type()) ||
```
ad3154, line 555: Need to reverse the order on this check so the null check comes first...

PR #131, opened 2023-01-15
Control function maintenance improvements

### LINKED ISSUE #130: [Core]: Creating a partner matching an existing CF doesn't map the two together
**Describe the bug**
Partner registration seems to have a bug where if you create a partner after the target device has already claimed, the stack will not go back through and match it with that previously claimed control function. We should tolerate creation and destruction of these partners irrespective of if they've claimed yet.

**Supporting Documentation**
Using the VT example as, well, an example:
If you add a delay between creating the ICF and the partner like this, the partner is never valid.
```
TestInternalECU = std::make_shared<isobus::InternalControlFunction>(TestDeviceNAME, 0x1C, 0);
std::this_thread::sleep_for(std::chrono::milliseconds(3000));
TestPartnerVT = std::make_shared<isobus ::PartneredControlFunction>(0, vtNameFilters);
```

Log output:
```
Loaded object pool from VT3TestPool.iop
[NM]: New Control function 245
[NM]: New Control function 37
[NM]: New Control function 249
[NM]: New Control function 247
[NM]: New Control function 38
[NM]: New Control function 240
```

 - OS: Windows 11
 - Compiler: MSVC 19.34.31933.0
 - CAN Driver: PEAK

ad3154:
## What's new:

Added lots of improvements to control function list/table maintenance.

* Fix issues with mapping a `PartneredControlFunction` with an `ExternalControlFunction` when the partner was created after the stack already had an `ExternalControlFunction` object matching that ECU. In other words, if you made a partner after address claiming was complete, the stack would not resolve that partner to the right ECU. Now it should work no matter when you create a partner.
* Fixed some missing checks for can channel ID.
* Fixed not cleaning up our internal tables if a partner was deleted. Now you can delete partners that were created dynamically and we will automatically either prune it from the appropriate lists or replace the `PartneredControlFunction` with a `ExternalControlFunction`.
* Added a mutex to protect control function tables, that way we can synchronize creation and deletions of partners against the CAN Network Manager's update thread.
* Changed some u32's to `size_t` when they were iterators for an stl container

Closes #130

### DIFF
--- a/isobus/src/can_network_manager.cpp
+++ b/isobus/src/can_network_manager.cpp
@@ -178,10 +178,14 @@ namespace isobus
 
 	void CANNetworkManager::update()
 	{
+		const std::lock_guard<std::mutex> lock(ControlFunction::controlFunctionProcessingMutex);
+
 		if (!initialized)
 		{
 			initialize();
 		}
+
+		update_new_partners();
 
 		process_rx_messages();
 
@@ -189,7 +193,7 @@ namespace isobus
 
 		if (InternalControlFunction::get_any_internal_control_function_changed_address({}))
 		{
-			for (std::uint32_t i = 0; i < InternalControlFunction::get_number_internal_control_functions(); i++)
+			for (std::size_t i = 0; i < InternalControlFunction::get_number_internal_control_functions(); i++)
 			{
 				InternalControlFunction *currentInternalControlFunction = InternalControlFunction::get_internal_control_function(i);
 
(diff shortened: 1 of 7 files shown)


isobus/src/can_network_manager.cpp
```diff
@@ -281,6 +285,39 @@ namespace isobus
 		CANNetworkManager::CANNetwork.receive_can_message(tempCANMessage);
 	}
 
+	void CANNetworkManager::on_partner_deleted(PartneredControlFunction *partner, CANLibBadge<PartneredControlFunction>)
+	{
+		CANStackLogger::CAN_stack_log("[NM]: A partner was deleted.");
```
GwnDaan, line 290: Maybe add information which partner was deleted, e.g:
```suggestion
		CANStackLogger::CAN_stack_log("[NM]: Partner " + isobus::to_string(static_cast<int>(partner->get_address()))  + " was deleted.");
```

GwnDaan (changes requested):
I can remember that I noticed there was something odd with the claiming of CF's, so good job with these improvements! I saw one small thing that could help in better debugging in general.

PR #138, opened 2023-01-19
[Core]: Add basic named reading of byte buffer

### LINKED ISSUE #137: Make standard types easier to read from byte arrays
We work with a lot of byte arrays in this stack. We currently perform bit operations manually to read the standard types - uint16_t, uint32_t, uint64_t, float, double - from these arrays. For example:
https://github.com/ad3154/ISO11783-CAN-Stack/blob/22094dd22a460e46d8778d9a40c26ac5d529a3d0/isobus/src/isobus_virtual_terminal_client.cpp#L2721-L2726

I feel like we can improve this to:
```cpp
std::uint16_t objectID = data.read_uint16_at(1);
std::uint32_t value = data.read_uint32_at(4);
```
Or if they were sequential (used same example for clarity):
```cpp
std::uint16_t objectID = data.read_uint16();
std::uint8_t reserved = data.read_uint8();
std::uint32_t value = data.read_uint32();
```

There are also other example where you would only want to read certain bits, for this we could make use of masks:
```cpp
std::uint8_t someProperty = data.read_masked_uint8_at(0x04, 1); // Where 1 is the index and 0x04 the mask
// Or if we know that the reading is sequential:
std::uint8_t otherProperty = data.read_masked_uint8(0x04);
```

Also reading some booleans should be doable:
```cpp
bool someProperty = data.read_bool_at(1);
bool otherProperty = data.read_bool();
// Or if we want to apply some mask
bool someMaskedProperty = data.read_masked_bool_at(0x10, 3); // Where 1 is the index and 0x10 the mask
bool otherMaskedProperty = data.read_masked_bool(0x010);
```

Even though reading bits and bytes interchangeably in a byte array system could get quite difficult comprehend, I think this 'named' reading would already be more easy to get your head around for new contributors than the current approach.

Furthermore, I don't expect the performance loss to be noticeable if any. I think the overhead would be slightly greater but the overall code size of the program would probably reduce? I really don't have that much knowledge about the trade-offs, but I can do some testing on this if wanted? And also make a proof of concept of some sort? I'd love to hear opinions on this.

GwnDaan:
Allows reading of the following types directly from `CANMessage` object using dedicated functions:
- uint8_t 
- uint16_t
- uint32_t
- uint64_t
- bool

Also adjusted the `process_rx_message` in `VirtualTerminalClient` to use these functions already.

Closes #137

### DIFF
--- a/isobus/src/can_message.cpp
+++ b/isobus/src/can_message.cpp
@@ -7,6 +7,7 @@
 /// @copyright 2022 Adrian Del Grosso
 //================================================================================================
 #include "isobus/isobus/can_message.hpp"
+#include <cassert>
 
 namespace isobus
 {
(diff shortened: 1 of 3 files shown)


isobus/src/can_message.cpp
```diff
@@ -62,4 +62,97 @@ namespace isobus
 		return CANPortIndex;
 	}
 
+	std::uint8_t CANMessage::get_uint8_at(const std::size_t index)
+	{
+		std::uint8_t retVal = 0;
```
ad3154, line 67: Since the default return values of 0 could be "valid" or legal values, do you think it might make sense to assert on the index boundary check? I think assertion would be fairly reasonable to do, as trying to access a value without checking the message DLC is more of a development error? Or, maybe instead it makes sense to make these always do `.at` so that the error gets caught more easily?

GwnDaan: I see, that makes sense. I'll change it

isobus/src/can_message.cpp
```diff
@@ -7,6 +7,7 @@
 /// @copyright 2022 Adrian Del Grosso
 //================================================================================================
 #include "isobus/isobus/can_message.hpp"
+#include "assert.h"
```
ad3154, line 10:
```suggestion
#include <cassert>
```

isobus/src/can_message.cpp
```diff
@@ -62,4 +63,79 @@ namespace isobus
...
+			retVal |= static_cast<std::uint64_t>(data.at(index + 5)) << 16;
+			retVal |= static_cast<std::uint64_t>(data.at(index + 6)) << 8;
+			retVal |= data.at(index + 7);
+		}
+		return retVal;
+	}
+	bool isobus::CANMessage::get_bool_at(const std::size_t byteIndex, const std::uint8_t bitIndex, const std::uint8_t length)
+	{
+		assert(length <= 8 - bitIndex && "length must be less than or equal to 8 - bitIndex");
+		uint8_t mask = ((1 << length) - 1) << bitIndex;
```
ad3154, line 137:
```suggestion
		std::uint8_t mask = ((1 << length) - 1) << bitIndex;
```

isobus/src/can_message.cpp
GwnDaan: Fixed in https://github.com/ad3154/ISO11783-CAN-Stack/pull/138/commits/0cd2603d4f9f2f75d37433d11e85ecb4f59ac04f. I chose for the `vector::at()` as asserts typically get removed with "release" builds, and with the stack still in somewhat early development, I feel like it will be helpful someday to not have the boundary check optimized out for now. We can change this later when optimizations are needed I guess.

ad3154 (commented):
Looking pretty good - have you tested this yet to make sure it still works as expected? If not I can also try it if needed.

GwnDaan:
> have you tested this yet to make sure it still works as expected?

Sort of... I ran the `vt_version_3_object_pool` example which worked fine, and checked the diff whether my changes were correct. But I could've probably missed something, so feel free do any additional testing if you feel like!

ad3154 (approved):

PR #140, opened 2023-01-21
[VT]: Fix accessing a u24 as a u32 when parsing VT NACK

ad3154:
When we were getting NACKs it was causing a crash due to accessing index 8 after converting to the named read functions.

### DIFF
--- a/isobus/src/can_message.cpp
+++ b/isobus/src/can_message.cpp
@@ -84,6 +84,24 @@ namespace isobus
 		return retVal;
 	}
 
+	std::uint32_t CANMessage::get_uint24_at(const std::size_t index, const ByteFormat format)
+	{
+		std::uint32_t retVal;
+		if (ByteFormat::LittleEndian == format)
+		{
+			retVal = data.at(index);
+			retVal |= static_cast<std::uint32_t>(data.at(index + 1)) << 8;
+			retVal |= static_cast<std::uint32_t>(data.at(index + 2)) << 16;
+		}
+		else
+		{
+			retVal = static_cast<std::uint32_t>(data.at(index + 2)) << 16;
+			retVal |= static_cast<std::uint32_t>(data.at(index + 1)) << 8;
+			retVal |= data.at(index + 2);
+		}
+		return retVal;
+	}
+
 	std::uint32_t CANMessage::get_uint32_at(const std::size_t index, const ByteFormat format)
 	{
 		std::uint32_t retVal;
--- a/isobus/src/isobus_virtual_terminal_client.cpp
+++ b/isobus/src/isobus_virtual_terminal_client.cpp
@@ -2586,7 +2586,7 @@ namespace isobus
 				{
 					if (AcknowledgementType::Negative == static_cast<AcknowledgementType>(message->get_uint8_at(0)))
 					{
-						std::uint32_t targetParameterGroupNumber = message->get_uint32_at(5, CANMessage::ByteFormat::BigEndian);
+						std::uint32_t targetParameterGroupNumber = message->get_uint24_at(5);
 						if (static_cast<std::uint32_t>(CANLibParameterGroupNumber::ECUtoVirtualTerminal) == targetParameterGroupNumber)
 						{
 							CANStackLogger::CAN_stack_log("[VT]: The VT Server is NACK-ing our VT messages. Disconnecting.");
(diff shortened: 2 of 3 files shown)


ad3154:
@GwnDaan Do you think we need `get_uint24_at`? Or is that too weird?

GwnDaan:
@ad3154 I think I've seen the u24 in a few more places throughout the standards, so I guess we could implement it as a function? Though they are definitely not as common as the others, idk.

ad3154:
> @ad3154 I think I've seen the u24 in a few more places throughout the standards, so I guess we could implement it as a function? Though they are definitely not as common as the others, idk.

I suppose we may as well add it... I will convert my PR to that instead

isobus/src/can_message.cpp
```diff
@@ -84,6 +84,24 @@ namespace isobus
...
+		if (ByteFormat::LittleEndian == format)
+		{
+			retVal = data.at(index);
+			retVal |= static_cast<std::uint32_t>(data.at(index + 1)) << 8;
+			retVal |= static_cast<std::uint32_t>(data.at(index + 2)) << 16;
+		}
+		else
+		{
+			retVal = static_cast<std::uint32_t>(data.at(index)) << 16;
+			retVal = static_cast<std::uint32_t>(data.at(index + 1)) << 8;
```
GwnDaan, line 99:
```suggestion
			retVal = static_cast<std::uint32_t>(data.at(index) + 2) << 16;
			retVal |= static_cast<std::uint32_t>(data.at(index + 1)) << 8;
```

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -2586,7 +2586,7 @@ namespace isobus
 				{
 					if (AcknowledgementType::Negative == static_cast<AcknowledgementType>(message->get_uint8_at(0)))
 					{
-						std::uint32_t targetParameterGroupNumber = message->get_uint32_at(5, CANMessage::ByteFormat::BigEndian);
+						std::uint32_t targetParameterGroupNumber = message->get_uint24_at(5);
```
GwnDaan, line 2589:
```suggestion
						std::uint32_t targetParameterGroupNumber = message->get_uint24_at(5, CANMessage::ByteFormat::BigEndian);
```

GwnDaan (changes requested):
Looking good, though a few missed things I think

isobus/src/can_message.cpp
ad3154: Copy-paste strikes again

isobus/src/isobus_virtual_terminal_client.cpp
ad3154: I think little endian is correct? The PGN is encoded like this:
```
buffer[5] = static_cast<std::uint8_t>(parameterGroupNumber & 0xFF);
buffer[6] = static_cast<std::uint8_t>((parameterGroupNumber >> 8) & 0xFF);
buffer[7] = static_cast<std::uint8_t>((parameterGroupNumber >> 16) & 0xFF);
```

GwnDaan: Oh huh, it was in big endian before I think? Maybe that was a bug aswell? https://github.com/ad3154/ISO11783-CAN-Stack/blob/22094dd22a460e46d8778d9a40c26ac5d529a3d0/isobus/src/isobus_virtual_terminal_client.cpp#L2591-L2593

GwnDaan: Oh nvm I can't read, it has the `<<` instead of `>>` haha

GwnDaan: > ```
> buffer[5] = static_cast<std::uint8_t>(parameterGroupNumber & 0xFF);
> buffer[6] = static_cast<std::uint8_t>((parameterGroupNumber >> 8) & 0xFF);
> buffer[7] = static_cast<std::uint8_t>((parameterGroupNumber >> 16) & 0xFF);
> ```

Huh I'm totally confused now. Isn't the little endian is the `<<`?

GwnDaan: > Huh I'm totally confused now. Isn't the little endian the `<<`?
Okay nvm again. That was the writing part of course, whereas we are reading it now so should be good

ad3154: Haha, no worries 🙂

PR #148, opened 2023-01-22
[#41]: Add cmake-lint and cmake-format checks to GitHub actions

### LINKED ISSUE #41: Define a CMake style for the repo
For the 1.2% of this codebase that is cmake, there should be a coding standard that can be linted against in PR's. This ensures contributors are abiding by a common set of guidelines, just like the C++ code. One thing I noticed was that conditional statements were not indented.

**Proposal**
Use [cmake-lint](https://cmake-format.readthedocs.io/en/latest/cmake-lint.html) and [cmake-format](https://cmake-format.readthedocs.io/en/latest/cmake-format.html)

**Implementation**
Once the tooling and styles are decided, a PR can be issued to do the following:
1. Add the necessary configuration files if defaults are not used
1. Determine how the linter/formatter can know which files need to be formatted/linted if automatic recursive detection is not supported
1. Add information to the README on the usage of the linter and formatter
1. Go through and fix all current lint/format issues in the codebase
1. Future: Add support to call the linter and/or formatter from cmake as part of CI to block non-compliant PR's

Thoughts?

GwnDaan:
**What's new:**
- Added the cmake-format configuration file: `.cmake-format.yaml`.
- Combined clang-format and cmake-format checks into one GitHub action with multiple jobs.
- Ran cmake-format on the current CMakeLists.txt files, such that the cmake-format and cmake-lint accepts them.
- Set `find_package(Isobus REQUIRED)` if the `BUILD_EXAMPLES` is not set, instead of always checking `CONFIG`.
- Added cmake-format command to format all `CMakeLists.txt` files to `CONTRIBUTING.md`.

Closes #41


ad3154:
I appreciate this, it's been on the back burner since forever ago 😣

ad3154 (approved):

PR #150, opened 2023-01-24
[Meta]: Small QOL updates to Github actions files

GwnDaan:
- Updated clang-format linting action:
  - The `socket_can` matrix path is not yet there anymore and resulted into running clang-format in the whole repo instead. But this only took around 50 seconds so it's also acceptable to just run one clang-format linting action for the whole repo as it won't take too long and we wouldn't have the same problem again in the future.
  - Also removed some other properties as they were either default one or optional and wouldn't get used.
  - Bumped the action from 4.9.0 to 4.10.1
- Added `-DBUILD_TESTING=ON` to the cmake build action so the `ctest` job can actually run testing. (Also my vscode wanted to do some formatting on the `cmake.yml` file which I guess is fine as it looks more inline with the other action files now.)

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -2,9 +2,9 @@ name: CMake
 
 on:
   push:
-    branches: [ "main" ]
+    branches: ["main"]
   pull_request:
-    branches: [ "main" ]
+    branches: ["main"]
 
 env:
   # Customize the CMake build type here (Release, Debug, RelWithDebInfo, etc.)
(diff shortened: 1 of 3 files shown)


GwnDaan:
Added some instructions to the readme for how to build tests, since it had some missing steps for me when I tried following it

.github/workflows/cmake.yml
```diff
@@ -18,20 +18,19 @@ jobs:
...
+      - uses: actions/checkout@v3
 
-    - name: Configure CMake
-      # Configure CMake in a 'build' subdirectory. `CMAKE_BUILD_TYPE` is only required if you are using a single-configuration generator such as make.
-      # See https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html?highlight=cmake_build_type
-      run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DBUILD_EXAMPLES=ON
+      - name: Configure CMake
+        # Configure CMake in a 'build' subdirectory. `CMAKE_BUILD_TYPE` is only required if you are using a single-configuration generator such as make.
+        # See https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html?highlight=cmake_build_type
+        run: cmake -B ${{github.workspace}}/build -DCMAKE_BUILD_TYPE=${{env.BUILD_TYPE}} -DBUILD_EXAMPLES=ON -DBUILD_TESTING=ON -DCAN_DRIVER=SocketCAN
```
ad3154, line 26: I think we've got `-DBUILD_EXAMPLES=ON -DBUILD_TESTING=ON` here now when we only need one

.github/workflows/linting.yml
```diff
@@ -4,21 +4,11 @@ jobs:
   clang-format:
     name: clang-format style
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        path:
-          - "isobus"
-          - "socket_can"
-          - "examples"
```
ad3154, line 12: We may want to find some way to still exclude some of the 3rd party driver files, as their licenses may not allow us to modify them, such as if they are LGPL licensed

.github/workflows/cmake.yml
ad3154: Ah, never mind I mis-read it haha

.github/workflows/linting.yml
GwnDaan: Hmm, I see. Do we have any that should be excluded right now? I think we can use `exclude-regex` option for this? E.g:
```yml
# Some files can be excluded from the clang-format style check, like 3rd party driver files,
# as their licenses may not allow us to modify them, such as if they are LGPL licensed.
# The exclude-regex option can be used to exclude these files from the style check:
exclude-regex: ^(PCANBasic\.h)$
```
Or is there something else you had in mind?

ad3154: I have one on a branch here https://github.com/ad3154/ISO11783-CAN-Stack/blob/feature/innomaker-usb2can/hardware_integration/include/isobus/hardware_integration/libusb.h but nothing to worry about in main, as long as there's a way to do it I'm content 😁

GwnDaan: Hahah, all good, I've added the file to exclusion for you already (in the hope you aren't gonna rename it haha). You can easily append multiple files by using the `|` OR in the exclude regex filter 😄

ad3154 (approved):

PR #152, opened 2023-01-24
[Meta]: Add sonar cloud scanner

ad3154:
Added github action for scanning the repo with sonar cloud

### DIFF
--- a/sonar-project.properties
+++ b/sonar-project.properties
@@ -1,0 +1,14 @@
+sonar.projectKey=ad3154_ISO11783-CAN-Stack
+sonar.organization=isobus-plusplus
+
+# This is the name and version displayed in the SonarCloud UI.
+#sonar.projectName=ISO11783-CAN-Stack
+#sonar.projectVersion=1.0
+
+# Path is relative to the sonar-project.properties file. Replace "\" by "/" on Windows.
+sonar.sources=isobus/src,examples,hardware_integration,utility
+
+# Encoding of the source code. Default is default system encoding
+#sonar.sourceEncoding=UTF-8
+
+sonar.cfamily.cache.enabled=false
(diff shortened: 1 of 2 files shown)


ad3154:
Hmmm... either the code is really good, or it doesn't produce much until it runs on the default branch...

sonar-project.properties
```diff
@@ -0,0 +1,14 @@
+sonar.projectKey=ad3154_ISO11783-CAN-Stack
+sonar.organization=isobus-plusplus
+
+# This is the name and version displayed in the SonarCloud UI.
+#sonar.projectName=ISO11783-CAN-Stack
+#sonar.projectVersion=1.0
+
+# Path is relative to the sonar-project.properties file. Replace "\" by "/" on Windows.
+sonar.sources=isobus/src,examples,hardware_integration,utility
```
GwnDaan, line 9:
```suggestion
sonar.sources=isobus,examples,hardware_integration,utility

# Exclude some files from the analysis.
sonar.exclusions=**/PCANBasic.h
```

GwnDaan (changes requested):
Looks good for now I think. I believe the current CLI won't produce 'coverage' results as described here [linux-cmake-gcovr-gh-actions-sc](https://github.com/sonarsource-cfamily-examples/linux-cmake-gcovr-gh-actions-sc), but we can add that later, the percentage of code coverage won't be high anyway I'm afraid 😅

GwnDaan:
> Hmmm... either the code is really good, or it doesn't produce much until it runs on the default branch...

Yea.. I think the code analysis on a PR only takes into account the code that has changed, which for this PR is none. When we merge it, it will tell us a lot more, at least I hope so.

GwnDaan (approved):

PR #155, opened 2023-01-26
[Meta]: Add coverage analyses action of SonarCloud

GwnDaan:
This PR adds the parts of the GitHub action necessary for SonarCloud to generate coverage analyses.

Reference used: https://github.com/sonarsource-cfamily-examples/linux-cmake-gcovr-gh-actions-sc/blob/main/.github/workflows/build.yml

### DIFF
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -12,19 +12,36 @@ jobs:
     env:
       BUILD_WRAPPER_OUT_DIR: build_wrapper_output_directory # Directory where build-wrapper output will be placed
     steps:
-      - uses: actions/checkout@v3
+      - name: Checkout
+        uses: actions/checkout@v3
         with:
-          fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
+          fetch-depth: 0 # Shallow clones should be disabled for a better relevancy of analysis
+      - name: Set up Python 3.8 for gcovr
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+      - name: install gcovr 5.0
+        run: |
+          pip install gcovr==5.0 # 5.1 is not supported
       - name: Install sonar-scanner and build-wrapper
         uses: SonarSource/sonarcloud-github-c-cpp@v1
       - name: Run build-wrapper
         run: |
           mkdir build
-          cmake -S . -B build -DBUILD_EXAMPLES=ON -DCAN_DRIVER=SocketCAN
+          cmake -S . -B build -DBUILD_EXAMPLES=ON -DBUILD_TESTING=ON -DCAN_DRIVER=SocketCAN
           build-wrapper-linux-x86-64 --out-dir ${{ env.BUILD_WRAPPER_OUT_DIR }} cmake --build build/ --config Release
+      - name: Run tests to generate coverage statistics
+        run: |
+          ./build/unit_tests
+      - name: Collect coverage into one XML report
+        run: |
+          gcovr --sonarqube > coverage.xml
       - name: Run sonar-scanner
         env:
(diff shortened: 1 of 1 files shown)


GwnDaan:
Hmm, @ad3154 are you sure `SONAR_TOKEN` is correctly set for this project? Strange that this error comes up while the other analyses have worked.
```
ERROR: Error during SonarScanner execution
ERROR: Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator```

**Comment by ad3154** [comment]
> Hmm, @ad3154 are you sure `SONAR_TOKEN` is correctly set for this project? Strange that this error comes up while the other analyses have worked.
> 
> ```
> ERROR: Error during SonarScanner execution
> ERROR: Project not found. Please check the 'sonar.projectKey' and 'sonar.organization' properties, the 'SONAR_TOKEN' environment variable, or contact the project administrator``` 
> ```

image

I thought so at least, that is strange...

**Comment by GwnDaan** [comment]
I also see a few posts out there saying it has to do with permissions, but I can't really find any concrete solution. It might be the case that it is breaking now I'm making the PR unlike the other ones?

**Comment by ad3154** [comment]
Yeah, maybe... I have you added to the Sonar project but maybe your actions can't access the repo's secrets? I will try and poke around the settings...

**Comment by ad3154** [comment]
```
Anyone with collaborator access to this repository can use these secrets and variables for actions. They are not passed to workflows that are triggered by a pull request from a fork.
```
Yeahhhh that's our issue... 

Well, that is a bit unfortunate. I can add you as a collaborator to the repo which would fix this and allow you write access to the main repo also, but if we're going to start going down that path I probably need to add some additional PR rules and processes to prepare us for more _public_ write access to the repo, haha

ad3154:
Or, we can change the action to only run on main after the PR is complete, but I do not love that option... it prevents us from being proactive and forces us to be more reactive

GwnDaan:
Hmm, I see. We can make a `if` statement in the sonarcloud action and let it only run when `if: "${{ env.SONAR_TOKEN != '' }}"` is satisfied. This way PR's from contributors with write access have the benefit of analysis, while PR's from other people are not blocked. Though this wouldn't be optimal I think it's better than only running on the main branch.

GwnDaan:
It probably also satisfies to give me Global Execute Analysis permission or Execute Analysis permission in the sonarcloud environment. That way I can associatie a secret `SONAR_TOKEN` on my 'fork' repository which will be used when creating PR's I think. While this would not be a permanent solution, neither would giving everyone collaboration rights on GitHub.. Reference: https://docs.sonarqube.org/latest/user-guide/user-account/generating-and-using-tokens/

ad3154:
Hmm.... well, it's a start I suppose! I think I've given you the execute permissions you need in the sonarcloud environment

GwnDaan:
Well, this is vague... And I guess it shouldn't be used in the way I thought it could work lol. 

This PR skipped failed on the `if: "${{ env.SONAR_TOKEN != '' }}"` check (as can be seen [here](https://github.com/ad3154/ISO11783-CAN-Stack/actions/runs/4018407014/jobs/6903984899)). But a PR to my fork's main branch got through successfully [here](https://github.com/GwnDaan/ISO11783-CAN-Stack/actions/runs/4018455198/jobs/6904094645). HOWEVER on [SonarCloud](https://sonarcloud.io/project/pull_requests_list?id=ad3154_ISO11783-CAN-Stack) it acted as if it were run on #1...

So I guess we should NOT go this way haha. Nevertheless the `if` check is included now, so this PR is ready to review/merge if you wish.

GwnDaan:
Nevermind, I need to change one more thing, hold up

GwnDaan:
Should be good now!

ad3154 (approved):

PR #156, opened 2023-01-26
[Meta]: Add missing tests directory parameter

GwnDaan:
I missed this parameter in #155 I'm afraid. There were no coverage results in [this action](https://github.com/ad3154/ISO11783-CAN-Stack/runs/10912856878). And I guess we won't find out if this fixes the missing coverage results either without merging it... Or is there any other way I'm missing here?


GwnDaan:
Nevermind, I forgot something more as well I think...

GwnDaan:
This should do it I guess...

ad3154 (approved):

PR #160, opened 2023-01-27
[Core]: Add InnoMaker USB2CAN plugin

### LINKED ISSUE #134: Add support for Inno-Maker USB2CAN
I have an inno-maker USB2CAN that has an open source driver here https://github.com/INNO-MAKER/usb2can

This adapter has a native Linux driver (Socket CAN) but it would be fairly trivial to add support for Windows, which would be convenient given the adapter's low cost. Can maybe also add MacOS but I do not have a Mac to test on, so we would need someone to help test that if we add it.

GwnDaan:
After struggling for a while, here is the USB2CAN plugin. The documentation for it is minimal lol. I was able to run the VT example successfully while using this plugin in the end though.

Note: Make sure to add the `InnoMakerUsb2CanLib.dll` file to the directory you are running the executable from. E.g. when building the `vt_version_3_object_pool` example from top level in debug mode. Add the dll file to the `/build/examples/vt_version_3_object_pool/Debug/` folder.

Closes #134


GwnDaan:
@ad3154 Maybe useful to check if this works on your single bus variant as well?

ad3154:
> The documentation for it is minimal lol

Yeahhhhhhh that is partially why I struggled as well. I had to email them to get the right bit timings...

GwnDaan:
> Yeahhhhhhh that is partially why I struggled as well. I had to email them to get the right bit timings...

Hahahah. I had to get them from the C# example they provided...

ad3154:
I don't think the actual .lib files are committed? I only see the PCAN ones in "hardware_integration\lib\Windows"

```
error : '.../hardware_integration/lib/Windows/InnoMakerUsb2CanLib64.lib', needed by 'examples/vt_version_3_object_pool/VT3ExampleTarget.exe', missing and no known rule to make it
```

GwnDaan:
Oh welp. They we're filtered out by `.gitignore`. They should be added now

ad3154:
I think maybe we need that DLL committed also since we're not compiling it ourselves...

As another note, I am getting a lot of `Command line warning D9002: ignoring unknown option '--coverage'` so we might want to check `AND NOT MSVC` or something when adding that, IDK

GwnDaan:
> I think maybe we need that DLL committed also since we're not compiling it ourselves...

Hmm, I think we could add it somewhere yeah, but I don't think we are able to automatically move it to the correct path since that is project specific right?

> As another note, I am getting a lot of `Command line warning D9002: ignoring unknown option '--coverage'` so we might want to check `AND NOT MSVC` or something when adding that, IDK

I see, didn't know that was a thing on windows. Gotta love platform specific stuff haha

ad3154:
Yeah, it'd be tough to move to the right place... Best we could probably do is move it like we do the IOP file for the example, to make sure it ends up somewhere in the build directory, but yeah in the end it'll be up to the consumer to install that DLL to an appropriate location at least for now.

ad3154:
That is one thing that is nice about the PEAK driver, is that their DLLs get installed with their installer, so they are always present. Innomaker's "here's a loose DLL" approach is.... not my favorite.

GwnDaan:
> That is one thing that is nice about the PEAK driver, is that their DLLs get installed with their installer, so they are always present. Innomaker's "here's a loose DLL" approach is.... not my favorite.

Definitely true... There are more things about innomaker's approach that I question haha

>Best we could probably do is move it like we do the IOP file for the example, to make sure it ends up somewhere in the build directory

That would mean adding a move command to each and every example we provide right? Since that is the only place where the executable location is known. I'm not really a fan I think of the strategy where one example is provided with multiple possible can drivers... I feel like it would be better to have concise documentation on this topic instead?

ad3154:
> That would mean adding a move command to each and every example we provide right? 

I guess that's not what I had in mind, I just meant we could move it to wherever isobus.lib ends up, but that also doesn't really help much... the user would have to move it anyways.

So yeah we should probably add documentation for use of each driver we add, and leave the examples the way they are?

Maybe it also makes sense to add an additional CMake AUTHOR_WARNING message about that DLL when they choose that driver?

GwnDaan:
Sounds good. I think all of the important things for this PR are implemented now

ad3154:
Tested with my single channel probe and it worked OK!

PR #162, opened 2023-01-28
[Tests]: Add partnered CF claim test and cover NAME/NAMEFilter with unit tests

GwnDaan:
### What's new:
- Resolved some issues with VirtualCANPlugin, mostly memory related.
- Transfer claim tests to use the new VirtualCANPlugin.
- Extended claim test with partnered CF's testing.
- Covered the remaining parts of NAME and NAMEFilter with unit tests.
- Add extra logging in NAME for when parameters out of range are provided.


GwnDaan:
Ah, it would have been cool to see the SonarCloud analysis now haha

ad3154 (approved):

isobus/src/can_NAME.cpp
```diff
@@ -79,10 +84,12 @@ namespace isobus
 
 	void NAME::set_function_code(std::uint8_t value)
 	{
-		uint64_t tempNAME = rawName;
-		tempNAME &= ~static_cast<std::uint64_t>(0xFF0000000000);
-		tempNAME |= (static_cast<std::uint64_t>(value & 0xFF) << 40);
-		set_full_name(tempNAME);
+		if (value > 0xFF)
```
ad3154, line 87: Ah I didn't catch this before I merged it, but this check does nothing, as a uint8_t cannot be > 0xFF, so some compilers will throw a warning

GwnDaan: Oh I see, I can remove it in the other PR if you want?

ad3154: Yeah if you don't mind that'd be good

PR #163, opened 2023-01-29
[Meta]: Examples with automatic driver selection and less global variables

GwnDaan:
### What's new:
- Add automatic includes of available can drivers with the `isobus/hardware_integration/available_can_drivers.hpp` header file.
- Examples can now always be compiled. It will give a nice logging message if a CAN driver is missing and exit the application.
- Change the CANHardwareInterface to use `std::shared_ptr<CANHardwarePlugin>` instead of raw pointers.
- Minimize the amount of global variables. I think this is a good practice to reduce 'undefined behaviour'.
- Changed `exit` to `_exit` in signal handlers ([reference](https://stackoverflow.com/a/8833602))
- Updated tutorial website according to these changes.

I didn't intend to do so much at once but while I was at it I thought why not. Any feedback on these changes is welcome, I'm nowhere near experienced in c++ haha. 

I also wonder if we should change the other raw pointers to use `shared_ptr` and similar in the future? I think SonarCloud hints at that a bit.

### DIFF
--- a/examples/nmea2000/main.cpp
+++ b/examples/nmea2000/main.cpp
@@ -10,8 +10,6 @@
 #include <iterator>
 #include <memory>
 
-static std::shared_ptr<isobus::InternalControlFunction> TestInternalECU = nullptr;
-
 using namespace std;
 
 void nmea2k_callback(isobus::CANMessage *message, void *)
@@ -40,7 +38,7 @@ void signal_handler(int signum)
 {
 	CANHardwareInterface::stop();
 	isobus::FastPacketProtocol::Protocol.remove_multipacket_message_callback(0x1F001, nmea2k_callback, nullptr);
-	exit(signum);
+	_exit(EXIT_FAILURE);
 }
 
 void update_CAN_network()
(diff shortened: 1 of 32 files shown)


GwnDaan:
There are some warnings I still need to fix

GwnDaan:
Should be good now! I think this PR is ready for review

ad3154 (approved):

ad3154:
All good stuff 😁

examples/nmea2000/main.cpp
```diff
@@ -43,11 +34,11 @@ void nmea2k_transmit_complete_callback(std::uint32_t parameterGroupNumber,
 	}
 }
 
-void signal_handler(int signum)
+void signal_handler(int)
 {
 	CANHardwareInterface::stop();
 	isobus::FastPacketProtocol::Protocol.remove_multipacket_message_callback(0x1F001, nmea2k_callback, nullptr);
-	exit(signum);
+	_exit(EXIT_FAILURE);
```
ad3154, line 41: This worked on one linux PC, but not another.
```
../../../examples/vt_version_3_object_pool/main.cpp:31:2: error: ‘_exit’ was not declared in this scope; did you mean ‘_Exit’?
   31 |  _exit(EXIT_FAILURE);
      |  ^~~~~
      |  _Exit
ninja: build stopped: subcommand failed.
```

ad3154: Changing to `_Exit` seemed to work on both... maybe we should change it to that?

PR #164, opened 2023-02-01
[#65] Add dynamic object pool scaling

### LINKED ISSUE #65: Support for different widths and heights of data masks in VT example
The terminal on which the example was tested supports 2 pixel masks modes. One has a 260x260 pixel mask and the other is 600x600 pixels. 

The example gets clipped when running on the 260x260mask:

While the 600x600 mask displays:

Therefore I'm assuming the VT example implements a data mask of around 500x500 pixels. What would be the best approach to support different masks? Is there anything written about this in the iso 11783 standard?

ad3154:
## What's New
* The VT Client can now automatically scale an object pool to your destination VT! 🥳

Here are some screenshots from my VT showing some different scale factors. My terminal has 480x480 data mask area, and 72x72 soft key designators.

image

Closes #65 

### Minor stuff
* Made one minor update to the example object pool to fix some object alignment.
* Cleaned up some lingering doxygen warnings.
* Clarified tutorial build instruction


GwnDaan:
Hmm, it seems to fail on my end. When I add 
```c++
TestVirtualTerminalClient->set_object_pool_scaling(0, 400, 400);
```
to the vt_version3_object_pool after it's set, which was still missing I guess? I get the following error when it's trying to scale:
image
image
image

ad3154:
Dang! I'll see if I can reproduce it... I also need to write some tests to make Sonar happy...

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -1530,6 +1534,17 @@ namespace isobus
 		}
 	}
 
+	void VirtualTerminalClient::set_object_pool_scaling(std::uint8_t poolIndex,
+	                                                    std::uint32_t originalDataMaskDimensions_px,
+	                                                    std::uint32_t originalSoftKyeDesignatorHeight_px)
+	{
+		if (poolIndex < objectPools.size())
```
GwnDaan, line 1541: I feel like an assertion here be easier to debug?

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -1827,17 +1842,19 @@ namespace isobus
...
-									                                                                         process_internal_object_pool_upload_callback);
+									transmitSuccessful = CANNetworkManager::CANNetwork.send_can_message(static_cast<std::uint32_t>(CANLibParameterGroupNumber::ECUtoVirtualTerminal),
+									                                                                    nullptr,
+									                                                                    objectPools[i].objectPoolSize + 1, // Account for Mux byte
+									                                                                    myControlFunction.get(),
+									                                                                    partnerControlFunction.get(),
+									                                                                    CANIdentifier::CANPriority::PriorityLowest7,
+									                                                                    process_callback,
+									                                                                    this,
+									                                                                    process_internal_object_pool_upload_callback);
```
GwnDaan, line 1857: What's the reason behind this change? I tend to find it a bit harder to read personally

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -3438,36 +3476,1772 @@ namespace isobus
...
+				else
+				{
+					CANStackLogger::error("[VT]: Failed to resize an object: " +
+					                      isobus::to_string(static_cast<int>((*poolIterator)) | (static_cast<int>((*poolIterator + 1))) << 8) +
+					                      " with type " +
+					                      isobus::to_string(static_cast<int>((*poolIterator + 2))) +
+					                      " with size " +
+					                      isobus::to_string(static_cast<int>(objectSize)));
+				}
+				poolIterator += get_number_bytes_in_object(&poolIterator[0]);
```
GwnDaan, line 3620:
```suggestion
				poolIterator += objectSize;
```

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -3438,36 +3476,1772 @@ namespace isobus
...
+					{
+						retVal = FontSize::Size8x12;
+					}
+					else
+					{
+						retVal = FontSize::Size6x8;
+					}
+				}
+			}
+			break;
```
GwnDaan, line 4507: For maintainability reasons, wouldn't this be easier to parse it using a map instead?

test/vt_client_tests.cpp
```diff
@@ -0,0 +1,738 @@
...
+
+#include "isobus//utility/iop_file_interface.hpp"
+#include "isobus/hardware_integration/can_hardware_interface.hpp"
+#include "isobus/hardware_integration/virtual_can_plugin.hpp"
+#include "isobus/isobus/can_network_manager.hpp"
+#include "isobus/isobus/isobus_virtual_terminal_client.hpp"
+
+using namespace isobus;
+
+class DerivedTestVtClient : public VirtualTerminalClient
```
GwnDaan, line 11:
```suggestion
class DerivedTestVTClient : public VirtualTerminalClient
```

GwnDaan (changes requested):
Looking good! Looks like it was quite some work. I have a few minor suggestions here and there

isobus/src/isobus_virtual_terminal_client.cpp
ad3154: How do you propose to do that? I agree these trees of conditionals is not very nice to read or maintain, so if we can switch it to a nicer structure that'd be ideal. But if the scale factor can be totally arbitrary, like 0.6f or something which is not in this list of conditionals, how would we clamp that to match some key value in the map?

isobus/src/isobus_virtual_terminal_client.cpp
ad3154: I think this is an anachronism from early in my development on this branch. I will switch it back to the original way

ad3154:
@GwnDaan I think I've fixed that crash you were seeing, and we've just got the one unresolved topic left

GwnDaan:
The scaling part works great now, Good job! 😄

isobus/src/isobus_virtual_terminal_client.cpp
GwnDaan: See #173 for my proposal

ad3154:
Our coverage dropping percent is a bit sad, but it is at least somewhat attributable to this: https://github.com/gcovr/gcovr/issues/529

gcovr doesn't like initializer lists it seems...

I have noticed a lot of lines are only partially covered due to gcovr expecting us to test every exception-only path, so it might make sense to disable exception-only paths in its analysis `--exclude-throw-branches`

Regardless, I'll try and add some additional tests...

GwnDaan (dismissed):
Looking dope! This has surely been a long-winded feature to implement. I propose we merge this in sooner than later. I feel like a PR for additional test cases can always be created afterwards, and we have a few PR's now that depend on this one getting merged.

ad3154:
> Looking dope! This has surely been a long-winded feature to implement. I propose we merge this in sooner than later. I feel like a PR for additional test cases can always be created afterwards, and we have a few PR's now that depend on this one getting merged.

Agreed. We're still kind of iterating on our CI settings and things anyways, and we can certainly come through before a stable release and shore up our testing further. I'll merge it in for now after fixing the one conflict.

PR #167, opened 2023-02-02
[Core]: Optimize get pgn callback by returning ptr

GwnDaan:
- Don't always wake up hardware interface thread as it raises watchdog timer exceptions on some platforms (e.g. ESP32). I know sonar doesn't like it, but I guess it gets enough inputs for it to wake up such that is safe to ignore the warning.
- Switch from copying `ParameterGroupNumberCallbackData` to returning a pointer to optimize memory performance.

### DIFF
--- a/isobus/src/can_partnered_control_function.cpp
+++ b/isobus/src/can_partnered_control_function.cpp
@@ -13,6 +13,7 @@
 #include "isobus/isobus/can_network_manager.hpp"
 
 #include <algorithm>
+#include <cassert>
 
 namespace isobus
 {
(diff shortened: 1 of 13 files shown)


isobus/src/can_partnered_control_function.cpp
```diff
@@ -205,20 +206,18 @@ namespace isobus
...
 
-	ParameterGroupNumberCallbackData PartneredControlFunction::get_parameter_group_number_callback(std::uint32_t index) const
+	std::unique_ptr<ParameterGroupNumberCallbackData> PartneredControlFunction::get_parameter_group_number_callback(std::size_t index) const
 	{
-		ParameterGroupNumberCallbackData retVal(0, nullptr, nullptr);
-
 		if (index < get_number_parameter_group_number_callbacks())
 		{
-			retVal = parameterGroupNumberCallbacks[index];
+			return std::make_unique<ParameterGroupNumberCallbackData>(parameterGroupNumberCallbacks[index]);
```
GwnDaan, line 218: I begin to doubt, doesn't this also make a copy of parameterGroupNumberCallbacks[index]? If so, it might be more logical to return a reference instead I think?

isobus/src/can_partnered_control_function.cpp
```diff
@@ -205,20 +205,18 @@ namespace isobus
...
 		}
-		return retVal;
+		throw std::out_of_range("Index out of range");
```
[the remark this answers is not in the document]
ad3154: I'm not a big fan of runtime, non-standard exceptions personally. I'd much rather an `assert` than an exception to catch these kinds of things

isobus/src/can_partnered_control_function.cpp
ad3154: This does construct a new object and calls the copy constructor of the underlying object which could be a bit expensive since this is a pretty hot loop. Returning a reference makes sense I think!

ad3154 (approved):

PR #174, opened 2023-02-09
[VT/TC]: Add language command interface

ad3154:
This is intended to be a component of VT and TC client interfaces.
I'll add it to VT once #65 merges so that I can more easily add unit tests.

### DIFF
--- a/test/language_command_interface_tests.cpp
+++ b/test/language_command_interface_tests.cpp
@@ -7,20 +7,9 @@
 #include "isobus/isobus/can_partnered_control_function.hpp"
 #include "isobus/isobus/can_stack_logger.hpp"
 #include "isobus/isobus/isobus_language_command_interface.hpp"
+#include "isobus/utility/system_timing.hpp"
 
 using namespace isobus;
-
-// A log sink for the CAN stack
-class CustomLogger : public isobus::CANStackLogger
-{
-public:
-	void sink_CAN_stack_log(CANStackLogger::LoggingLevel, const std::string &) override
-	{
-		// Do nothing
-	}
-};
-
-static CustomLogger logger;
 
 TEST(LANGUAGE_COMMAND_INTERFACE_TESTS, BasicConstructionAndInit)
 {
@@ -34,14 +23,13 @@ TEST(LANGUAGE_COMMAND_INTERFACE_TESTS, BasicConstructionAndInit)
 	EXPECT_EQ(0, interfaceUnderTest.get_language_command_timestamp());
 	interfaceUnderTest.initialize(); // double init
 	EXPECT_EQ(true, interfaceUnderTest.get_initialized());
-	interfaceUnderTest.send_request_language_command();
 }
 
 TEST(LANGUAGE_COMMAND_INTERFACE_TESTS, InvalidICF)
 {
 	LanguageCommandInterface interfaceUnderTest(nullptr);
 	interfaceUnderTest.initialize();
-	interfaceUnderTest.send_request_language_command();
+	ASSERT_FALSE(interfaceUnderTest.send_request_language_command());
 }
 
 TEST(LANGUAGE_COMMAND_INTERFACE_TESTS, ValidPartner)
@@ -55,14 +43,17 @@ TEST(LANGUAGE_COMMAND_INTERFACE_TESTS, ValidPartner)
 
 	auto vtPartner = std::make_shared<PartneredControlFunction>(0, vtNameFilters);
 	LanguageCommandInterface interfaceUnderTest(internalECU, vtPartner);
-	ParameterGroupNumberRequestProtocol::assign_pgn_request_protocol_to_internal_control_function(internalECU); // Pre-assign our ICF
 	interfaceUnderTest.initialize();
-	interfaceUnderTest.send_request_language_command();
+	ASSERT_TRUE(interfaceUnderTest.get_initialized());
+	// Technically our address is bad, so this should still not send
+	//! @todo Test with valid address
+	ASSERT_FALSE(interfaceUnderTest.send_request_language_command());
 }
 
 TEST(LANGUAGE_COMMAND_INTERFACE_TESTS, Uninitialized)
 {
 	LanguageCommandInterface interfaceUnderTest(nullptr);
+	ASSERT_FALSE(interfaceUnderTest.get_initialized());
 }
 
 TEST(LANGUAGE_COMMAND_INTERFACE_TESTS, MessageContentParsing)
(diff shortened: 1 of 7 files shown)


test/language_command_interface_tests.cpp
```diff
@@ -0,0 +1,161 @@
...
+	auto vtPartner = std::make_shared<PartneredControlFunction>(0, vtNameFilters);
+	LanguageCommandInterface interfaceUnderTest(internalECU, vtPartner);
+	ParameterGroupNumberRequestProtocol::assign_pgn_request_protocol_to_internal_control_function(internalECU); // Pre-assign our ICF
+	interfaceUnderTest.initialize();
+	interfaceUnderTest.send_request_language_command();
+}
+
+TEST(LANGUAGE_COMMAND_INTERFACE_TESTS, Uninitialized)
+{
+	LanguageCommandInterface interfaceUnderTest(nullptr);
```
GwnDaan, line 65: Missing asserts? I propose we add an if statement in the `initialize()` function of `LanguageCommandInterface` that checks if `myControlFunction` is a nullptr, if so log a error message. Also since the initialize wouldn't have completed successfully, we should be able to assert it with:
```cpp
ASSERT_FALSE(interfaceUnderTest.get_initialized());
```

test/language_command_interface_tests.cpp
```diff
@@ -0,0 +1,161 @@
...
+	EXPECT_EQ(LanguageCommandInterface::DateFormats::ddmmyyyy, interfaceUnderTest.get_commanded_date_format());
+	EXPECT_EQ(LanguageCommandInterface::DistanceUnits::Metric, interfaceUnderTest.get_commanded_distance_units());
+	EXPECT_EQ(LanguageCommandInterface::AreaUnits::NoAction, interfaceUnderTest.get_commanded_area_units());
+	EXPECT_EQ(LanguageCommandInterface::VolumeUnits::US, interfaceUnderTest.get_commanded_volume_units());
+	EXPECT_EQ(LanguageCommandInterface::MassUnits::Metric, interfaceUnderTest.get_commanded_mass_units());
+	EXPECT_EQ(LanguageCommandInterface::TemperatureUnits::Reserved, interfaceUnderTest.get_commanded_temperature_units());
+	EXPECT_EQ(LanguageCommandInterface::PressureUnits::Reserved, interfaceUnderTest.get_commanded_pressure_units());
+	EXPECT_EQ(LanguageCommandInterface::ForceUnits::ImperialUS, interfaceUnderTest.get_commanded_force_units());
+	EXPECT_EQ(LanguageCommandInterface::UnitSystem::Metric, interfaceUnderTest.get_commanded_generic_units());
+	EXPECT_LT(interfaceUnderTest.get_language_command_timestamp(), 100000);
```
GwnDaan, line 126:
```suggestion
	EXPECT_LT(SystemTiming::get_timestamp_ms() - interfaceUnderTest.get_language_command_timestamp(), 1);
```

test/language_command_interface_tests.cpp
```diff
@@ -0,0 +1,161 @@
...
+	interfaceUnderTest.process_rx_message(&testMessage, &interfaceUnderTest);
+
+	// Cover null message
+	interfaceUnderTest.process_rx_message(nullptr, &interfaceUnderTest);
+
+	// Cover null parent
+	interfaceUnderTest.process_rx_message(&testMessage, nullptr);
+
+	// Cover all null parameters
+	interfaceUnderTest.process_rx_message(nullptr, nullptr);
```
GwnDaan, line 156: For these 3, we can assert that values of language command interface remained unchanged

test/language_command_interface_tests.cpp
```diff
@@ -0,0 +1,161 @@
...
+	testMessage.set_data(testData2, 8);
+	// Cover bad one bad reserved byte
+	interfaceUnderTest.process_rx_message(&testMessage, &interfaceUnderTest);
+
+	testData2[6] = 43;
+	testData2[7] = 0xFF;
+	testMessage.set_data_size(0); // Resets the CAN message data vector
+	testMessage.set_data(testData2, 8);
+	// Cover bad one bad reserved byte
+	interfaceUnderTest.process_rx_message(&testMessage, &interfaceUnderTest);
```
GwnDaan, line 147: Maybe add `//! @todo assert that an warning log message came through` for these 3 test cases

test/language_command_interface_tests.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+	// Cover null parent
+	interfaceUnderTest.process_rx_message(&testMessage, nullptr);
+
+	// Cover all null parameters
+	interfaceUnderTest.process_rx_message(nullptr, nullptr);
+
+	// Cover some logger conditions
+	CANStackLogger::set_can_stack_logger_sink(&logger);
+	interfaceUnderTest.process_rx_message(nullptr, nullptr);
```
GwnDaan, line 160: We can't do any meaningful asserts for this I think. I feel like we are better of removing this. What are your thoughts?

GwnDaan (changes requested):
Thanks for implementing this feature! I had a few small remarks

test/language_command_interface_tests.cpp
```diff
@@ -0,0 +1,161 @@
...
+class CustomLogger : public isobus::CANStackLogger
+{
+public:
+	void sink_CAN_stack_log(CANStackLogger::LoggingLevel, const std::string &) override
+	{
+		// Do nothing
+	}
+};
+
+static CustomLogger logger;
```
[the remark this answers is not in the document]
ad3154: Generally I agree, but every time I use the logger it increases the number of uncovered conditions, probably related to exception-only paths, so in addition to testing the logger in its own file, I think we might want to run gcovr with `--exclude-throw-branches`

test/language_command_interface_tests.cpp
```diff
@@ -0,0 +1,161 @@
...
+	std::vector<isobus::NAMEFilter> vtNameFilters;
+	const isobus::NAMEFilter testFilter(isobus::NAME::NAMEParameters::FunctionCode, static_cast<std::uint8_t>(isobus::NAME::Function::VirtualTerminal));
+	vtNameFilters.push_back(testFilter);
+
+	NAME clientNAME(0);
+	auto internalECU = std::make_shared<InternalControlFunction>(clientNAME, 0x26, 0);
+
+	auto vtPartner = std::make_shared<PartneredControlFunction>(0, vtNameFilters);
+	LanguageCommandInterface interfaceUnderTest(internalECU, vtPartner);
+	ParameterGroupNumberRequestProtocol::assign_pgn_request_protocol_to_internal_control_function(internalECU); // Pre-assign our ICF
```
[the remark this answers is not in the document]
ad3154: It's not strictly required, I was trying to cover a missing condition in the initialize with this, but it's not super valuable.

test/language_command_interface_tests.cpp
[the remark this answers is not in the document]
GwnDaan: I also think we should add that in. It would've been useful to include these 'throw-branches' if we were throwing exceptions ourselves, but we aren't. So it's fine by me

test/language_command_interface_tests.cpp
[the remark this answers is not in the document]
GwnDaan: Ohh I see, it's the condition where an parameter control function is already assigned right? I think we should divert this to it's own test file when we implement it later. Since we are now only 'covering' it but not actually testing/'asserting' it.

test/language_command_interface_tests.cpp
ad3154: Agreed, removed.

ad3154:
Thanks for the review @GwnDaan , I've applied your suggestions

GwnDaan (approved):

Issue #175, opened 2023-02-09
Improving project setup for (external) contributions


GwnDaan:
Hi! To optimize the current workflow of PR's and keep it neat, such that we prepare for more contributions in the future, I propose we change the following:
- Combine the workflow action `cmake.yml` and `sonar.yml`, since `sonar.yml` is only a extension on top of `cmake.yml`. They both build the stack and execute tests, and sonar does some analysis on top of that.
- SonarCloud analysis is not executed on PR's from external forks, this due to the sonar secret not accessible in it's workflow run. `pull_request_target` should be able to fix that ([source](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/#improvements-for-public-repository-forks))
- Secure secrets from abuse when using `pull_request_target` with environment protection rules ([reference](https://dev.to/petrsvihlik/using-environment-protection-rules-to-secure-secrets-when-building-external-forks-with-pullrequesttarget-hci))
- Automatically assign reviewers to new PR's, this should be doable by defining so called '[code owners](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners)'

We also might wanna look into moving this repo to an organization profile, I see a few benefits for it:
- Use the name 'isobus++' for the organization which I think would result in more (google/github/etc.) search hits than only having ISO11773-CAN-Stack. 
- Being able to extend to multiple projects in the future (e.g. VT server app, object pool creator, utility programs, etc.) while keeping it under the familiar name 'isobus++'. Which could also increase brand awareness if done right.
- Sounds to me like it's easier to manage permissions for contributors, though I'm not sure on this one

ad3154:
One easy branding-win would be to rename the repo to isobus++ (or isobus-plus-plus if github doesn't allow '+'?). According to this https://docs.github.com/en/repositories/creating-and-managing-repositories/renaming-a-repository renaming the repo will not break access that is using the old name

GwnDaan:
> One easy branding-win would be to rename the repo to isobus++ (or isobus-plus-plus if github doesn't allow '+'?). According to this https://docs.github.com/en/repositories/creating-and-managing-repositories/renaming-a-repository renaming the repo will not break access that is using the old name

I agree with that! That would be a good first step I think.

GwnDaan:
[...]

PR #176, opened 2023-02-09
[#175]: Run analysis on external contributions

GwnDaan:
### What's new
- SonarCloud should now be able to run on external PR's when an admin has approved the PR's content.
- Merged all similar CI workflows into one single workflow with multiple jobs.

**Note (@ad3154):**  for this to work well, a dummy environment has to be created in the project settings called "Integrate Pull Request" according to [source](https://dev.to/petrsvihlik/using-environment-protection-rules-to-secure-secrets-when-building-external-forks-with-pullrequesttarget-hci).


ad3154:
I've created that environment now according to those instructions

GwnDaan:
> I've created that environment now according to those instructions

Sounds good! It's kind of frustratig that the build workflow will only run on the next PR after this PR is merged, at least that is how I understand it. 

Or is there some other way to force the workflow to run? That would be useful to refine it and make sure it works without having to merge...

I think I can try to do some testing with it in my forked repo.

ad3154:
I'm not sure if there's a way to force run it... I'll go ahead and merge it in and we can see how it goes.

GwnDaan:
I'll make a PR for a fix, and then we can see whether it works for a PR or not

PR #177, opened 2023-02-09
[Meta]: Run build workflow when approve job is skipped

GwnDaan:
This was the case when the workflow trigger event ran when a commit was pushed to main branch. See [runs/4136493254](https://github.com/ad3154/ISO11783-CAN-Stack/actions/runs/4136493254)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,6 +28,7 @@ jobs:
     name: Setup
     runs-on: ubuntu-latest
     needs: approve
+    if: ${{ needs.approve.result == 'success' || needs.approve.result == 'skipped' }}
     steps:
       - name: Set up Python 3.8 for gcovr
         uses: actions/setup-python@v4
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,8 +24,8 @@ jobs:
       - name: Approve
         run: echo For security reasons, all pull requests need to be approved first before running any automated build CI.
 
-  setup:
-    name: Setup
+  build_test_analyze:
+    name: Build, Test and Analyze
     runs-on: ubuntu-latest
     needs: approve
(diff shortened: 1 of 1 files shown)


GwnDaan:
It didn't like the multiple jobs thing I made which makes sense I think. So I switched it to a single job for building, testing and analyzing instead. It should run without errors on the next PR (I hope 😄)

ad3154:
It seems like maybe the environment variables don't persist between the stages? Combining the setup and build stages would probably fix it, or manually setting the sonar paths as environment variables in that stage?

GwnDaan:
> It seems like maybe the environment variables don't persist between the stages? Combining the setup and build stages would probably fix it, or manually setting the sonar paths as environment variables in that stage?

Yea, I should've combined the stages in the latest commit. Due to the nature of 'pull_request_target', these changes will only take effect on a next PR I believe. This is due to the fact that it uses the base commit of the branch for it's workflow, and so any changes won't take effect until merged to prevent any abuse of the repo's secrets. Feel free to review the changes and merge when you have the time :)

ad3154 (approved):

GwnDaan:
Hmm.. it still skipped the build/test/analyse workflow step on the commit on the main branch. I guess it wasn't the proper fix haha...

ad3154:
Hmm, do you think it's because of this `if` statement?
```
if: github.event_name == 'pull_request_target'
```

GwnDaan:
That's correct. I tried to fix it by adding 
```
if: ${{ needs.approve.result == 'success' || needs.approve.result == 'skipped' }}
```
to the build/test/analyse workflow step but that wasn't enough apparently. It should've been:
```
if: ${{ always() && needs.approve.result == 'success' || needs.approve.result == 'skipped' }} 
```
I made a PR for it: #178

PR #180, opened 2023-02-11
[Core]: Mitigate an order of operations issue with static destructors

ad3154:
Sometimes the order of static object destruction can be a bit unpredictable, so sometimes I was seeing that the lists that contain the internal and partner control functions would be cleared before we could access them in the destructors of those objects specifically when those objects were global statics (like in the examples). So I added a trivial check to ensure those lists contains > 0 objects before trying to do any further operations with it inside those destructors.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -70,9 +70,9 @@ jobs:
         run: |
           sonar-scanner \
             --define sonar.cfamily.build-wrapper-output="${{ env.BUILD_WRAPPER_OUT_DIR }}" \
-            --define sonar.coverageReportPaths=coverage.xml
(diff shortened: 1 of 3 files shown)


GwnDaan:
I see how this would fix the problem, but I wonder if switching to a vector of weak_ptr would be in general be a more reliable alternative?

ad3154:
With the current state, I should probably also check `internalControlFunctionList.end() != thisObject` to guarantee the object is in the vector? Converting away from raw pointers is probably the long term way to go, but I would like to address that as its own PR I think since that'd be a bigger refactor

GwnDaan:
Sounds good to me!

GwnDaan (approved):

PR #182, opened 2023-02-12
[Core]: Improvements to CANHardwareInterface 

GwnDaan:
This PR is mostly refactoring the CANHardwareInterface class and adding some basic tests to it. I planned to change it up more, but before doing that I think it's useful to get thoughts of others on this changed version. 

The changes I made:
- Replace raw pointers with smart pointers.
- Reduce number of raw locks with gaurded and unique locks.
- Switch canLibNeedsUpdate from bool with mutex, to atomic_bool
- Improved consistency in naming, e.g. camelCase for variables and snake_case for functions

It's been some time since I made these changes and I notice now that I went away from the single-return path for some smaller functions (mostly to use early-returns instead). I know not everyone likes it, but I see a few benefits for it:
- Have the logging message close to the if condition that fails
- Not have a lot of indented if-statements (easier to read on mobile platforms)
- Most of the time reduces the lines of code
- Don't have to trace which else-statement belongs to which if-statement

That said, I'm open to hear the disadvantages.

### DIFF
--- a/hardware_integration/include/isobus/hardware_integration/can_hardware_interface.hpp
+++ b/hardware_integration/include/isobus/hardware_integration/can_hardware_interface.hpp
@@ -38,33 +38,27 @@ public:
 	class CanLibUpdateCallbackInfo
 	{
 	public:
-		/// @brief Constructs a `CanLibUpdateCallbackInfo`, sets default values
-		CanLibUpdateCallbackInfo();
-
 		/// @brief Allows easy comparison of callback data
 		/// @param obj the object to compare against
-		bool operator==(const CanLibUpdateCallbackInfo &obj);
+		bool operator==(const CanLibUpdateCallbackInfo &obj) const;
 
-		void (*callback)(); ///< The callback
-		void *parent; ///< Context variable, the owner of the callback
+		void (*callback)() = nullptr; ///< The callback
+		void *parent = nullptr; ///< Context variable, the owner of the callback
 	};
 
 	/// @brief A class to store information about Rx callbacks
 	class RawCanMessageCallbackInfo
 	{
 	public:
-		/// @brief Constructs a `RawCanMessageCallbackInfo`, sets default values
-		RawCanMessageCallbackInfo();
-
 		/// @brief Allows easy comparison of callback data
 		/// @param obj the object to compare against
-		bool operator==(const RawCanMessageCallbackInfo &obj);
+		bool operator==(const RawCanMessageCallbackInfo &obj) const;
 
-		void (*callback)(isobus::HardwareInterfaceCANFrame &rxFrame, void *parentPointer); ///< The callback
-		void *parent; ///< Context variable, the owner of the callback
+		void (*callback)(isobus::HardwareInterfaceCANFrame &rxFrame, void *parentPointer) = nullptr; ///< The callback
+		void *parent = nullptr; ///< Context variable, the owner of the callback
 	};
 
-	static CANHardwareInterface CAN_HARDWARE_INTERFACE; ///< Static singleton instance of this class
+	CANHardwareInterface() = delete;
 
 	/// @brief Returns the number of configured CAN channels that the class is managing
 	/// @returns The number of configured CAN channels that the class is managing
@@ -74,17 +68,23 @@ public:
 	/// @details Allocates the proper number of `CanHardware` objects to track
 	/// each CAN channel's Tx and Rx message queues. If you pass in a smaller number than what was
 	/// already configured, it will delete the unneeded `CanHardware` objects.
-	/// @note All changes to channel count will be ignored if `start` has been called and the threads are running
+	/// @note The function will fail if the channel is already assigned to a driver or the interface is already started
 	/// @param value The number of CAN channels to manage
 	/// @returns `true` if the channel count was set, otherwise `false`.
 	static bool set_number_of_can_channels(std::uint8_t value);
 
 	/// @brief Assigns a CAN driver to a channel
-	/// @param[in] aCANChannel The channel to assign to
+	/// @param[in] channelIndex The channel to assign to
 	/// @param[in] canDriver The driver to assign to the channel
-	/// @note All changes to driver assignment will be ignored if `start` has been called and the threads are running
+	/// @note The function will fail if the channel is already assigned to a driver or the interface is already started
 	/// @returns `true` if the driver was assigned to the channel, otherwise `false`
-	static bool assign_can_channel_frame_handler(std::uint8_t aCANChannel, std::shared_ptr<CANHardwarePlugin> canDriver);
+	static bool assign_can_channel_frame_handler(std::uint8_t channelIndex, std::shared_ptr<CANHardwarePlugin> canDriver);
+
+	/// @brief Removes a CAN driver from a channel
+	/// @param[in] channelIndex The channel to remove the driver from
+	/// @note The function will fail if the channel is already assigned to a driver or the interface is already started
+	/// @returns `true` if the driver was removed from the channel, otherwise `false`
+	static bool unassign_can_channel_frame_handler(std::uint8_t channelIndex);
 
 	/// @brief Starts the threads for managing the CAN stack and CAN drivers
 	/// @returns `true` if the threads were started, otherwise false (perhaps they are already running)
@@ -97,7 +97,7 @@ public:
 	/// @brief Called externally, adds a message to a CAN channel's Tx queue
 	/// @param[in] packet The packet to add to the Tx queue
 	/// @returns `true` if the packet was accepted, otherwise `false` (maybe wrong channel assigned)
-	static bool transmit_can_message(isobus::HardwareInterfaceCANFrame &packet);
+	static bool transmit_can_message(const isobus::HardwareInterfaceCANFrame &packet);
 
 	/// @brief Adds an Rx callback. The added callback will be called any time a CAN message is received.
 	/// @param[in] callback The callback to add
@@ -116,7 +116,7 @@ public:
 	/// @note All changes to the update delay will be ignored if `start` has been called and the threads are running
 	static void set_can_driver_update_period(std::uint32_t value);
 
-	/// @brief Adds a periodic udpate callback
+	/// @brief Adds a periodic update callback
 	/// @param[in] callback The callback to add
 	/// @param[in] parentPointer Generic context variable, usually a pointer to the owner class for this callback
 	/// @returns `true` if the callback was added, `false` if it was already in the list
@@ -129,14 +129,8 @@ public:
 	static bool remove_can_lib_update_callback(void (*callback)(), void *parentPointer);
 
 private:
-	/// @brief Private constructor, prevents more of these classes from being needlessly created
-	CANHardwareInterface();
-
-	/// @brief Private destructor
-	~CANHardwareInterface();
-
 	/// @brief Stores the Tx/Rx queues, mutexes, and driver needed to run a single CAN channel
-	struct CanHardware
+	struct CANHardware
 	{
 		std::mutex messagesToBeTransmittedMutex; ///< Mutex to protect the Tx queue
 		std::deque<isobus::HardwareInterfaceCANFrame> messagesToBeTransmitted; ///< Tx message queue for a CAN channel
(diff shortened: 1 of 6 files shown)


hardware_integration/src/can_hardware_interface.cpp
```diff
@@ -14,354 +14,327 @@
...
-std::mutex CANHardwareInterface::canLibNeedsUpdateMutex;
-std::mutex CANHardwareInterface::canLibUpdateCallbacksMutex;
-bool CANHardwareInterface::threadsStarted = false;
-bool CANHardwareInterface::canLibNeedsUpdate = false;
-std::uint32_t CANHardwareInterface::canLibUpdatePeriod = CANLIB_UPDATE_RATE;
-CANHardwareInterface CANHardwareInterface::CAN_HARDWARE_INTERFACE;
+std::mutex CANHardwareInterface::rxCallbacksMutex;
+std::mutex CANHardwareInterface::periodicUpdateCallbacksMutex;
+std::atomic_bool CANHardwareInterface::threadsStarted = false;
+std::atomic_bool CANHardwareInterface::canLibNeedsUpdate = false;
```
ad3154, line 28: `= false` isn't portable and won't work on with GCC because it would cause a copy to be created. `= { false }` is portable.

hardware_integration/include/isobus/hardware_integration/can_hardware_interface.hpp
```diff
@@ -166,28 +160,20 @@ class CANHardwareInterface
...
-	static std::mutex rxCallbackMutex; ///< A mutex for protecting the `rxCallbacks`
-	static std::mutex canLibNeedsUpdateMutex; ///< A mutex for protecting the `canLibNeedsUpdate` variable
-	static std::mutex canLibUpdateCallbacksMutex; ///< A mutex for protecting the `canLibUpdateCallbacks`
-	static std::condition_variable threadConditionVariable; ///< A condition variable to allow for signaling the CAN thread from `updateCANLibPeriodicThread`
-	static bool threadsStarted; ///< Stores if `start` has been called yet
-	static bool canLibNeedsUpdate; ///< Stores if the CAN thread needs to update the CAN stack this iteration
+	static std::mutex rxCallbacksMutex; ///< A mutex for protecting the `rxCallbacks`
+	static std::mutex periodicUpdateCallbacksMutex; ///< A mutex for protecting the `periodicUpdateCallbacks`
+	static std::condition_variable threadConditionVariable; ///< A condition variable to allow for signaling the CAN thread from `periodicUpdateThread`
+	static std::atomic_bool threadsStarted; ///< Stores if `start` has been called yet
```
ad3154, line 175: Need #include <atomic> in this file as gcc doesn't pull it in currently

hardware_integration/include/isobus/hardware_integration/virtual_can_plugin.hpp
```diff
@@ -78,7 +85,7 @@ class VirtualCANPlugin : public CANHardwarePlugin
 	const bool receiveOwnMessages; ///< If `true`, the driver will receive its own messages
 
 	std::shared_ptr<VirtualDevice> ourDevice; ///< A pointer to the virtual device of this instance
-	bool running; ///< If `true`, the driver is running
+	std::atomic_bool running; ///< If `true`, the driver is running
```
ad3154, line 88: Need #include <atomic> as gcc doesn't pull it in currently

hardware_integration/include/isobus/hardware_integration/can_hardware_interface.hpp
```diff
@@ -38,33 +38,27 @@ class CANHardwareInterface
 	class CanLibUpdateCallbackInfo
 	{
 	public:
-		/// @brief Constructs a `CanLibUpdateCallbackInfo`, sets default values
-		CanLibUpdateCallbackInfo();
```
ad3154, line 42: Is it better to have `CanLibUpdateCallbackInfo() = default;` ?

hardware_integration/include/isobus/hardware_integration/can_hardware_interface.hpp
```diff
@@ -38,33 +38,27 @@ class CANHardwareInterface
...
+		void (*callback)() = nullptr; ///< The callback
+		void *parent = nullptr; ///< Context variable, the owner of the callback
 	};
 
 	/// @brief A class to store information about Rx callbacks
 	class RawCanMessageCallbackInfo
 	{
 	public:
-		/// @brief Constructs a `RawCanMessageCallbackInfo`, sets default values
-		RawCanMessageCallbackInfo();
```
ad3154, line 57: Is it better to have RawCanMessageCallbackInfo() = default; ?

hardware_integration/src/can_hardware_interface.cpp
```diff
@@ -14,354 +14,327 @@
...
+		isobus::CANStackLogger::error("[HardwareInterface] Cannot start interface more than once.");
+		return false;
+	}
+
+	canThread = std::make_unique<std::thread>(can_thread_function);
+	periodicUpdateThread = std::make_unique<std::thread>(update_can_lib_periodic_function);
+
+	threadsStarted = true;
+
+	for (std::uint32_t i = 0; i < hardwareChannels.size(); i++)
```
ad3154, line 144: Can probably change to size_t

hardware_integration/src/can_hardware_interface.cpp
```diff
@@ -372,77 +345,73 @@ void CANHardwareInterface::can_thread_function()
...
 	{
-		std::unique_lock<std::mutex> lMutex(threadMutex);
-		threadConditionVariable.wait(lMutex);
-		CanHardware *pCANHardware;
+		std::unique_lock<std::mutex> threadLock(threadMutex);
+		threadConditionVariable.wait(threadLock);
 
 		if (threadsStarted)
 		{
 			for (std::uint32_t i = 0; i < hardwareChannels.size(); i++)
```
ad3154, line 353: Can probably change to size_t here also - I guess it's just a general comment we might want to adjust those throughout while we're reworking this file

hardware_integration/src/can_hardware_interface.cpp
```diff
@@ -372,77 +345,73 @@ void CANHardwareInterface::can_thread_function()
...
 							rxCallbacks[j].callback(tempCanFrame, rxCallbacks[j].parent);
 						}
 					}
-					rxCallbackMutex.unlock();
+					rxCallbacksMutex.unlock();
 				}
 			}
 
-			if (get_clear_can_lib_needs_update())
+			if (canLibNeedsUpdate)
```
ad3154, line 381: Where does this get reset to false?

ad3154 (commented):
Improving this file is log overdue, so I'm glad you're taking a look!

hardware_integration/include/isobus/hardware_integration/can_hardware_interface.hpp
GwnDaan: I'm lost here. I can't gasp the difference between no constructor at all and a default one: https://stackoverflow.com/questions/68821529/what-is-the-difference-between-constructor-default-and-the-compiler-generated

ad3154: Hmm, I think in this case since there's no protected or private variables it is the same? This is one situation where maybe Sonar has a preference on it

GwnDaan:
#183 was a fix were a hardware channel is used while it is a nullptr. But with the current setup of `hardwareChannels`, which is a vector of `unique_ptr`, we don't have nullptr's right? Can you verify @ad3154?

hardware_integration/src/can_hardware_interface.cpp
GwnDaan: Oh yea good point, it didn't reset... Should be fixed now, I added `canLibNeedsUpdate = false;` directly underneath.

hardware_integration/include/isobus/hardware_integration/can_hardware_interface.hpp
GwnDaan: Okay, let's see. At least if Sonar works this Sunday haha....

ad3154:
> #183 was a fix were a hardware channel is used while it is a nullptr. But with the current setup of `hardwareChannels`, which is a vector of `unique_ptr`, we don't have nullptr's right? Can you verify @ad3154?

Well, technically `unique_ptr` can be `nullptr`, but it really shouldn't be... I will see if it works OK

ad3154:
One thing I've noticed while testing this is that if the CAN driver is invalid, but the CAN threads started, the thread goes out of scope when we return in main, which is bad... If I add a `Stop` before returning then everything is happy again.
image

ad3154:
> > #183 was a fix were a hardware channel is used while it is a nullptr. But with the current setup of `hardwareChannels`, which is a vector of `unique_ptr`, we don't have nullptr's right? Can you verify @ad3154?
> 
> Well, technically `unique_ptr` can be `nullptr`, but it really shouldn't be... I will see if it works OK

image
I still don't fully understand why it's nullptr here... I might need to load up my linux VM and try it there instead of my WSL2 environment...

GwnDaan:
> One thing I've noticed while testing this is that if the CAN driver is invalid, but the CAN threads started, the thread goes out of scope when we return in main, which is bad... If I add a `Stop` before returning then everything is happy again. image
> 
> Not really a bug, per-se, just something I noticed. `terminate called without an active exception`

Should we manually call `stop()` if `CANHardwareInterface` is still running when deinitialized?

ad3154:
> > One thing I've noticed while testing this is that if the CAN driver is invalid, but the CAN threads started, the thread goes out of scope when we return in main, which is bad... If I add a `Stop` before returning then everything is happy again. image
> > Not really a bug, per-se, just something I noticed. `terminate called without an active exception`
> 
> Should we manually call `stop()` if `CANHardwareInterface` is still running when deinitialized?

Sonar might not like that, but probably? At the very least we want to join the threads I think. If we don't drain the queues that's not a huge deal

GwnDaan:
> Sonar might not like that, but probably? At the very least we want to join the threads I think. If we don't drain the queues that's not a huge deal

That's probably true, as it also doesn't like calling `close()` in `VirtualCANPlugin`:
https://sonarcloud.io/project/issues?open=AYZG_p6ctvhA436Aznxh&id=ad3154_ISO11783-CAN-Stack

I think as you say we should just do the bare minimum, e.g. stopping/join the threads in the deconstructor only. That might keep Sonar happy as well. 😄

hardware_integration/include/isobus/hardware_integration/can_hardware_interface.hpp
GwnDaan: Sonar doesn't seem to bother about not having the default constructor it seems?

ad3154: OK, good enough for me then I suppose

GwnDaan:
> I think as you say we should just do the bare minimum, e.g. stopping/join the threads in the deconstructor only. That might keep Sonar happy as well. 😄

I added this in the latest commit as POC. Let's see what Sonar thinks about it. 

> One thing I've noticed while testing this is that if the CAN driver is invalid, but the CAN threads started, the thread goes out of scope when we return in main, which is bad... If I add a `Stop` before returning then everything is happy again. image
> 
> Not really a bug, per-se, just something I noticed. `terminate called without an active exception`

@ad3154 Can you check if the latest commit of this PR fixes the issue?

ad3154:
> > I think as you say we should just do the bare minimum, e.g. stopping/join the threads in the deconstructor only. That might keep Sonar happy as well. 😄
> 
> I added this in the latest commit as POC. Let's see what Sonar thinks about it.
> 
> > One thing I've noticed while testing this is that if the CAN driver is invalid, but the CAN threads started, the thread goes out of scope when we return in main, which is bad... If I add a `Stop` before returning then everything is happy again. image
> > Not really a bug, per-se, just something I noticed. `terminate called without an active exception`
> 
> @ad3154 Can you check if the latest commit of this PR fixes the issue?

Same thing unfortunately 
image

GwnDaan:
> Same thing unfortunately

And now?

ad3154:
> > Same thing unfortunately
> 
> And now?

Seems like it is working now... I will re-test on one other machine to be sure

ad3154:
> > > Same thing unfortunately
> > 
> > 
> > And now?
> 
> Seems like it is working now... I will re-test on one other machine to be sure

Tested on my RHEL 9.1 VM and it also worked there, so I think we're good

ad3154 (approved):

PR #183, opened 2023-02-12
[Core]: Fix a crash that can occur if a hardware channel is nullptr

ad3154:
This fixes a crash that might manifest if your CAN hardware is invalid and you call `Start` on the hardware interface.

image

### DIFF
--- a/hardware_integration/src/can_hardware_interface.cpp
+++ b/hardware_integration/src/can_hardware_interface.cpp
@@ -382,29 +382,32 @@ void CANHardwareInterface::can_thread_function()
 			{
 				pCANHardware = hardwareChannels[i];
 
-				pCANHardware->receivedMessagesMutex.lock();
-				bool processNextMessage = (!pCANHardware->receivedMessages.empty());
-				pCANHardware->receivedMessagesMutex.unlock();
-
-				while (processNextMessage)
+				if (nullptr != pCANHardware)
 				{
-					isobus::HardwareInterfaceCANFrame tempCanFrame;
-
 					pCANHardware->receivedMessagesMutex.lock();
-					tempCanFrame = pCANHardware->receivedMessages.front();
-					pCANHardware->receivedMessages.pop_front();
-					processNextMessage = (!pCANHardware->receivedMessages.empty());
+					bool processNextMessage = (!pCANHardware->receivedMessages.empty());
 					pCANHardware->receivedMessagesMutex.unlock();
 
-					rxCallbackMutex.lock();
-					for (std::uint32_t j = 0; j < rxCallbacks.size(); j++)
+					while (processNextMessage)
 					{
-						if (nullptr != rxCallbacks[j].callback)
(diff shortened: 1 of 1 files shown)


hardware_integration/src/can_hardware_interface.cpp
```diff
@@ -376,7 +376,8 @@ void CANHardwareInterface::can_thread_function()
 		threadConditionVariable.wait(lMutex);
 		CanHardware *pCANHardware;
 
-		if (threadsStarted)
+		if ((threadsStarted) &&
+		    (nullptr != pCANHardware))
```
GwnDaan, line 380: Isn't pCANHardware always a nullptr when it reaches this if statement?

GwnDaan: Might want to add the `nullptr != pCANHardware` check to line 384 instead?

ad3154: Ah yep, haha, I guess I need more coffee yet this morning

GwnDaan: Though I feel like `hardwareChannels` shouldn't hold on to `nullptr`s in general in the future?

ad3154: Some amount of nullptr's might be inevitable, like if someone creates 3 can channels, then destroys the one at index 0, we would want the indices to remain the same, so leaving the dead channel as nullptr makes some sense I think. In this case, I'm not entirely sure where the nullptr is coming from yet, as there is an object assigned before start is called.

ad3154: I'm also not seeing the crash on Windows with PCAN, only on Linux with socket CAN... more investigation needed

GwnDaan: > Some amount of nullptr's might be inevitable, like if someone creates 3 can channels, then destroys the one at index 0, we would want the indices to remain the same, so leaving the dead channel as nullptr makes some sense I think. In this case, I'm not entirely sure where the nullptr is coming from yet, as there is an object assigned before start is called.

Hmm, I see. The indices are only needed for CANHardwareFrame's channel index assignments, which itself is needed for logic within CANNetwork, right? 

I suggest that in the future we can remove this limitation by assigning a weak_ptr of a CANNetwork object to the CANHardware struct, which contains the info for each hardware channel. (Though this can only be done when and if we deviate from the singleton pattern for CANNetwork I think). Would that work?

ad3154: Yeah, if we want to be free of that limitation then as you said we'll need to map between CANNetwork channels and hardware channels, and I think that would work OK?

GwnDaan: We'll see I guess. It seems like a lot of work for a small improvement.

GwnDaan (approved):

PR #188, opened 2023-02-19
[Core]: Added a way to disable the hardware plugin receive threads

ad3154:
I was having a lot of trouble while testing task controller message encoding because the Rx thread for the virtual can plugin could come along and read messages out of the queues before I could manually read them to verify their content. So, these changes allow us to temporarily stop the Rx threads so that we can synchronously receive messages while running tests to ensure we never randomly lose messages (which we were before). I also added a timeout value to the virtual can read function to 100% avoid an infinite wait. Generally this should not affect anything negatively as the Rx thread will just re-try the read, but it has the upside of preventing a hard lock up while running tests if a message fails to send or something.

### DIFF
--- a/hardware_integration/include/isobus/hardware_integration/can_hardware_interface.hpp
+++ b/hardware_integration/include/isobus/hardware_integration/can_hardware_interface.hpp
@@ -103,6 +103,10 @@ public:
 	/// @returns `true` if the callback was removed, `false` if no callback matched the two parameters
 	static bool remove_can_lib_update_callback(void (*callback)(), void *parentPointer);
 
+	/// @brief Stops only the recieve threads for each channel. Useful for testing. Keeps the Tx and main threads running.
+	/// @returns `true` if the receive threads were joined
+	static bool stop_recieve_threads();
+
 private:
 	/// @brief A class to store information about CAN lib update callbacks
 	class CanLibUpdateCallbackInfo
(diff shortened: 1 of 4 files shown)


ad3154:
I'll rebase this branch and push up my latest TC tests so I can show you, one sec

ad3154:
@GwnDaan  Check this out: https://github.com/ad3154/Isobus-plus-plus/blob/feature/task-controller-client/test/tc_client_tests.cpp

Basically I'm actually transmitting the TC messages all the way through the stack, catching them in the virtual CAN plugin, and verifying the content matches the standard. This way of testing message encoding really exercises all the stuff in between as well, which is nice.

GwnDaan:
Ohh, I see. You are setting the receiveOwnSentMessages like option to true. And then reading the message which is bounced back to the test plugin. That works fine I think, can't really think of any issues with that approach, but I'm still sobering up so that doesn't help. 

Another way would be to use two separate objects, where one mimics the idea of having a separate device on the bus. We don't register the second can plugin to the hardware interface but instead read from it ourselves. See https://github.com/ad3154/Isobus-plus-plus/commit/80153a8629b86f1bf1d7363306c951ca9d18bd23 for this approach. What are your thoughts?

ad3154:
Ah that seems to work too. That approach is less weird than having to stop the rx threads. If you make a PR back to my branch on the main repo, I'd switch approaches and close this PR

GwnDaan:
> Ah that seems to work too. That approach is less weird than having to stop the rx threads. If you make a PR back to my branch on the main repo, I'd switch approaches and close this PR

I have little time to do that the coming few days because of carnival, if you don't mind you can also just cherry-pick my commit.

PR #195, opened 2023-02-26
Add Task Controller Client

### LINKED ISSUE #171: Isobus Task Controller Client Interface
Now that VT auto-scaling is working I'd like to turn my attention to the Task Controller Client interface (ISO11783-10).

This has been one of our most requested features recently.

Here's a high level breakdown of what we'll need to implement:

- [x] Implement base TC objects as C++ objects
- [x] Add language command interface that can be used in both TC and VT
- [x] Handshaking with TC Server (Get Version messages, number of booms/sections supported)
- [x] Construct device descriptor object pools (DDOP)
- [x] DDOP Upload state machine
- [x] Process Data PGN Parsing
- [x] DDOP hashing, version load/store
- [x] Unit Tests Coverage > 80%

It would also be interesting to know exactly how we want to let users build DDOPs, since they are fairly free-form. For example, we could let people construct C++ objects dynamically that represent the objects in their pool, or we could use a file with a proprietary format (like some kind of JSON or YML) that's less nasty than ISO XML, or we could use ISO XML... IOP files.. etc. I'm open to whatever people think is the easiest or best solution there. We could always support multiple methods as well. Suggestions welcome.

ad3154:
# Task Controller Client

This PR adds the first iteration of a task controller client. It implements much of the standard but still needs a few things to be complete which we can iterate on in the near future. I'm making the PR now to start to get some eyes on it, get Sonar's feedback, and to remind me to write more unit tests.

So far, booms created with the TC client will show up as an importable ISO boom at least on my terminal, with the proper units, offsets, and other metadata.

Closes #171 

## What's In this PR?
* Adds the task controller client main class
* Adds a full, binary serializable C++ object oriented implementation of a device descriptor object pool
* Adds a new example program that is still a WIP but sets up a complete DDOP for a sprayer
* Adds unit tests for task controller
* Adds the language command interface to VT so that TC can use it if needed for TC servers version 3 and under
* Adds a little set utility functions to check platform endianness at runtime

## What still needs work? (As far as I know)
* Peer assignment has not been added
* Version that we report to the TC is limited to 3 at the moment (this will be increased once I figure out how to handle it better because a DDOP created for version 4 is not compatible with 3, but we don't check the version till during the connection process... need to integrate the two better)

## Example output
New log messages will help show you what's happening with the TC connection:
image

Here's an example DDOP created by the example program (with some sections omitted for brevity).
```
<DVC A="DVC-1" B="Isobus++ UnitTest" C="1.0.0" D="9004800012E00002" E="123" F="20302E312B2B49" G="FF555500506E65">
	<DET A="DET-5" B="18" C="4" D="Section 0" E="4" F="9">
		<DOR A="66" P848_dpd_value="0"/>
		<DOR A="34" P848_dpd_value="0"/>
		<DOR A="50" P848_dpd_value="0"/>
	</DET>
	<DET A="DET-4" B="84" C="3" D="Product" E="3" F="9">
		<DOR A="86" P848_dpd_value="0"/>
		<DOR A="85" P848_dpd_value="0"/>
	</DET>
	<DET A="DET-3" B="9" C="2" D="Boom" E="2" F="1">
		<DOR A="18" P848_dpd_value="0"/>
		<DOR A="83" P848_dpd_value="0"/>
		<DOR A="82" P848_dpd_value="0"/>
		<DOR A="11" P848_dpd_value="0"/>
		<DOR A="17" P848_dpd_value="0"/>
		<DOR A="16" P848_dpd_value="0"/>
		<DOR A="15" P848_dpd_value="0"/>
	</DET>
	<DET A="DET-2" B="5" C="6" D="Connector" E="1" F="1">
		<DOR A="8" P848_dpd_value="0"/>
		<DOR A="7" P848_dpd_value="0"/>
		<DOR A="6" P848_dpd_value="0"/>
	</DET>
	<DET A="DET-1" B="1" C="1" D="Sprayer" E="0" F="0">
		<DOR A="4" P848_dpd_value="0"/>
		<DOR A="2" P848_dpd_value="0"/>
	</DET>
	<DPD A="83" B="0122" C="3" D="8" E="Setpoint Work State"/>
	<DPD A="82" B="00A1" C="1" D="8" E="Actual Work State 1-16"/>
	<DPD A="86" B="0048" C="3" D="9" E="Tank Volume" F="91"/>
	<DPD A="85" B="0049" C="1" D="9" E="Tank Capacity" F="91"/>
	<DPD A="12" B="0074" C="1" D="16" E="Area Total" F="87"/>
	<DPD A="13" B="0121" C="2" D="8" E="Setpoint Work State"/>
	<DPD A="11" B="0043" C="1" D="8" E="Actual Working Width" F="90"/>
	<DPD A="7" B="0087" C="2" D="0" E="Connector Y" F="89"/>
	<DPD A="6" B="0086" C="2" D="0" E="Connector X" F="89"/>
	<DPD A="4" B="0077" C="3" D="16" E="Total Time" F="88"/>
	<DPD A="3" B="0003" C="0" D="16" E="Request Default PD"/>
	<DPD A="2" B="008D" C="1" D="8" E="Actual Work State"/>
	<DPT A="66" B="0043" C="2134" D="Width" E="90"/>
	<DPT A="50" B="0087" C="-18288" D="Offset Y" E="90"/>
	<DPT A="34" B="0086" C="-20" D="Offset X" E="90"/>
	<DPT A="17" B="0088" C="0" D="Offset Z" E="89"/>
	<DPT A="16" B="0087" C="0" D="Offset Y" E="89"/>
	<DPT A="15" B="0086" C="0" D="Offset X" E="89"/>
	<DPT A="8" B="009D" C="9" D="Type"/>
	<DVP A="88" B="0" C="1.0000000000" D="1" E="minutes"/>
	<DVP A="91" B="0" C="0.0010000000" D="0" E="L"/>
	<DVP A="87" B="0" C="1.0000000000" D="0" E="m^2"/>
	<DVP A="90" B="0" C="0.0010000000" D="0" E="m"/>
	<DVP A="89" B="0" C="1.0000000000" D="0" E="mm"/>
</DVC>
```


ad3154:
I am marking this PR as ready for review and merge. It now has enough features to satisfy most use cases, and the example is a fully functional section control example that I was able to run on my terminal, which is very _very_ cool, haha.
image

haykodarb:
Hi Adrian! 

This is all fantastic work and it's amazing to see an open source project tackling this.

Could I ask what TC Simulator program you are using?

ad3154:
> Could I ask what TC Simulator program you are using?

Hello! In the screenshot above I'm using a real TC, a Raven Viper4. Is there a simulator you'd recommend?
In a perfect world where I had infinite money, the Vector CANoe TC is the gold standard I wish I had access to.

examples/task_controller_client/section_control_implement_sim.cpp
```diff
@@ -0,0 +1,318 @@
...
+{
+	return 1;
+}
+
+constexpr std::uint32_t SectionControlImplementSimulator::get_section_control_state() const
+{
+	return 1;
+}
+
+bool SectionControlImplementSimulator::create_ddop(std::shared_ptr<isobus::DeviceDescriptorObjectPool> poolToPopulate, isobus::NAME clientName)
```
GwnDaan, line 71: Can be const (according to sonar haha)

examples/task_controller_client/section_control_implement_sim.cpp
```diff
@@ -0,0 +1,318 @@
...
+	retVal &= poolToPopulate->add_device_process_data("Actual Rate", static_cast<std::uint16_t>(isobus::DataDescriptionIndex::ActualVolumePerAreaApplicationRate), static_cast<std::uint16_t>(ImplementDDOPObjectIDs::VolumePerAreaPresentation), static_cast<std::uint8_t>(isobus::task_controller_object::DeviceProcessDataObject::PropertiesBit::MemberOfDefaultSet), static_cast<std::uint8_t>(isobus::task_controller_object::DeviceProcessDataObject::AvailableTriggerMethods::OnChange) | static_cast<std::uint8_t>(isobus::task_controller_object::DeviceProcessDataObject::AvailableTriggerMethods::TimeInterval), static_cast<std::uint16_t>(ImplementDDOPObjectIDs::ActualRate));
+	retVal &= poolToPopulate->add_device_property("Operation Type", 3, static_cast<std::uint16_t>(isobus::DataDescriptionIndex::ActualCulturalPractice), isobus::task_controller_object::Object::NULL_OBJECT_ID, static_cast<std::uint16_t>(ImplementDDOPObjectIDs::ActualCulturalPractice));
+
+	// Set up sections for section control
+	// Using 7 ft sections
+	for (std::size_t i = 0; i < NUMBER_SECTIONS; i++)
```
GwnDaan, line 117:
```suggestion
	for (std::uint8_t i = 0; i < NUMBER_SECTIONS; i++)
```
Number of sections shouldn't be able to go above 256

examples/task_controller_client/section_control_implement_sim.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+		boom->add_reference_to_child_object(static_cast<std::uint16_t>(ImplementDDOPObjectIDs::BoomXOffset));
+		boom->add_reference_to_child_object(static_cast<std::uint16_t>(ImplementDDOPObjectIDs::BoomYOffset));
+		boom->add_reference_to_child_object(static_cast<std::uint16_t>(ImplementDDOPObjectIDs::BoomZOffset));
+		boom->add_reference_to_child_object(static_cast<std::uint16_t>(ImplementDDOPObjectIDs::ActualWorkingWidth));
+		boom->add_reference_to_child_object(static_cast<std::uint16_t>(ImplementDDOPObjectIDs::ActualCondensedWorkingState1To16));
+		boom->add_reference_to_child_object(static_cast<std::uint16_t>(ImplementDDOPObjectIDs::SetpointCondensedWorkingState1To16));
+		boom->add_reference_to_child_object(static_cast<std::uint16_t>(ImplementDDOPObjectIDs::SectionControlState));
+
+		for (std::size_t i = 0; i < NUMBER_SECTIONS; i++)
```
GwnDaan, line 162:
```suggestion
		for (std::uint8_t i = 0; i < NUMBER_SECTIONS; i++)
```

examples/task_controller_client/section_control_implement_sim.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+			case static_cast<std::uint16_t>(isobus::DataDescriptionIndex::ActualVolumeContent):
+			{
+				value = 3785000; // 1000 Gal in ml;
+			}
+			break;
+
+			case static_cast<std::uint16_t>(isobus::DataDescriptionIndex::SectionControlState):
+			{
+				value = 1; // AUTO mode
```
GwnDaan, line 202: Not sure what this does, but could this be related to `sim->get_section_control_state`?

examples/task_controller_client/section_control_implement_sim.cpp
```diff
@@ -0,0 +1,318 @@
...
+			}
+			break;
+
+			case static_cast<std::uint16_t>(isobus::DataDescriptionIndex::PrescriptionControlState):
+			{
+				value = 1; // AUTO mode
```
GwnDaan, line 208: Same here? `sim->get_prescription_control_state`

examples/task_controller_client/section_control_implement_sim.cpp
```diff
@@ -0,0 +1,318 @@
...
+			break;
+
+			default:
+			{
+				value = 0;
+			}
+			break;
+		}
+	}
+	return true; // You would want to return false in a real controller if a value could not be retrieved
```
GwnDaan, line 259: Kind of curious here? Why can't we return false in our example? How would a real controller be different from this one except for returning real values?

hardware_integration/src/virtual_can_plugin.cpp
```diff
@@ -86,3 +86,9 @@ bool VirtualCANPlugin::read_frame(isobus::HardwareInterfaceCANFrame &canFrame)
 	}
 	return false;
 }
+
+bool VirtualCANPlugin::get_queue_empty() const
+{
+	const std::unique_lock<std::mutex> lock(mutex);
```
GwnDaan, line 92:
```suggestion
	const std::lock_guard<std::mutex> lock(mutex);
```

hardware_integration/src/virtual_can_plugin.cpp
```diff
@@ -86,3 +86,9 @@ bool VirtualCANPlugin::read_frame(isobus::HardwareInterfaceCANFrame &canFrame)
...
+	return (0 == ourDevice->queue.size());
```
GwnDaan, line 93:
```suggestion
	return ourDevice->queue.empty();
```

isobus/include/isobus/isobus/isobus_device_descriptor_object_pool.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+#include "isobus/isobus/isobus_task_controller_client_objects.hpp"
+
+#include <memory>
+
+namespace isobus
+{
+	/// @brief Defines a device descriptor object pool
+	/// @details This class can be used to build up a task conroller DDOP by adding objects to it
+	/// in a heierarchy, then calling generate_binary_object_pool to get the object pool in
```
GwnDaan, line 21:
```suggestion
	/// @details This class can be used to build up a task controller DDOP by adding objects to it
	/// in a hierarchy, then calling generate_binary_object_pool to get the object pool in
```

isobus/include/isobus/isobus/isobus_device_descriptor_object_pool.hpp
```diff
@@ -0,0 +1,142 @@
...
+		/// @param[in] deviceElementDesignator Descriptive text for the object, UTF-8, 32-128 chars max depending on TC version
+		/// @param[in] deviceElementNumber The Element number for process data variable	addressing
+		/// @param[in] parentObjectID Object ID of parent DeviceElementObject or DeviceObject in order to establish a hierarchical order of DeviceElements
+		/// @param[in] deviceEelementType The type of element, such as "device" or "bin"
+		/// @param[in] uniqueID The object ID of the object. Must be unique in the DDOP.
+		/// @returns `true` if the object was added to the DDOP, `false` if the object cannot be added (duplicate or some other error)
+		bool add_device_element(std::string deviceElementDesignator,
+		                        std::uint16_t deviceElementNumber,
+		                        std::uint16_t parentObjectID,
+		                        task_controller_object::DeviceElementObject::Type deviceEelementType,
```
GwnDaan, line 65:
```suggestion
		                        task_controller_object::DeviceElementObject::Type deviceElementType,
```

isobus/include/isobus/isobus/isobus_device_descriptor_object_pool.hpp
```diff
@@ -0,0 +1,142 @@
...
+		                                   std::uint16_t uniqueID);
+
+		/// Constructs a binary DDOP using the objects that were previously added
+		/// @param[in,out] resultantPool The binary representation of the DDOP, or an empty vector if this function returns false
+		/// @returns `true` if the object pool was generated and is valid, otherwise `false`.
+		bool generate_binary_object_pool(std::vector<std::uint8_t> &resultantPool);
+
+		/// @brief Gets an object from the DDOP that corresponds to a certain object ID
+		/// @returns Pointer to the object matching the provided ID, or nullptr if no match was found
+		task_controller_object::Object *get_object_by_id(std::uint16_t objectID);
```
GwnDaan, line 116: Might be safer to return an `weak_ptr` or `shared_ptr`? Might be worth it just to be sure we never end up with dangling pointers, even though it would require refactoring objectList to use `shared_ptr` instead. What do you think?

isobus/src/isobus_task_controller_client.cpp
```diff
@@ -0,0 +1,1711 @@
...
+		measurementMaximumThresholdCommands.clear();
+		measurementOnChangeThresholdCommands.clear();
+	}
+
+	void TaskControllerClient::process_queued_commands()
+	{
+		std::lock_guard<std::mutex> lock(clientMutex);
+		bool transmitSuccessful = true;
+
+		while ((0 != queuedValueRequests.size()) && transmitSuccessful)
```
GwnDaan, line 623:
```suggestion
		while ((!queuedValueRequests.empty()) && transmitSuccessful)
```

isobus/src/isobus_task_controller_client.cpp
```diff
@@ -0,0 +1,1711 @@
...
+				std::uint32_t newValue = 0;
+				if (currentCallback.callback(currentRequest.elementNumber, currentRequest.ddi, newValue, currentCallback.parent))
+				{
+					transmitSuccessful = send_value_command(currentRequest.elementNumber, currentRequest.ddi, newValue);
+					break;
+				}
+			}
+			queuedValueRequests.pop_front();
+		}
+		while ((0 != queuedValueCommands.size()) && transmitSuccessful)
```
GwnDaan, line 638:
```suggestion
		while ((!queuedValueCommands.empty()) && transmitSuccessful)
```

isobus/src/isobus_task_controller_client_objects.cpp
```diff
@@ -0,0 +1,472 @@
...
+				else
+				{
+					retVal.push_back(' ');
+				}
+			}
+			//retVal.push_back(static_cast<std::uint8_t>(extendedStructureLabel.size()));
+			//for (std::size_t i = 0; i < extendedStructureLabel.size(); i++)
+			//{
+			//	retVal.push_back(extendedStructureLabel[i]);
+			//}
```
GwnDaan, line 130: Better to remove this commented out code I think?

utility/include/isobus/utility/platform_endianness.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+namespace isobus
+{
+	/// @brief Returns if the platform is little endian
+	/// @returns `true` if the platform is little endian, otherwise false
+	bool isLittleEndian();
+
+	/// @brief Returns if the platform is big endian
+	/// @returns `true` if the platform is big endian, otherwise false
+	bool isBigEndian();
```
GwnDaan, line 21: Functions should be snake_case

GwnDaan (changes requested):
Good job! I can tell a lot of time and effort went into this haha. I didn't have the time to look at the +6000 newly added lines thoroughly 😅 , but I did have some small points of interest while scanning over it. Love to hear your thoughts on them

examples/task_controller_client/section_control_implement_sim.cpp
ad3154: Yeah I totally made those functions and forgot to use them, haha

isobus/src/isobus_task_controller_client_objects.cpp
ad3154: Oh, this reminds me I need to deal with the extended structure label for version4+, I can clean that up

isobus/include/isobus/isobus/isobus_device_descriptor_object_pool.hpp
ad3154: Seems like a smart idea, I can clean this up

examples/task_controller_client/section_control_implement_sim.cpp
ad3154: Oh, it's just shorthand for returning 0 for values handled by the switch's default statement. The actual use of the return type is for the TC client to know if it needs to keep calling more callbacks to search for one that can satisfy the element number + DDI combination it needs. But in the example this is the only value command callback, so we always want to return true. Maybe that isn't clear enough... 😣

examples/task_controller_client/main.cpp
```diff
@@ -0,0 +1,129 @@
...
+
+	while (running)
+	{
+		if (!tcClientStarted)
+		{
+			if (rateController.create_ddop(myDDOP, TestInternalECU->get_NAME()))
+			{
+				TestTCClient->configure(myDDOP, 1, NUMBER_OF_SECTIONS_TO_CREATE, 1, true, false, true, false, true);
+				TestTCClient->add_request_value_callback(SectionControlImplementSimulator::request_value_command_callback, &rateController);
+				TestTCClient->add_value_command_callback(SectionControlImplementSimulator::value_command_callback, &rateController);
```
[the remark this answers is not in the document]
ad3154: Perhaps, but I think then it would make sense to take `weak_ptr<void>` instead of `void *` which maybe we'd want to do across the whole project at one time if we choose to go that route? I guess I'd punt on that if possible

isobus/src/isobus_task_controller_client_objects.cpp
ad3154: Added version 4 support in 3b09683f4c91639846d09444363652f1d044b037 which properly uses this code block

examples/task_controller_client/section_control_implement_sim.cpp
ad3154: Clarified the comments to hopefully make it more clear to a reader what's happening there.

examples/task_controller_client/section_control_implement_sim.cpp
```diff
@@ -139,10 +139,10 @@ bool SectionControlImplementSimulator::create_ddop(std::shared_ptr<isobus::Devic
...
-		auto boom = reinterpret_cast<isobus::task_controller_object::DeviceElementObject *>(poolToPopulate->get_object_by_id(static_cast<std::uint16_t>(ImplementDDOPObjectIDs::SprayBoom)));
-		auto product = reinterpret_cast<isobus::task_controller_object::DeviceElementObject *>(poolToPopulate->get_object_by_id(static_cast<std::uint16_t>(ImplementDDOPObjectIDs::LiquidProduct)));
+		auto sprayer = std::static_pointer_cast<isobus::task_controller_object::DeviceElementObject>(poolToPopulate->get_object_by_id(static_cast<std::uint16_t>(ImplementDDOPObjectIDs::MainDeviceElement)).lock());
+		auto connector = std::static_pointer_cast<isobus::task_controller_object::DeviceElementObject>(poolToPopulate->get_object_by_id(static_cast<std::uint16_t>(ImplementDDOPObjectIDs::Connector)).lock());
+		auto boom = std::static_pointer_cast<isobus::task_controller_object::DeviceElementObject>(poolToPopulate->get_object_by_id(static_cast<std::uint16_t>(ImplementDDOPObjectIDs::SprayBoom)).lock());
+		auto product = std::static_pointer_cast<isobus::task_controller_object::DeviceElementObject>(poolToPopulate->get_object_by_id(static_cast<std::uint16_t>(ImplementDDOPObjectIDs::LiquidProduct)).lock());
```
GwnDaan, line 145: Actually, it might be better to return a `shared_ptr` instead for the `get_object_by_id` maybe? Since you'd have to check if `.lock()`is actually returning something everytime otherwise. And as far as I can tell with a task controller object, you don't have to worry about it being in a 'valid' state since it's standalone I imagine?

GwnDaan (changes requested):

examples/task_controller_client/section_control_implement_sim.cpp
ad3154: Well, you'll still have to check the shared_ptr isn't nullptr, as it would return an empty shared_ptr in the case that an object is not found. But, it would make for less use of `lock` when it will almost always be valid

GwnDaan: Ah I see, that's right. If only we were using Rust where null/nullptr doesn't exist haha... 😛

ad3154: Do you still think I should change it, just to avoid `lock`? or leave it?

GwnDaan: Uhm, I don't see any cons of changing from weak to shared. Avoiding `lock` here seems to me like less confusion can occur, so yea I think changing it is still worth it

ad3154: OK! I will make that change, thanks for your feedback 😁

GwnDaan (dismissed):

Issue #161, opened 2023-01-28
Using singleton static CANNetworkManager vs multiple instances


GwnDaan:
While refactoring some unit tests I saw that we always refer to the one singleton:
```cpp
static CANNetworkManager CANNetwork; ///< Static singleton of the one network manager. Use this to access stack functionality.
```

I wonder if using only one CANNetwork is the optimal approach? I find it a bit hard to gasp what it stands for and how it's used. Namely the automatic address claiming when you initialize a control function seems like "magic". You never have to explicitly call the CANNetwork to claim/register this CF. This might be confusing to new users of the stack?
It also handles multiple can channels at once from what I can tell, which I feel like can be separated in different objects? I think separating each 'device' on the stack to use it's own manager would allow us to write unit tests more easily. 

Though there probably was a good reason to write it like this which I would love to hear and think about! 😄

ad3154:
I have had parts this particular discussion with some peers before and would be happy to lay out my thoughts. I am not perfect, and I want the stack to be as friendly as possible to relative newbies with CAN, but powerful to experienced users, so suggestions are always good.

### The Singleton

The network manager singleton is not my favorite - I would be OK with splitting it up 1 per channel and allowing users to instantiate them. It is the current way for a reason that doesn't matter so much now. Initially when I wrote the stack, I wanted the protocol classes to be dynamic. Basically, I wanted to be able to compile in or out single protocols without `ifdef`. For example, if you have a project that does not use Fast Packet, it would be cool if you could simply not compile the cpp file for fast packet, and the stack would still build, just without Fast Packet. In the end though, this is what causes the pain that we went through with that `-whole-archive` stuff, so it's maybe not worth it, and we've started to go away from that anyways with TP and ETP being now housed in side the network manager.

The current way also simplifies the hardware layer a fair bit. There's just one receive queue, and all messages are processed by the stack by a single thread in order of reception. If we break it up it may involve some additional complexity, but hopefully not too much? Since the network manager updates via a callback from the hardware layer, I think we could register multiple of them with relatively few issues.

So yeah, I guess I did it for a reason, but that reason is now not so relevant. If we do end up re-structuring it, I am sure there will be some issues with how the protocols expect there to be a singleton, so maybe each will need to hold a reference to it's parent network or something, but those are challenges we could overcome. Plus, I do agree we would get more flexibility in unit tests if we split it up. That especially would be nice.

### Address Claiming

Sorry in advance for the wall of text.

[...]

PR #196, opened 2023-02-28
Actually pass the parent pointer to periodic updates.

Y-Less:
`add_can_lib_update_callback`, like many other callback methods, takes a `parentPointer` and stores said pointer in `CanLibUpdateCallbackInfo`.  However, unlike basically every other example of callbacks that I could find this meta-data is not actually passed to the callback itself, in fact the only place it seems to be used is in `remove_can_lib_update_callback` to confirm that you're removing the correct instance.

This updates the signatures and calls to actually pass the parent pointer to the callbacks themselves.

### DIFF
--- a/examples/diagnostic_protocol/main.cpp
+++ b/examples/diagnostic_protocol/main.cpp
@@ -20,7 +20,7 @@ void signal_handler(int)
 	running = false;
 }
 
-void update_CAN_network(void *parentPointer)
+void update_CAN_network(void *)
 {
 	isobus::CANNetworkManager::CANNetwork.update();
 }
--- a/examples/diagnostic_protocol/main.cpp
+++ b/examples/diagnostic_protocol/main.cpp
@@ -20,7 +20,7 @@ void signal_handler(int)
 	running = false;
 }
 
-void update_CAN_network()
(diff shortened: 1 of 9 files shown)


ad3154:
Hey there! Welcome to the project, and thanks for your contribution!

I think this is good for consistency, especially if we ever refactor the network manager to be a multiple instance object like was described in #161, in which case we will need to make this change anyways.

One thing I will note is that many compilers will complain about passing a value that is not used. One way to avoid this is to just omit the variable name from the consumer function. For example, if you change `void update_CAN_network(void *parentPointer)` to `void update_CAN_network(void *)` in the example files, most compilers will be happy and not generate a warning at compile time.

If you don't mind going back through and changing that in the example files, that would be most appreciated, and I'd be happy to approve your PR and get it merged in!

Thanks,

Adrian

ad3154:
address_claim_tests.cpp Also appears to be failing - just need to update that file similar to how you updated the example files and it'll be happy again.

Y-Less:
Done.  I'm not sure how I missed that `add_can_lib_update_callback` instance in the test when I did the initial changes.  Oh well...

ad3154 (approved):

PR #203, opened 2023-03-08
[ETP]: Fix unable to transfer specific big data lengths

GwnDaan:
This PR solves an issue where an object pool with some specfic size wasn't able to get transferred due to `numberBytesLeft` overflowing to exactly `0` for some packet. An example would be: an object pool of size 65543 bytes. Here after the first ETP data packet of is being sent, the `numberBytesLeft` would be 65536. This overflows to 0 as an uint16_t would only be able to contain the number 65535. A fix would be to change it to an uint32 instead as this PR proposes.

Related to Sonar issue: https://sonarcloud.io/project/issues?issues=AYXrWJ0Khx23eJJBSqGV&open=AYXrWJ0Khx23eJJBSqGV&id=ad3154_ISO11783-CAN-Stack

### DIFF
--- a/isobus/src/can_extended_transport_protocol.cpp
+++ b/isobus/src/can_extended_transport_protocol.cpp
@@ -695,7 +695,7 @@ namespace isobus
 										0xFF,
 										0xFF
 									};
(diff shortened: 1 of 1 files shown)


ad3154 (approved):
Great catch!

Issue #206, opened 2023-03-09
Add internal control function filtering to PGN callbacks


Y-Less:
First - is there some better support forum for basic usage questions?  Forums/discord/e-mail etc?  I know that this isn't what Github issues are meant to be used for sorry.

Secondly.  I'm a bit lost as to the purpose of internal functions when using partnered functions.  According to the documentation you pass an internal function as the source when sending messages to partnered functions.  But then responses from the partner get sent to the partnered function directly after using `add_parameter_group_number_callback`.  So it seems like the internal function is the original source, and the partnered function is both the destination and the response receiver.  Why is the internal function (which was the source of the original message) not the recipient of the reply; why is it not even referenced in the response callback; and would sending a new message from the response callback need to acquire that internal function again from somewhere?

In short - why is it used for outbound messages, but ignored for their responses?

ad3154:
You could [start a GitHub discussion](https://github.com/ad3154/Isobus-plus-plus/discussions)? The community is pretty small right now, so we're mostly just here on GitHub right now. I think I might be able to convert issues to discussions as well if they are more like clarification stuff.

So, you are right in thinking that the internal control function is source to send messages from, and the partner is a destination. This concept is reversed though when receiving destination specific messages from a partner.
You can also of course pass in `nullptr` as the partner to send a broadcast (and indeed this is required for some PGNs).

The "response receiver" as you call it is whatever callback you add via the partner class (or using `add_global_parameter_group_number_callback`), and generally the context is completely up to you, it doesn't have to be tied to the partner in any way really except in the implicit sense that you are telling the network manager that you care about messages from that partner. The CAN message class you get from the network manager actually contains pointers to both as well! The destination control function (`get_destination_control_function`) contains a pointer to your internal control function which is the destination of that message. Though, if you want to use it as in internal control function again in there, you'd want to check the type and cast it appropriately.

[...]

PR #208, opened 2023-03-10
Add optional destination function filtering for partner control function responses.

Y-Less:
Basically implementing the outcome of #206, so you can listen for responses only to certain source internal functions.

### DIFF
--- a/isobus/include/isobus/isobus/can_network_manager.hpp
+++ b/isobus/include/isobus/isobus/can_network_manager.hpp
@@ -58,32 +58,44 @@ namespace isobus
 		void add_control_function(std::uint8_t CANPort, ControlFunction *newControlFunction, std::uint8_t CFAddress, CANLibBadge<AddressClaimStateMachine>);
 
 		/// @brief This is how you register a callback for any PGN destined for the global address (0xFF)
+		/// @details You may pass a destination `InternalControlFunction`, which will filter for only those messages
+		/// that target this source/destination pair (see https://github.com/ad3154/Isobus-plus-plus/issues/206).
 		/// @param[in] parameterGroupNumber The PGN you want to register for
 		/// @param[in] callback The callback that will be called when parameterGroupNumber is recieved from the global address (0xFF)
 		/// @param[in] parent A generic context variable that helps identify what object the callback is destined for. Can be nullptr if you don't want to use it.
-		void add_global_parameter_group_number_callback(std::uint32_t parameterGroupNumber, CANLibCallback callback, void *parent);
+		/// @param[in] destinationFunction An optional internal function destination to filter messages by
+		void add_global_parameter_group_number_callback(std::uint32_t parameterGroupNumber, CANLibCallback callback, void *parent, InternalControlFunction *destinationFunction = nullptr);
 
 		/// @brief This is how you remove a callback for any PGN destined for the global address (0xFF)
+		/// @details You may pass a destination `InternalControlFunction`, which will filter for only those messages
+		/// that target this source/destination pair (see https://github.com/ad3154/Isobus-plus-plus/issues/206).
 		/// @param[in] parameterGroupNumber The PGN of the callback to remove
 		/// @param[in] callback The callback that will be removed
 		/// @param[in] parent A generic context variable that helps identify what object the callback was destined for
-		void remove_global_parameter_group_number_callback(std::uint32_t parameterGroupNumber, CANLibCallback callback, void *parent);
+		/// @param[in] destinationFunction An optional internal function destination to filter messages by
+		void remove_global_parameter_group_number_callback(std::uint32_t parameterGroupNumber, CANLibCallback callback, void *parent, InternalControlFunction *destinationFunction = nullptr);
 
 		/// @brief Returns the number of global PGN callbacks that have been registered with the network manager
 		/// @returns The number of global PGN callbacks that have been registered with the network manager
 		std::uint32_t get_number_global_parameter_group_number_callbacks() const;
 
 		/// @brief Registers a callback for ANY control function sending the associated PGN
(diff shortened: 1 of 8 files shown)


isobus/include/isobus/isobus/can_network_manager.hpp
```diff
@@ -58,32 +58,44 @@ namespace isobus
...
 
 		/// @brief This is how you register a callback for any PGN destined for the global address (0xFF)
+		/// @details You may pass a destination `InternalControlFunction`, which will filter for only those messages
+		/// that target this source/destination pair (see https://github.com/ad3154/Isobus-plus-plus/issues/206).
 		/// @param[in] parameterGroupNumber The PGN you want to register for
 		/// @param[in] callback The callback that will be called when parameterGroupNumber is recieved from the global address (0xFF)
 		/// @param[in] parent A generic context variable that helps identify what object the callback is destined for. Can be nullptr if you don't want to use it.
-		void add_global_parameter_group_number_callback(std::uint32_t parameterGroupNumber, CANLibCallback callback, void *parent);
+		/// @param[in] destinationFunction An optional internal function destination to filter messages by
+		void add_global_parameter_group_number_callback(std::uint32_t parameterGroupNumber, CANLibCallback callback, void *parent, InternalControlFunction *destinationFunction = nullptr);
```
ad3154, line 67: Probably don't need it on `add_global_parameter_group_number_callback` as an internal control function will never be the destination (the destination is always a broadcast)

isobus/include/isobus/isobus/can_network_manager.hpp
```diff
@@ -58,32 +58,44 @@ namespace isobus
...
 
 		/// @brief Registers a callback for ANY control function sending the associated PGN
+		/// @details You may pass a destination `InternalControlFunction`, which will filter for only those messages
+		/// that target this source/destination pair (see https://github.com/ad3154/Isobus-plus-plus/issues/206).
 		/// @param[in] parameterGroupNumber The PGN you want to register for
 		/// @param[in] callback The callback that will be called when parameterGroupNumber is recieved from any control function
 		/// @param[in] parent A generic context variable that helps identify what object the callback is destined for. Can be nullptr if you don't want to use it.
-		void add_any_control_function_parameter_group_number_callback(std::uint32_t parameterGroupNumber, CANLibCallback callback, void *parent);
+		/// @param[in] destinationFunction An optional internal function destination to filter messages by
+		void add_any_control_function_parameter_group_number_callback(std::uint32_t parameterGroupNumber, CANLibCallback callback, void *parent, InternalControlFunction *destinationFunction = nullptr);
```
ad3154, line 89: Probably don't need it on the `any_control_function` methods, as those will never check the internal control function as the destination. They are designed to accept messages from any CF. If you're adding it to some of these just to satisfy the constructor for the `ParameterGroupNumberCallbackData` it might make more sense to create a constructor for that instead that just sets the ICF to nullptr.

isobus/include/isobus/isobus/can_network_manager.hpp
```diff
@@ -133,18 +145,24 @@ namespace isobus
...
 		/// @brief Adds a PGN callback for a protocol class
+		/// @details You may pass a destination `InternalControlFunction`, which will filter for only those messages
+		/// that target this source/destination pair (see https://github.com/ad3154/Isobus-plus-plus/issues/206).
 		/// @param[in] parameterGroupNumber The PGN to register for
 		/// @param[in] callback The callback to call when the PGN is received
 		/// @param[in] parentPointer A generic context variable that helps identify what object the callback was destined for
+		/// @param[in] destinationFunction An optional internal function destination to filter messages by
 		/// @returns `true` if the callback was added, otherwise `false`
-		bool add_protocol_parameter_group_number_callback(std::uint32_t parameterGroupNumber, CANLibCallback callback, void *parentPointer);
+		bool add_protocol_parameter_group_number_callback(std::uint32_t parameterGroupNumber, CANLibCallback callback, void *parentPointer, InternalControlFunction *destinationFunction = nullptr);
```
ad3154, line 155: Probably don't need it for the protocols? They bypass a lot of checks that the partner control function callbacks do and manage multiple internal control functions in the same callback anyways already.

isobus/include/isobus/isobus/nmea2000_fast_packet_protocol.hpp
```diff
@@ -46,16 +46,22 @@ namespace isobus
...
 
 		/// @brief Similar to add_parameter_group_number_callback but tells the stack to parse those PGNs as Fast Packet
+		/// @details You may pass a destination `InternalControlFunction`, which will filter for only those messages
+		/// that target this source/destination pair (see https://github.com/ad3154/Isobus-plus-plus/issues/206).
 		/// @param[in] parameterGroupNumber The PGN to parse as fast packet
 		/// @param[in] callback The callback that the stack will call when a matching message is received
 		/// @param[in] parent Generic context variable
-		void register_multipacket_message_callback(std::uint32_t parameterGroupNumber, CANLibCallback callback, void *parent);
+		/// @param[in] destinationFunction An optional internal function destination to filter messages by
+		void register_multipacket_message_callback(std::uint32_t parameterGroupNumber, CANLibCallback callback, void *parent, InternalControlFunction *destinationFunction = nullptr);
```
ad3154, line 55: Good plan to add it to fast packet, I probably would have forgotten about that. We probably also need to add some code to its cpp file to handle this new parameter here:
image

ad3154 (changes requested):
Just a few minor comments. Thanks for adding this.

One additional note is that you can fix the failing clang-format check by running `find . -iname *.hpp -o -iname *.cpp| xargs clang-format -i` in linux or WSL if you have it installed `sudo apt install clang-format`. Otherwise on Windows you would need to install LLVM to get clang-format and run it from command prompt or powershell

ad3154:
Added in #250

PR #209, opened 2023-03-10
Packet builder

Y-Less:
Another draft PR for comments, since it is a larger one.  I've basically written a class to do all the maths and position tracking for building up packets with lots of compressed data in.

The current code has lots of instances of code to build packets like:

```cpp
data[0] = EXTENDED_CONNECTION_ABORT_MULTIPLEXOR;
data[1] = static_cast<std::uint8_t>(reason);
data[2] = 0xFF;
data[3] = 0xFF;
data[4] = 0xFF;
data[5] = static_cast<std::uint8_t>(pgn & 0xFF);
data[6] = static_cast<std::uint8_t>((pgn >> 8) & 0xFF);
data[7] = static_cast<std::uint8_t>((pgn >> 16) & 0xFF);
```

Or deconstruct packets like:

```cpp
auto &data = message->get_data();
parentInterface->languageCommandTimestamp_ms = SystemTiming::get_timestamp_ms();
parentInterface->languageCode.clear();
parentInterface->languageCode.push_back(static_cast<char>(data.at(0)));
parentInterface->languageCode.push_back(static_cast<char>(data.at(1)));
parentInterface->timeFormat = static_cast<TimeFormats>((data.at(2) >> 4) & 0x03);
parentInterface->decimalSymbol = static_cast<DecimalSymbols>((data.at(2) >> 6) & 0x03);
parentInterface->dateFormat = static_cast<DateFormats>(data.at(3));
parentInterface->massUnitSystem = static_cast<MassUnits>(data.at(4) & 0x03);
parentInterface->volumeUnitSystem = static_cast<VolumeUnits>((data.at(4) >> 2) & 0x03);
parentInterface->areaUnitSystem = static_cast<AreaUnits>((data.at(4) >> 4) & 0x03);
parentInterface->distanceUnitSystem = static_cast<DistanceUnits>((data.at(4) >> 6) & 0x03);
parentInterface->genericUnitSystem = static_cast<UnitSystem>(data.at(5) & 0x03);
parentInterface->forceUnitSystem = static_cast<ForceUnits>((data.at(5) >> 2) & 0x03);
parentInterface->pressureUnitSystem = static_cast<PressureUnits>((data.at(5) >> 4) & 0x03);
parentInterface->temperatureUnitSystem = static_cast<TemperatureUnits>((data.at(5) >> 6) & 0x03);
```

This is (IMHO) very complex to read and error-prone, though admittedly it is very explicit in what is happening and where things are being stored.

Anyway, those two examples rewritten with this PR would look something like:

```cpp
isobus::ParameterGroupBuilder builder {};
builder.write<std::uint8_t>(EXTENDED_CONNECTION_ABORT_MULTIPLEXOR);
builder.write(static_cast<std::uint8_t>(reason));
builder.pad(24);
builder.write(pgn, 24);
```

```cpp
isobus::ParameterGroupBuilder builder(message->get_data());
parentInterface->languageCommandTimestamp_ms = SystemTiming::get_timestamp_ms();
builder.read((char *)&parentInterface->languageCode, 16);
builder.skip(4);
builder.read(parentInterface->timeFormat, 2);
builder.read(parentInterface->decimalSymbol, 2);
builder.read(parentInterface->dateFormat);
builder.read(parentInterface->massUnitSystem, 2);
builder.read(parentInterface->volumeUnitSystem, 2);
builder.read(parentInterface->areaUnitSystem, 2);
builder.read(parentInterface->distanceUnitSystem, 2);
builder.read(parentInterface->genericUnitSystem, 2);
builder.read(parentInterface->forceUnitSystem, 2);
builder.read(parentInterface->pressureUnitSystem, 2);
builder.read(parentInterface->temperatureUnitSystem, 2);
```

Might just need a tiny bit of tweaking for types/widths on some of the variables, but otherwise that's complete.

Now I've already mixed up `pad` (for writing) and `skip` (for reading) in my own code, so it might make more sense to have the reading and writing capabilities split in to two different classes, but the core algorithms are there.  Interesting, one thing I noticed is that the spec goes in to some detail about how odd-bit-width data spans byte boundaries, and this code correctly implements that, but then all the PGNs in the spec use padding bits to explicitly avoid those cases.  And thus this code also checks for those common cases where data is byte-aligned or fully contained within a byte to better optimise for those cases.


Y-Less:
I don't know where the best place for this is.  I've not committed any tests (but did run some and am using it locally).  This is mostly just the initial basic algorithm.  And even within that I'm sure some parts could be compressed - there are a few bits of repeated code, especially in the sections handing large data spanning multiple bytes, but honestly I don't think rearranging the code is worth it to save a few lines at the expense of complexity.

Y-Less:
Two big formatting things I've noticed that probably need changing - I've used `#pragma once` instead of include guards, and early returns, neither of which seem tot be used anywhere else in the codebase.

ad3154:
I'll take a look at it and make some comments. Something like this has been requested before, so it could be helpful for people who do not want to do manual shifting and masking. The style is fairly off from the rest of the repo, so a general comment would be to try and match the rest of the repo's style if possible, I'll try and comment some examples.

To answer some of your questions up-front, for `#pragma` generally we lean on the Autosar rule:
image
Since it's compiler specific behavior and is not always even implemented, it should be avoided for a highly cross-platform library like ours.

As for multiple return paths, I like to lean on MISRA's rule:
image

This is a fairly controversial rule in C++ I will admit, as there's nothing "wrong" with early returns. Most of the benefit though is that code will always have a normal code flow top to bottom, there's only 1 place where you need to put a breakpoint to catch the return value of the function (makes debugging easier I would argue, especially on devices that have a maximum number of breakpoints (!!) like the stm32f4 which can only have something like 7 breakpoints), and lastly it helps ensure allocated resources get cleaned up, since you'd either need to copy the lines of code that do that for each return statement, or do it once at the bottom (though, a pure RAII approach mitigates this). The main downside to a single return path as I see it is that there is additional indenting in the code due to making those early returns into more highly nested conditionals, but I have rarely found that a compelling reason to lose the benefits while debugging. I don't think I'd prevent a merge due to an early return, but I don't prefer them, especially in a library marketed towards an industry that will skew towards that Autosar/Misra side of things.

isobus/include/isobus/isobus/can_parameter_group_builder.hpp
```diff
@@ -0,0 +1,503 @@
...
+			if (space)
+			{
+				// Bits left to mask.
+				*data = *data & (255 >> space);
+			}
+			return true;
+		}
+
+	public:
+		ParameterGroupBuilder() :
```
ad3154, line 224: Can probably be `= default`

isobus/include/isobus/isobus/can_parameter_group_builder.hpp
```diff
@@ -0,0 +1,503 @@
...
+		std::size_t get_write_byte_offset() const
+		{
+			// Which byte to write to.
+			return writeOffset / 8;
+		}
+
+		std::size_t get_write_bit_offset() const
+		{
+			// Which bit to write to in the current byte.
+			return writeOffset % 8;
```
ad3154, line 31: For non-templated functions, prefer putting the implementation in a .cpp file. This will help a number of things such as allowing our static analysis to better check the file, as it will not really be checked right now since the single header is included nowhere.

isobus/include/isobus/isobus/can_parameter_group_builder.hpp
```diff
@@ -0,0 +1,503 @@
...
+					++data;
+					*data = 0;
+				}
+				else
+				{
+					bits -= space;
+					// Perfectly matched.
+					remaining = 8;
+					input = 0;
+					space = 8;
```
ad3154, line 208: You use the magic number 8 frequently. If it's just the number of bits in a byte, it might be nicer to get that constant from `#include <limits>` or otherwise define a `static constexpr std::uint8_t` for it

isobus/include/isobus/isobus/can_parameter_group_builder.hpp
```diff
@@ -0,0 +1,503 @@
...
+	public:
+		ParameterGroupBuilder() :
+		  buffer()
+		{
+		}
+
+		ParameterGroupBuilder(std::vector<std::uint8_t> const &data) :
+		  buffer()
+		{
+			buffer = data;
```
ad3154, line 232: I think this makes a copy of the data, which is not ideal. The best way to deal with this would be for the network manager to pass back a weak_ptr or shared_ptr in a CAN message and store that here, but since it's not doing that (yet), there's a lot of overhead. I also don't think we'd want to store a `std::vector<std::uint8_t> &` in the class either, since it could go out of scope in the parent scope... So I guess it's fine for the moment, but it's a fairly heavy memory price. Could refactor it if/when we improve the safety of the messages from the network manager.

isobus/include/isobus/isobus/can_parameter_group_builder.hpp
```diff
@@ -0,0 +1,503 @@
...
+		}
+
+		bool write(char * data, bool includeNull)
+		{
+			return write((std::uint8_t const *)data, includeNull);
+		}
+
+		bool write(std::uint8_t const * data, bool includeNull)
+		{
+			// Base case.  Write each byte separately so they don't get put in little-endian, which
```
ad3154, line 315: Little-endian is the normal way in the standard, so probably meant `big` in this comment

isobus/include/isobus/isobus/can_parameter_group_builder.hpp
```diff
@@ -0,0 +1,503 @@
+#pragma once
+
+#include <vector>
+
+namespace isobus
+{
+	// Note that this class currently only works for packets eight bytes or fewer, because that is
```
ad3154, line 7: Ideally we'd want it to work up to 117440505 bytes to meet the generic use case.

isobus/include/isobus/isobus/can_parameter_group_builder.hpp
```diff
@@ -0,0 +1,503 @@
...
+	// all the use-cases I needed.  The spec has a whole section on how data can span byte
+	// boundaries, and how to align the bits when a data-type that is not a multiple of eight bits
+	// crosses said boundaries.  Then every single parameter group uses padding to avoid those
+	// cases!  The five-bit tractor command types have three bits of padding to make them exactly
+	// one byte.  Why even specify the scheme if it isn't used?
```
ad3154, line 12: They do a good job of padding, that is true, but I have often seen SPNs span bytes in an unaligned fashion in proprietary messaging, so it is more common than you might think

isobus/include/isobus/isobus/can_parameter_group_builder.hpp
```diff
@@ -0,0 +1,503 @@
...
+					++byte;
+				}
+			}
+			return true;
+		}
+
+		template <typename T>
+		bool read(T & data)
+		{
+			return read_bits((std::uint8_t *)&data, sizeof (T) * 8);
```
ad3154, line 382: Casting to `uint8_t *` from T is not endianness friendly, plus it's a C style cast which should not be used. This happens multiple places, and may require a fairly substantial refactor to fully resolve.

isobus/include/isobus/isobus/can_parameter_group_builder.hpp
```diff
@@ -0,0 +1,503 @@
...
+			}
+			data = false;
+			return false;
+		}
+
+		template <>
+		bool read<char *>(char * & data)
+		{
+			// Read until NULL.
+			return read<std::uint8_t*>((std::uint8_t * &)data);
```
ad3154, line 410: It seems like a dangerous assumption that a `uint8_t *` will have null termination here and other places. That's a non-trivial assumption that cannot be enforced. `std::vector<std::uint8_t>` is not a `std::string()::c_str()`

isobus/include/isobus/isobus/can_parameter_group_builder.hpp
```diff
@@ -0,0 +1,503 @@
...
+				// Trying to read too much data.
+				readOffset -= bits;
+				return false;
+			}
+			// Initialise the current destination byte.
+			*data = 0;
+			while (bits)
+			{
+				// Move in the new data.
+				*data = (*data & (255 >> space)) | ((buffer[byte] >> input) << output);
```
ad3154, line 178: 255 in a few places can be replaced with a constant in `<limits>`

ad3154 (changes requested):
I've made a first pass through it. Generally I think the library could use something that is similar to this, but it seems like a number of your code's assumptions have made their way into here, and I have concerns about it not working correctly depending on a machine's endianness. It also isn't quite generic enough to fit a library use-case, up to the ETP protocol size limit.

So, I guess where we go from there is kind of up to you. This kind of thing is an ideal candidate for heavy unit testing since it's easy to provide lots of stateless test cases, so if you want to keep working on it and improving it, and if you can write unit tests for it, I think we could revisit adding it in. Or, if you think it's meeting your needs in your software, you are of course welcome to keep using it in your fork.

It can be tough to convey emotion through a GitHub PR review, so I just wanted to say that I do appreciate all your PR submissions, including this one. I probably spent at least a good hour doing this review (it takes time to read the code to come up-to-speed on what it's doing), which I wouldn't do if I didn't think it was valuable.

isobus/include/isobus/isobus/can_parameter_group_builder.hpp
Y-Less: Ah, that's an old comment I forgot to remove.  It did only do eight bytes, but I removed that restriction already.

isobus/include/isobus/isobus/can_parameter_group_builder.hpp
Y-Less: This is more a rant than anything.  the code does handle byte spans correctly (at least I believe it does).

isobus/include/isobus/isobus/can_parameter_group_builder.hpp
Y-Less: Yes it is, and I know how that bad assumption sneaked in (I did a find/replace on `unsigned char` to `uint8_t`).  TBH I was already unhappy with the way strings were handled in here, so I'll have to revisit that entire part.

Y-Less:
Thanks for the comments.  I realised it was very far from the rest of the code style, sorry about that.  While I'm happy to work on the code style changes, are there any more generic comments about the general approach/API/etc?

> It can be tough to convey emotion through a GitHub PR review, so I just wanted to say that I do appreciate all your PR submissions, including this one. I probably spent at least a good hour doing this review (it takes time to read the code to come up-to-speed on what it's doing), which I wouldn't do if I didn't think it was valuable.

And I appreciate your time and effort for reviews.  Don't worry, I didn't infer any ill intent/bad emotions at all; I've been on your side of code reviews plenty of times before.

ad3154:
> are there any more generic comments about the general approach/API/etc?
 
 Overall I think the API is pretty intuitive, you build up a packet sequentially and likewise disassemble it sequentially, and it hides the masking and shifting well. "pad"and "skip" and even "read" and "write" might be not the most clear on what's going on but the core of it is all there for sure.

PR #211, opened 2023-03-12
[Core]: Add event dispatcher as more maintainable way of handling callbacks

GwnDaan:
## What's new
With the new `EventDispatcher` class we should be able to implement listeners / callbacks in a more maintainable way. Why more maintainable?:
- The listeners automatically clean themselves up, so you don't have to worry about unregistering them
- There is a lot less code to write to when using callbacks in the library itself
- All types can be inserted due to the use of the template system
- No more typedeffed raw pointers to functions, which makes it a lot easier to register a callback as you don't have to look up the exact types you need to use.

## The catch
If this sounds too good to be true, it probably is... The compiled `libIsobus.a` file size went from 13.5 MB to 15.8MB when compiling this PR compiled to the main branch. Not sure if this is game breaker?

Why this PR? I only changed the VT event callbacks to use the new system, and the difference is 370 lines less code to maintain (When not counting the newly added `utility/include/isobus/utility/event_dispatcher.hpp` and `test/event_dispatcher_tests.cpp` files for additions.)

Is this worth it? I don't know... Love to hear your thoughts

### DIFF
--- a/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
+++ b/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
@@ -295,51 +295,52 @@ namespace isobus
 		/// @brief A struct for storing information of a VT key input event
 		struct VTKeyEvent
 		{
-			KeyActivationCode keyEvent; ///< The key event
-			std::uint8_t keyNumber; ///< The key number
+			VirtualTerminalClient *parentPointer; ///< A pointer to the parent VT client
 			std::uint16_t objectID; ///< The object ID
 			std::uint16_t parentObjectID; ///< The parent object ID
-			VirtualTerminalClient *parentPointer; ///< A pointer to the parent VT client
+			std::uint8_t keyNumber; ///< The key number
+			KeyActivationCode keyEvent; ///< The key event
 		};
 
 		/// @brief A struct for storing information of a VT pointing event
 		struct VTPointingEvent
 		{
-			KeyActivationCode keyEvent; ///< The key event
+			VirtualTerminalClient *parentPointer; ///< A pointer to the parent VT client
 			std::uint16_t xPosition; ///< The x position
 			std::uint16_t yPosition; ///< The y position
 			std::uint16_t parentObjectID; ///< The parent object ID
-			VirtualTerminalClient *parentPointer; ///< A pointer to the parent VT client
+			KeyActivationCode keyEvent; ///< The key event
 		};
 
 		/// @brief A struct for storing information of a VT input object selection event
 		struct VTSelectInputObjectEvent
 		{
+			VirtualTerminalClient *parentPointer; ///< A pointer to the parent VT client
 			std::uint16_t objectID; ///< The object ID
 			bool objectSelected; ///< Whether the object is selected
 			bool objectOpenForInput; ///< Whether the object is open for input
-			VirtualTerminalClient *parentPointer; ///< A pointer to the parent VT client
 		};
 
 		/// @brief A struct for storing information of a VT ESC message event
 		struct VTESCMessageEvent
 		{
+			VirtualTerminalClient *parentPointer; ///< A pointer to the parent VT client
 			std::uint16_t objectID; ///< The object ID
 			ESCMessageErrorCode errorCode; ///< The error code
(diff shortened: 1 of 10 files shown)


ad3154:
> file size went from 13.5 MB to 15.8MB

Jeez, classic `std::function` 😂

This does seem _very_ nice to have. I do also wonder about that file size.... 
I am leaning towards thinking it's fine, but let me try an experiment compiling this using arm compiler 6 to see what the size is before and after. I do worry about the size of the library getting too big for some smaller platforms, so I should probably test it...

GwnDaan:
Not sure if Sonar is right on this bug? According to https://en.cppreference.com/w/cpp/utility/forward the use of `std::forward` looks just valid in our case...

```cpp
template<class T, class... U> std::unique_ptr<T> make_unique2(U&&... u) 
{ 
return std::unique_ptr<T>(new T(std::forward<U>(u)...)); 
}  
```

ad3154:
I did some testing and I couldn't build on ARM compiler anyways because it doesn't have std::thread, so, I guess it doesn't really matter that much at this stage. I think moving forward with an std::function methodology will be nicer in more ways than one, so I think it's OK to keep going with this approach. If we want this to work on something with an RTOS it may be a bigger effort than I anticipated.

GwnDaan:
@ad3154 I think this PR is ready for your review :). I have another upcoming PR with [hardware optimizations](https://github.com/GwnDaan/ISO11783-CAN-Stack/tree/optimize-hardware-interface) ready, though it uses this PR as it's base haha

isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
```diff
@@ -290,146 +292,170 @@ namespace isobus
...
-
-		/// @brief Allows you to register for a callback when the active mask is changed
-		/// @details The VT sends this whenever there are missing object references or errors in the mask.
-		/// @param[in] value The callback to register
-		void register_vt_change_active_mask_event_callback(VTChangeActiveMaskCallback value);
+		/// @brief A struct for storing information of a VT key input event
+		struct VTKeyEvent
+		{
+			KeyActivationCode keyEvent; ///< The key event
+			std::uint8_t keyNumber; ///< The key number
```
ad3154, line 299: This is like, the most tiny, minor nit, but for all these structs, putting the uint8_t last may reduce the amount of memory padding that is needed for this struct. Likewise, putting the pointer first will also optimize this a little.

GwnDaan: Hmm, I think https://github.com/ad3154/Isobus-plus-plus/pull/211/commits/b2040cdeb98f58b46b8c1fd09113e85b0562397c resolves this? I am new to this memory padding stuff, and online they are rather vage about it. It seems like it's not really a standardized thing and the implementation differs across compilers and platforms? 

Though changing the order does make it a little harder to read I feel like, but since the structs are small and thus fast to read it's fine I think

ad3154: Yeah, good enough I think. The padding bytes depend a fair bit on architecture, mattering most to align memory addressing. So like, if your memory is byte addressable it doesn't matter much, but if it's word addressed and you have 32 bit words, then putting a u8 before a u32 may create 3 bytes of padding, for example, or the compiler (or even the microcode) could read both words and stitch them together, but that is not very efficient. I would hope that a smart compiler would re-order the struct in memory if optimizations are enabled and it thinks it can do so safely, but you never know.

ad3154: Looks like maybe the build is mad now due to the initializer list construction order changing

GwnDaan: Ohh yea, I forgot that. The build error is fixed in the latest commit🙉

ad3154:
We'll probably need to go through and update the tutorial docs for this change. In general we'll also want to convert TC to use this dispatcher as well, and maybe some other things. I also need to update the tutorial and docs since adding TC, so I will open up a follow-up issue for some of those things.

ad3154 (approved):

Issue #166, opened 2023-02-01
Functional requirements for hardware interface


GwnDaan:
I was trying to fix a lot of the issues from SonarCloud for the CANHardwareInterface class file but got a bit lost when trying to fix the larger functions mostly the threaded ones. So I'm switching to a bit more TDD approach for them I think. I propose we need to be able to do/test for in the hardware interface:

- [x] Start with #182 
- [x] Implementation of the isobus::send_can_message_frame_to_hardware
- [x] Implementation of the isobus::receive_can_message_frame_from_hardware
- [x] Function to add/remove/removeAll frame handler(s)
- [x] Function to add/remove/removeAll canFrame(s) to/from the rx queue
- [x] Function to add/remove/removeAll listener(s) for message received events
- [x] Function to add/remove/removeAll listener(s) for message sent events (as utility for applications)
- [x] Function to add/remove/removeAll listener(s) for update events
- [x] Function to set/get periodic update time 

Out of curiosity, why did we choose to make it a requirement that the frame handlers are added before the interface is started? Maybe for performance benefits?

I was also thinking that we could make a direct call to for example a `isobus::receive_can_message_from_hardware` function instead of the "raw can glue" we have now? With the listeners the application still has the ability to listen for messages in both directions, but they wont be able to break anything I guess. Same for making direct call to the `isobus::CANNetwork.update()` function. Though I see there could be weird behavior when they also call the update from within their own listener, hmm... I guess that's why it's the way it is now.

ad3154:
Time for another info dump from early development, haha!

Some of the reasons why the CAN hardware interface is like that are obsolete now. Initially I wanted the hardware interface to be completely optional, and removable so that the stack's main guts could be run on systems that didn't support std::thread and std::mutex but then those types ended up in the main stack anyways, so now that doesn't really matter that much... One reason it doesn't call `isobus::receive_can_message_from_hardware` directly was to prevent a circular dependency, which also doesn't really matter now, as we include the logger in the hardware layer.  Essentially the two layers have really become intertwined with more of the separation of duties happening at the plugin layer now instead of between the stack and the hardware interface instead of my original vision. This is fine I think as it's still very portable and hardware independent, so if we want to rework it a little to make it nicer with those thoughts out of the way that is A-OK with me.

PR #219, opened 2023-03-20
[Core]: Update hardware interface to use event dispatching system

GwnDaan:
## Changelog:
HardwareInterface:
- Major:
  - Call `CANNetworkManager::update()` directly by default
  - Call `CANNetworkManager::can_lib_process_rx_message()` directly by default
  - Use `EventDispatcher<>` for periodic update event
  - Use `EventDispatcher<isobus::HardwareInterfaceCANFrame>` for received/transmitted can frame event
  - Add `get_periodic_update_interval()` to get interval between periodic updates
  - Add more test cases for hardware interface
- Minor:
  - Renamed `can_thread` to `update_thread` 
  - Renamed `threadMutex` to `updateMutex`
  - Renamed `canLibNeedsUpdate` to `stackNeedsUpdate`
  - Renamed `periodicUpdateThread` to `wakeupThread`
  - Renamed `canLibUpdatePeriod` to `periodicUpdateInterval`
  - General optimizations to `update_thread_function` 

HardwareAbstraction:
- Renamed `send_can_message_to_hardware` to `send_can_message_frame_to_hardware`
- Added `receive_can_message_frame_from_hardware` as an abstraction layer
- Added `periodic_update_from_hardware` as an abstraction layer

### DIFF
--- a/examples/diagnostic_protocol/main.cpp
+++ b/examples/diagnostic_protocol/main.cpp
@@ -18,11 +18,6 @@ void signal_handler(int)
 	running = false;
 }
 
-void update_CAN_network(void *)
-{
-	isobus::CANNetworkManager::CANNetwork.update();
-}
-
 int main()
 {
 	std::signal(SIGINT, signal_handler);
--- a/examples/diagnostic_protocol/main.cpp
+++ b/examples/diagnostic_protocol/main.cpp
@@ -23,11 +23,6 @@ void update_CAN_network(void *)
 	isobus::CANNetworkManager::CANNetwork.update();
 }
 
-void raw_can_glue(isobus::HardwareInterfaceCANFrame &rawFrame, void *parentPointer)
-{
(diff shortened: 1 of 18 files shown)


GwnDaan:
Oh yea, also documentation needs to be updated if we wish to use these changes

ad3154:
I did some testing on this last night and it seemed to be working OK. In fact the CPU usage may have decreased somewhat, so that is nice. So, I'll go through and give it a review. Thanks for this!

I don't know if we'll really need to provide an option to stop calls to `update` and `can_lib_process_rx_message`, because chances are if someone wants to do that themselves, they probably aren't using that hardware layer anyways? I dunno, pretty low priority I think.

As a side note, I will be travelling for a few days, so I may be a bit slower/quieter than normal for a little bit.

GwnDaan:
Oh also this closes #166

ad3154 (approved):

PR #220, opened 2023-03-21
[Core]: Add support for Rusoku TouCAN

### LINKED ISSUE #133: Add TouCAN/CANAL Hardware plugin
The Rusoku TouCAN driver for Windows has recently been changed to MIT license. This adapter has a native driver for Linux (Socket CAN) so it would be neat to also include the Windows driver. 

It seems like it would be fairly straight forward to add support for this hardware.

https://github.com/rusoku/CANAL-DLL

Not a high priority but a nice-to-have

ad3154:
## What's new:

* Added a hardware plugin for the Rusoku TouCAN USB adapter via the VSCP CANAL api. 

Currently this is windows only, but they have a Mac driver out there as well, so I might also add that as a second effort later.

Closes #133


hardware_integration/src/toucan_vscp_canal.cpp
```diff
@@ -0,0 +1,115 @@
...
+	{
+		isobus::CANStackLogger::CAN_stack_log(isobus::CANStackLogger::LoggingLevel::Critical, "[TouCAN]: Error trying to connect to TouCAN probe. Check your device ID and serial number.");
+	}
+}
+
+bool TouCANPlugin::read_frame(isobus::HardwareInterfaceCANFrame &canFrame)
+{
+	long result = CANAL_ERROR_GENERIC;
+	structCanalMsg CANMsg = { 0 };
+	bool retVal = false;
```
GwnDaan, line 82: Seems like result has replaced this `retVal` variable, so it might as well be removed?

GwnDaan (approved):

PR #228, opened 2023-03-29
[Meta]: Update VT tutorial to match latest example

### LINKED ISSUE #217: Update VT Client Tutorial Docs
Since adding the event dispatcher in #211 we'll need to update the tutorial docs to describe the new callback registration methods.

GwnDaan:
### Changelog
Updates to VT tutorial:
- Use atomic running boolean flag instead of _exit()
- Add context pointer to CAN update callback
- Use available can drivers for driver initiation
- Add more context to smaller snippets (where which line should be located)
- Switch to the new dispatcher system for key event

Also contains some fixes for Sonar warnings in the VT example main file

Closes #217


GwnDaan:
Oh, some of these points should be applied to the other tutorials as well some time soon:
- Use atomic running boolean flag instead of _exit()
- Add context pointer to CAN update callback
- Use available can drivers for driver initiation

ad3154 (approved):

PR #230, opened 2023-03-31
[TC]: Add option to use a binary DDOP

### LINKED ISSUE #227: Using pre-generated binary DDOP format in TC client
Hi, I've noticed that the TC object pool (DDOP) is generated at every boot. For applications running on microcontrollers, this takes a few seconds due to the lack of processing power. Especially with larger pools I image the boot times being too long for great user-experience.

The ability to load in pre-generated binary object pools for the TC client. Possibly by overloading the current constructor?

ad3154:
## What's New:

* Adds the ability to use a binary DDOP in the form of either a vector of bytes, or a pointer and a size.
* Adds logic to search through a binary DDOP to locate the device object and extract the device structure label and localization label so that it can be checked against the TC's stored data.
* Optimizes parts of the state machine by pre-storing the structure and localization label on the TC worker thread instead of searching the whole DDOP for the label on the network manager's thread in the receive callback.

Closes #227


isobus/src/isobus_task_controller_client.cpp
```diff
@@ -623,14 +744,170 @@ namespace isobus
...
+
+						currentByteIndex += task_controller_object::DeviceObject::MAX_STRUCTURE_AND_LOCALIZATION_LABEL_LENGTH;
+						assert(currentByteIndex + task_controller_object::DeviceObject::MAX_STRUCTURE_AND_LOCALIZATION_LABEL_LENGTH < generatedBinaryDDOP.size()); // // Not enough bytes in your DDOP!
+						for (std::uint_fast8_t i = 0; i < task_controller_object::DeviceObject::MAX_STRUCTURE_AND_LOCALIZATION_LABEL_LENGTH; i++)
+						{
+							ddopLocalizationLabel[i] = (generatedBinaryDDOP.at(currentByteIndex + i)); // Read the localization label
+						}
+						break;
+					}
+				}
```
GwnDaan, line 894: We might be able to shorten this by using a lambda function? Afaik it will only run once on startup, so it won't have that much of an performance impact if any.

Something like
```cpp
	auto getDDOPSize = [&ddopUploadMode, &userSuppliedBinaryDDOPSize_bytes, &generatedBinaryDDOP]() {
		if (ddopUploadMode == DDOPUploadMode::UserProvidedBinaryPointer)
		{
			return userSuppliedBinaryDDOPSize_bytes;
		}
		else
		{
			return generatedBinaryDDOP.size();
		}
	};

	auto getDDOPByteAt = [&ddopUploadMode, &userSuppliedBinaryDDOPSize_bytes, &generatedBinaryDDOP](std::size_t index) {
		if (ddopUploadMode == DDOPUploadMode::UserProvidedBinaryPointer)
		{
			return userSuppliedBinaryDDOP[index];
		}
		else
		{
			return generatedBinaryDDOP.at(index);
		}
	};
```
might work?

Otherwise we also have to keep the code in the two cases synced, which I'm not really good at hahah

GwnDaan (changes requested):
Looking good! Only two small notes 😄

isobus/src/isobus_task_controller_client.cpp
```diff
@@ -114,16 +114,93 @@ namespace isobus
...
+	                                     bool reportToTCSupportsTCGEOWithPositionBasedControl,
+	                                     bool reportToTCSupportsPeerControlAssignment,
+	                                     bool reportToTCSupportsImplementSectionControl)
+	{
+		if (StateMachineState::Disconnected == get_state())
+		{
+			assert(!binaryDDOP.empty()); // Client will not work without a DDOP.
+			ddopStructureLabel.clear();
+			ddopLocalizationLabel.fill(0x00);
+			generatedBinaryDDOP = binaryDDOP;
```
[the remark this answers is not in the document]
ad3154: Yeah I debated on that as well... ultimately it just means instead of re-using the `generatedBinaryDDOP ` variable, I'd need to store another pointer, but that seems fine I suppose. I can change it

isobus/src/isobus_task_controller_client.cpp
ad3154: It does feel like a lot of duplication, doesn't it? I can try your solution!

GwnDaan (approved):

PR #232, opened 2023-04-01
[Core]: Add Hardware Integration to isobus namespace

### LINKED ISSUE #197: Inconsistent namespacing
**Describe the bug**
`CANHardwareInterface` and `SocketCANInterface` are not in the `isobus::` namespace.  I don't know what others aren't.  Is there a reason?  I know they're both part of the hardware abstraction layer, but maybe a more nested namespace like `isobus::hw::` would be more appropriate than not having them in at all.

ad3154:
This is probably a bit of a breaking change to our hardware plugin api, so it's good to get it out of the way sooner rather than later...

This change brings the hardware integration library under the isobus namespace for consistency. 
Previously it was in the global namespace, which is probably not ideal as described in #197.

Originally the idea was to have the hardware interface be completely separate and optional for the library, so I didn't want to put it in the same namespace, but these days it's pretty well entangled with the library and makes using it a lot easier, so might as well get it out of the global namespace.

I am open to comments if people think this is not the way to go.

Closes #197

### DIFF
--- a/examples/diagnostic_protocol/main.cpp
+++ b/examples/diagnostic_protocol/main.cpp
@@ -22,15 +22,15 @@ int main()
 {
 	std::signal(SIGINT, signal_handler);
 
-	std::shared_ptr<CANHardwarePlugin> canDriver = nullptr;
+	std::shared_ptr<isobus::CANHardwarePlugin> canDriver = nullptr;
 #if defined(ISOBUS_SOCKETCAN_AVAILABLE)
-	canDriver = std::make_shared<SocketCANInterface>("can0");
+	canDriver = std::make_shared<isobus::SocketCANInterface>("can0");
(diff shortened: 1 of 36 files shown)


GwnDaan (approved):
I agree with this change. I feel like we are long past the point of being able to integrate only the hardware integration part of the library. But that's not a bad thing if you ask me, it's fine with me with it being an nice "off-the-shelf" hardware integration for the "isobus" implementation part of this library, and not more than that. 
The other way around, where you only integrate the "isobus" implementation part of this library is still very much doable, so LGTM! 👍🏽

ad3154:
I've left this PR out here for a week, and so far there have been no negative comments, so I think I'll resolve the merge conflicts and get this merged in.

Apologies in advance for anyone who has their compilation break due to this change. It should be pretty quick to resolve though, just add `isobus::` to calls to the hardware layer and plugins, or `using namespace isobus` if you prefer that.

PR #236, opened 2023-04-12
[Core]: Added additional logging templates, more like printf

### LINKED ISSUE #222: Increase readability when using variables in log messages
Currently a lot of 'inbetween' formatting is necessary to get a running sentence. For example:
https://github.com/ad3154/Isobus-plus-plus/blob/44a2941ca23f50f0a7ed07b4f74b460c91db0815/isobus/src/isobus_task_controller_client.cpp#L839-L847
I find this quite a bit harder to read than using formatting specifiers, and therefore also harder to maintain.

A possibility would be to use the built-in `std::snprintf` to format messages. The same example from above would then look similar to:
```cpp
CANStackLogger::debug("[TC]: TC Server supports version %uh with %uh booms, %uh sections, and %uh position based control channels.",
                      messageData[1], messageData[5], messageData[6], messageData[7]);
```
Though one downside would be that confusion may arise when figuring out which variables belongs to which specifier. But I feel like a more readable sentence outweighs that.

**Additional information**
I found an example on [stackoverflow](https://stackoverflow.com/a/26221725) while gathering information that could be useful:
```c++
#include <memory>
#include <string>
#include <stdexcept>

template<typename ... Args>
std::string string_format( const std::string& format, Args ... args )
{
    int size_s = std::snprintf( nullptr, 0, format.c_str(), args ... ) + 1; // Extra space for '\0'
    if( size_s <= 0 ){ throw std::runtime_error( "Error during formatting." ); }
    auto size = static_cast<size_t>( size_s );
    std::unique_ptr<char[]> buf( new char[ size ] );
    std::snprintf( buf.get(), size, format.c_str(), args ... );
    return std::string( buf.get(), buf.get() + size - 1 ); // We don't want the '\0' inside
}
```

The code snippet above is licensed under [CC0 1.0](https://creativecommons.org/publicdomain/zero/1.0/).

ad3154:
* Added variadic logging functions to the CAN Stack Logger.

This isn't the most efficient, but it does make our code much nicer. Based on code from @GwnDaan, slightly modified.

Closes #222

### DIFF
--- a/isobus/include/isobus/isobus/can_stack_logger.hpp
+++ b/isobus/include/isobus/isobus/can_stack_logger.hpp
@@ -63,6 +63,10 @@ namespace isobus
 				std::snprintf(buf.get(), size, format.c_str(), args...);
 				CAN_stack_log(level, std::string(buf.get(), buf.get() + size - 1)); // We don't want the '\0' inside
 			}
+			else if (size_s < 0)
+			{
+				CAN_stack_log(level, format); // If snprintf had some error, at least print the format string
+			}
 		}
 
 		/// @brief Logs a string to the log sink with `Debug` severity. Wraps sink_CAN_stack_log.
--- a/isobus/include/isobus/isobus/can_stack_logger.hpp
+++ b/isobus/include/isobus/isobus/can_stack_logger.hpp
@@ -10,6 +10,7 @@
 #ifndef CAN_STACK_LOGGER_HPP
 #define CAN_STACK_LOGGER_HPP
 
+#include <memory>
 #include <mutex>
 #include <string>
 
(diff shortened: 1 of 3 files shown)


isobus/include/isobus/isobus/can_stack_logger.hpp
```diff
@@ -37,36 +38,98 @@ namespace isobus
...
+		static void CAN_stack_log(LoggingLevel level, const std::string &format, Args... args)
+		{
+			int size_s = std::snprintf(nullptr, 0, format.c_str(), args...) + 1; // Extra space for '\0'
+			if (size_s > 0)
+			{
+				auto size = static_cast<std::size_t>(size_s);
+				std::unique_ptr<char[]> buf(new char[size]);
+				std::snprintf(buf.get(), size, format.c_str(), args...);
+				CAN_stack_log(level, std::string(buf.get(), buf.get() + size - 1)); // We don't want the '\0' inside
+			}
```
GwnDaan, line 65: What if `snprintf` has a formatting error, e.g. it returns a negative number? We might want to print the string without formatting it so the log message is still preserved? Could be really frustrating to not see a log message show up I feel like

GwnDaan (changes requested):

isobus/include/isobus/isobus/can_stack_logger.hpp
ad3154: Yep, good point, I've added another condition to at least print the format string if snprintf returns a negative number.

GwnDaan (approved):

PR #237, opened 2023-04-14
[VT]: Added some additional getters to the VT client

ad3154:
- Added getters for the VT's control functions.
- Added a way to get the active working set master's address as reported by the VT server. 
- Added unit tests for these new functions, which required making the state machine states public similar to the TC client implementation.

### DIFF
--- a/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
+++ b/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
@@ -242,6 +242,34 @@ namespace isobus
 			ReservedRemoveAssignment = 31 ///< Used for Remove assignment command
 		};
 
+		/// @brief The internal state machine state of the VT client, mostly just public so tests can access it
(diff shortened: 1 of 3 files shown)


isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
```diff
@@ -292,6 +320,18 @@ namespace isobus
 		/// @brief Terminates the client and joins the worker thread if applicable
 		void terminate();
 
+		/// @brief Returns the partner control function being used by the client
+		/// @returns The partner control function being used by the client
+		std::shared_ptr<PartneredControlFunction> get_partner_control_function() const;
```
GwnDaan, line 325:
```suggestion
		/// @brief Returns the control function of the VT server with which this VT client communicates.
		/// @returns The partner control function for the VT server
		std::shared_ptr<PartneredControlFunction> get_partner_control_function() const;
```

GwnDaan (approved):

PR #240, opened 2023-04-18
[ISB]: Add support for ISOBUS shortcut button (ISB)

ad3154:
* Added support for the AEF ISOBUS shortcut button, including sending as an ISB, and receiving states from other ISBs.

An ISB is essentially a command to all implements to enter a safe state. See the description located at https://www.isobus.net/isobus/pGNAndSPN/10936, ISO 11783-7, or https://www.aef-online.org/fileadmin/user_upload/Content/pdfs/AEF_One_Pager.pdf for more details.

### DIFF
--- a/isobus/include/isobus/isobus/isobus_shortcut_button_interface.hpp
+++ b/isobus/include/isobus/isobus/isobus_shortcut_button_interface.hpp
@@ -79,10 +79,10 @@ namespace isobus
 		/// @brief Destructor for a ShortcutButtonInterface
 		virtual ~ShortcutButtonInterface();
 
-		/// @brief Adds a callback that will get called when any CF commands a stop to implement operations
-		/// and likewise will be called again when all tracked CFs are sending the "permit implement operations" state
-		/// @param[in] callback The callback to add
-		std::shared_ptr<void> add_stop_all_implement_operations_state_callback(std::function<void(StopAllImplementOperationsState ISBState)> callback);
+		/// @brief Gets the event dispatcher for when the assigned bus' ISB state changes.
+		/// The assigned bus is determined by which internal control function you pass into the constructor.
+		/// @returns The event dispatcher which can be used to register callbacks/listeners to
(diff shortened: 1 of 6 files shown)


isobus/include/isobus/isobus/isobus_shortcut_button_interface.hpp
```diff
@@ -0,0 +1,169 @@
...
+		/// @param[in] serverEnabled Enables the interface's transmission of the "Stop all implement operations" message.
+		ShortcutButtonInterface(std::shared_ptr<InternalControlFunction> internalControlFunction, bool serverEnabled = false);
+
+		/// @brief Destructor for a ShortcutButtonInterface
+		virtual ~ShortcutButtonInterface();
+
+		/// @brief Adds a callback that will get called when any CF commands a stop to implement operations
+		/// and likewise will be called again when all tracked CFs are sending the "permit implement operations" state
+		/// @param[in] callback The callback to add
+		std::shared_ptr<void> add_stop_all_implement_operations_state_callback(std::function<void(StopAllImplementOperationsState ISBState)> callback);
```
GwnDaan, line 85: It could be better to return the EventDispatcher itself instead here, then we leave the choice of which listener to use (listener with no context, or with safe/unsafe context) up to the user. 

Like here in the HardwareInterface:
https://github.com/ad3154/Isobus-plus-plus/blob/9a790740f3cd26cbd12ffa3493f6fd4be88b80f4/hardware_integration/src/can_hardware_interface.cpp#L223-L236

The VT client still uses the same approach as used here by you, but could be converted to the one in the HardwareInterface as well to make the it more consistent through the project :)

GwnDaan (changes requested):

isobus/include/isobus/isobus/isobus_shortcut_button_interface.hpp
ad3154: Oh, I see what you're saying, yeah I followed along with how the VT did it but this seems more flexible, so I can change it.

GwnDaan (approved):

PR #242, opened 2023-04-22
[DP]: Add ISOBUS Control Function Functionalities Message

ad3154:
## What's new

* This adds support for sending the CF Functionalities message as part of the diagnostic protocol class.
* The new class will automatically respond to requests from ANY ecu, either as a broadcast or point-to-point for the "Control Function Functionalities" PGN (0xFC8E, 64654)
* ISOBUS ECUs are required to support this message, so by default, we will report every consumer of our library as a "Minimum Control Function" of "generation 1" unless configured otherwise (if you are using the diagnostic protocol).
* In general you will want your application to configure this appropriately for what you are using. If you are using the VT client, you should configure this message to send "Universal Terminal Working Set" as supported. If you are using the TC client, same thing, you will want to report you are a "Task Controller Section Control Client" with the appropriate data such as number of booms and sections supported. There are several functionalities to be aware of!
* The functionalities are documented here https://www.isobus.net/isobus/option
* The PGN is loosely defined here https://www.isobus.net/isobus/pGNAndSPN/10250 but mostly just says refer to ISO11783-12

### DIFF
--- a/isobus/src/isobus_functionalities.cpp
+++ b/isobus/src/isobus_functionalities.cpp
@@ -1,0 +1,463 @@
+//================================================================================================
+/// @file isobus_functionalities.cpp
+///
+/// @brief Implements the management of the ISOBUS control function functionalities message.
+/// @author Adrian Del Grosso
+///
+/// @copyright 2023 Adrian Del Grosso
+//================================================================================================
+#include "isobus/isobus/isobus_functionalities.hpp"
+#include "isobus/isobus/can_stack_logger.hpp"
+
+#include <limits>
+
+namespace isobus
+{
+	void ControlFunctionFunctionalities::set_functionality_is_supported(Functionalities functionality, std::uint8_t functionalityGeneration, bool isSupported)
+	{
+		const std::lock_guard<std::mutex> lock(functionalitiesMutex);
+
+		auto existingFunctionality = get_functionality(functionality);
+
+		if (supportedFunctionalities.end() == existingFunctionality)
+		{
+			FunctionalityData newFunctionality(functionality);
+			newFunctionality.configure_default_data();
+			newFunctionality.generation = functionalityGeneration;
+			supportedFunctionalities.emplace_back(newFunctionality);
+		}
+	}
+
+	bool ControlFunctionFunctionalities::get_functionality_is_supported(Functionalities functionality)
+	{
+		bool retVal = false;
+		const std::lock_guard<std::mutex> lock(functionalitiesMutex);
+
+		auto existingFunctionality = get_functionality(functionality);
+
+		if (supportedFunctionalities.end() != existingFunctionality)
+		{
+			retVal = true;
+		}
+		return retVal;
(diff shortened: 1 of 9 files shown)


isobus/src/isobus_functionalities.cpp
```diff
@@ -0,0 +1,872 @@
...
+			break;
+		}
+		return retVal;
+	}
+
+	bool ControlFunctionFunctionalities::pgn_request_handler(std::uint32_t parameterGroupNumber,
+	                                                         ControlFunction *,
+	                                                         bool &acknowledge,
+	                                                         AcknowledgementType &,
+	                                                         void *parentPointer)
```
GwnDaan, line 835: The `parentPointer` is of type `void *` here, wouldn't it make more sense if we just specify the type we want the pointer to point to. E.g. `ControlFunctionFunctionalities *` here?

I'm not sure why we decided to use these `void *` pointers when using them internally in private/protected functions for parents? But I feel like I prefer overloading the function if we need to support pointers to multiple types of parents. I know we use it in more places throughout the project but I just now begin to question if it's optimal choice for readability/maintainability. What are your thoughts on this?

GwnDaan: Oh, and we will make Sonar happier as well by specifying a meaningful type :)

ad3154: In this case it's because we're matching the signature of the callback inside `ParameterGroupNumberRequestProtocol` which doesn't know anything about `ControlFunctionFunctionalities` and has to support any object type, as any external consumer can use the same `register_pgn_request_callback`. I do agree it's not the greatest for readability and knowing whats happening, but it does provide a type agnostic way to get a pointer to an instance of the registrant, which is the main goal I was trying to achieve. Open to other ideas on this of course

isobus/src/isobus_functionalities.cpp
```diff
@@ -0,0 +1,872 @@
...
+		if (static_cast<std::uint32_t>(CANLibParameterGroupNumber::ControlFunctionFunctionalities) == parameterGroupNumber)
+		{
+			acknowledge = false;
+			targetInterface->txFlags.set_flag(static_cast<std::uint32_t>(TransmitFlags::ControlFunctionFunctionalitiesMessage));
+			retVal = true;
+		}
+		return retVal;
+	}
+
+	void ControlFunctionFunctionalities::process_flags(std::uint32_t flag, void *parentPointer)
```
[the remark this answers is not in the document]
ad3154: Yeah similar thing here, it mostly just avoids templating a highly generic class

isobus/src/isobus_functionalities.cpp
GwnDaan: Ah I see, disregard my comment for now. We could maybe switch to the EventDispatcher system for the `register_pgn_request_callback` at some stage, but I'm not sure what side effects that might have or even performance loss.

PR #243, opened 2023-04-25
[Core]: Add support for J1939/ISOBUS commanded address message

ad3154:
* Added processing and use of the commanded address PGN. 
* Changed some CAN message getters to be const.


GwnDaan (approved):
LGTM! Haven't tested it though, not sure how to replicate the commanded address behavior in a real life environment

ad3154:
> LGTM! Haven't tested it though, not sure how to replicate the commanded address behavior in a real life environment

I was at least able to test it a bit by manually sending the series of messages, and it seemed to work OK.
image

PR #247, opened 2023-04-27
[HW]: Make MCP2515 plugin transmitting and receiving reliable

GwnDaan:
Previously the MCP2515 plugin wasn't always transmitting or receiving CAN messages in the correct order, this should be fixed in the PR. The changes made to achieve this:
- Use an cached index for reading the correct buffer when messages 'rollover'
- Log to the user when a frame failed to transmit, this is done when sending the next message to preserve communication speed
- Only use **one** of the three available buffers for transmitting. I'm not sure how to implement multiple transmitting buffers while keeping the messages transmitted in the correct order. Any ideas are welcome!

Also changed the timeout for transmitting/receiving messages with the TWAI plugin to 100ms. This might be useful for example when no CANbus is connected.

### DIFF
--- a/hardware_integration/include/isobus/hardware_integration/mcp2515_can_interface.hpp
+++ b/hardware_integration/include/isobus/hardware_integration/mcp2515_can_interface.hpp
@@ -173,6 +173,8 @@ namespace isobus
 
 		SPIHardwarePlugin *transactionHandler; ///< The SPI transaction handler
 		std::uint8_t rxIndex = 0; ///< The index of the rx buffer to read from next
+		std::uint8_t txIndex = 2; ///< The index of the tx buffer to write to next, start with 2 as it is the buffer with the highest priority
+		std::uint8_t txPriority = 3; ///< The priority of the next tx frame
 		const std::uint8_t cfg1; ///< Configuration value for CFG1 register
 		const std::uint8_t cfg2; ///< Configuration value for CFG2 register
 		const std::uint8_t cfg3; ///< Configuration value for CFG3 register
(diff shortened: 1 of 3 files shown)


ad3154:
image
Seems like if the priorities are set to the same value for each buffer, and if you can get interrupts back when a buffer is sent, plus know which buffer was sent inside the interrupt, then you could probably maintain some idea of which buffers are loaded and search from highest to lowest buffer to decide which one to use?

But that does sound pretty complex if you're having to manage the interrupt, and is not particularly efficient if then you have to read a register to know which buffer just sent... so maybe using one is best?

GwnDaan:
Before it took around 32 seconds to upload custom object pool of ~300kb, after commit https://github.com/ad3154/Isobus-plus-plus/pull/247/commits/b203e2dd972fef7ee8d5ff1317ca002106a53505 it is reduced to approximately 26 seconds. This is measured with the MCP2515 running together with an ESP32-S3.

The TWAI plugin on the other hand does exactly the same in around 18 seconds. I'm not sure if this a hardware limitation related to the MCP2515, the 16Mhz crystal used for it, or the driver in this library. It seems all the TX buffers of the MCP2515 are constantly full when uploading the object pool, so I don't think it's this driver, but you never know hahah. 

Still, all these timings seem reasonable to me as large uploads won't be that frequent I guess.

GwnDaan:
> It seems all the TX buffers of the MCP2515 are constantly full when uploading the object pool, so I don't think it's this driver, but you never know hahah.

Ahh, it might be the driver actually. We're not using any interrupt based system to read messages with the MCP2515 driver, it uses polling instead, but with a delay in between of 6ms (to prevent platform with lower tick rates from never switching threads). So when we're in a ETP session and we need to receive for example a CTS message, then it might take longer as when we would be using an interrupt to read messages.

We should probably support interrupt driven reading of messages at some point, but I don't think it has high priority right now. :)

ad3154 (approved):

PR #249, opened 2023-04-29
[Core]: Improved lost CF pruning

### LINKED ISSUE #192: Interfering internal CF addresses
**Describe the bug**
When creating two internal CF addresses with the same preferred address, they will both continue using the same address. This causes information for one devices to be handled by both and leading to undefined behavior. This bug occurred by running two executables, though it might also happen within the same executable?

**How to replicate**
Run the VT "auxiliary function" and "version3 object pool" examples simultaneously, with both of them preferring the same internal address.

ad3154:
* Added active control function pruning so that if a control function fails to claim after a request for the address claim PGN, we move that control function to the inactive list and invalidate its address. 

This was always meant to be a part of how we manage address claiming, but for whatever reason got left behind. This greatly increases the reliability of checking a control function's address being valid.

* Fixed a random crash in unit tests.
* Added additional logging to the address claim state machine to more clearly describe what is happening.

Closes #192

### DIFF
--- a/isobus/src/can_address_claim_state_machine.cpp
+++ b/isobus/src/can_address_claim_state_machine.cpp
@@ -288,6 +288,11 @@ namespace isobus
 							{
 								// Wait for things to shake out a bit, then claim a new address.
 								parent->set_current_state(State::WaitForRequestContentionPeriod);
+								parent->m_claimedAddress = NULL_CAN_ADDRESS;
+								CANStackLogger::warn("[AC]: Internal control function %016llx on channel %u must re-arbitrate its address because it was stolen by another ECU with NAME %016llx.",
+								                     parent->m_isoname.get_full_name(),
+								                     parent->m_portIndex,
+								                     NAMEClaimed);
 							}
 						}
 					}
--- a/isobus/src/can_address_claim_state_machine.cpp
+++ b/isobus/src/can_address_claim_state_machine.cpp
@@ -170,6 +170,10 @@ namespace isobus
 				{
 					if (send_address_claim(m_preferredAddress))
 					{
+						CANStackLogger::debug("[AC]: Internal control function %016llx has claimed address %u on channel %u",
+						                      m_isoname.get_full_name(),
+						                      m_preferredAddress,
+						                      m_portIndex);
 						set_current_state(State::AddressClaimingComplete);
 					}
 					else
@@ -189,6 +193,10 @@ namespace isobus
 						if ((nullptr == CANNetworkManager::CANNetwork.get_control_function(m_portIndex, i, {})) && (send_address_claim(i)))
 						{
 							addressFound = true;
+							CANStackLogger::debug("[AC]: Internal control function %016llx could not use the preferred address, but has claimed address %u on channel %u",
+							                      m_isoname.get_full_name(),
+							                      m_preferredAddress,
+							                      m_portIndex);
 							set_current_state(State::AddressClaimingComplete);
 							break;
 						}
--- a/isobus/src/can_address_claim_state_machine.cpp
+++ b/isobus/src/can_address_claim_state_machine.cpp
@@ -195,7 +195,7 @@ namespace isobus
 							addressFound = true;
 							CANStackLogger::debug("[AC]: Internal control function %016llx could not use the preferred address, but has claimed address %u on channel %u",
 							                      m_isoname.get_full_name(),
-							                      m_preferredAddress,
+							                      i,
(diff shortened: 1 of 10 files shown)


ad3154:
@GwnDaan Since this is linked to your issue originally, when you have time, I'd love a review

GwnDaan:
> @GwnDaan Since this is linked to your issue originally, when you have time, I'd love a review

Sounds good, it will still be a few days before I can do any testing or reviewing. But I'll be happy to test it when I get the time :)

GwnDaan:
Hi @ad3154, still seeing some weird behavior when trying to steal a CF:

To replicate, first start the version3_object_pool example (claims 0x1C):
image

Then start the aux_inputs example (modified to claim 0x1C as well):
image

But now the version3_object_pool example loses connection with the VT:
image

What are your thoughts?

GwnDaan:
The other way around fails now as well (at the time of creating #192 it worked fine IIRC):

Here I first started the` aux_input` example and then the `version3_object_pool` example:
image

image

isobus/src/can_address_claim_state_machine.cpp
```diff
@@ -170,6 +170,10 @@ namespace isobus
 				{
 					if (send_address_claim(m_preferredAddress))
 					{
+						CANStackLogger::debug("[AC]: Internal control function %016llx has claimed address %u on channel %u",
+						                      m_isoname.get_full_name(),
+						                      m_portIndex,
+						                      m_preferredAddress);
```
GwnDaan, line 176:
```suggestion
						CANStackLogger::debug("[AC]: Internal control function %016llx has claimed address %u on channel %u",
						                      m_isoname.get_full_name(),
						                      m_preferredAddress,
						                      m_portIndex);
```

isobus/src/can_address_claim_state_machine.cpp
```diff
@@ -189,13 +193,20 @@ namespace isobus
 						if ((nullptr == CANNetworkManager::CANNetwork.get_control_function(m_portIndex, i, {})) && (send_address_claim(i)))
 						{
 							addressFound = true;
+							CANStackLogger::debug("[AC]: Internal control function %016llx could not use the preferred address, but has claimed address %u on channel %u",
+							                      m_isoname.get_full_name(),
+							                      m_portIndex,
+							                      m_preferredAddress);
```
GwnDaan, line 199:
```suggestion
							CANStackLogger::debug("[AC]: Internal control function %016llx could not use the preferred address, but has claimed address %u on channel %u",
							                      m_isoname.get_full_name(),
							                      m_preferredAddress,
							                      m_portIndex);
```

GwnDaan (changes requested):

ad3154:
Thanks for the feedback! I will continue working on this...

ad3154:
I think I've got all the issues worked out in here now.

In the following log output, you can see the VT example runs normally, and properly detects when I unplug a device from the bus.

image

ad3154:
Okay, now the unit test should be fixed as well. 🙂

GwnDaan:
> The other way around fails now as well (at the time of creating #192 it worked fine IIRC):
> 
> Here I first started the` aux_input` example and then the `version3_object_pool` example: image
> 
> image

I still partially have this issue. It seems that when the aux_inputs example get's kicked of it's preferred address, it just claims it's preferred address back again according to the logs:

`aux_inputs` example (started before `version3_object_pool`  example):
image

`version3_object_pool` example logs (started right after the still running `aux_inputs` example):
image

ad3154:
Yeah, that's not the right behavior - I'll take a look at that, thanks

ad3154:
Looks like a logging issue... I was passing in the preferred address to that log statement instead of the loop variable.
image

GwnDaan:
Hmm then the issue with the object pool might be out of the scope of this PR? Not sure if it's related to CF's

> `version3_object_pool` example logs (started right after the still running `aux_inputs` example):
> image

ad3154:
After recreating it and looking at a candump, it seems like the address claiming messages and process are correct, but looking at the address table in the network manager, there's definitely something sketchy going on, as the slot that should be the Vt Example's ICF is occupied by the external control function that moved to a different address. So although the correct messages were sent, our internal tracking of them seems to be confused. I've come this far with it, I might as well try and fix it before merging this in... 
image

ad3154:
Okay, now I think I've got it. Ready for re-review @GwnDaan 

image

GwnDaan (approved):

PR #254, opened 2023-05-14
[Guidance]: Add basic read-only example

GwnDaan:
**Main changes:**
- Added read-only example
- Added a `changed` param to the event dispatchers
- Added the ability to choose which message you want to send periodically (Most applications don't want to sent both afaik)

**Other changes:**:
- Renamed a few of the variables/functions/classes. I tried to make the naming more consistent throughout the guidance interface. Now the main interface is called "AgriculturalGuidanceInterface" and all the other naming have the "Agriculture" part removed. Before, the word was not really consistently used as it was present in some places, while not in others.
- Added small timeout to the VirtualCANInterface, such that when a test fails to read a frame, it doesn't hang forever.

Sorry for changing so much, I happened to come across this while making the example.

### DIFF
--- a/isobus/include/isobus/isobus/isobus_guidance_interface.hpp
+++ b/isobus/include/isobus/isobus/isobus_guidance_interface.hpp
@@ -38,7 +38,12 @@ namespace isobus
 		/// @brief Constructor for a AgriculturalGuidanceInterface
 		/// @param[in] source The internal control function to use when sending messages, or nullptr for listen only
 		/// @param[in] destination The destination control function for transmitted messages, or nullptr for broadcasts
-		AgriculturalGuidanceInterface(std::shared_ptr<InternalControlFunction> source, std::shared_ptr<ControlFunction> destination);
+		/// @param[in] sentSystemCommandPeriodically If true, the system command message will be sent periodically. This should (only) be used by an guidance application trying to steer a machine.
+		/// @param[in] sentMachineInfoPeriodically If true, the machine info message will be sent periodically. This should (only) be used by the steering controller itself.
+		AgriculturalGuidanceInterface(std::shared_ptr<InternalControlFunction> source,
+		                              std::shared_ptr<ControlFunction> destination,
+		                              bool sentSystemCommandPeriodically = false,
+		                              bool sentMachineInfoPeriodically = false);
 
 		/// @brief Destructor for the AgriculturalGuidanceInterface
 		~AgriculturalGuidanceInterface();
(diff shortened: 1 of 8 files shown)


isobus/include/isobus/isobus/isobus_guidance_interface.hpp
```diff
@@ -191,36 +196,41 @@ namespace isobus
 				NotAvailable = 63 ///< Parameter not supported
 			};
 
-			/// @brief Constructor for a AgriculturalGuidanceMachineInfo
-			AgriculturalGuidanceMachineInfo() = default;
+			/// @brief Constructor for a GuidanceMachineInfo
+			/// @param[in] sender The control function that is sending this message
+			GuidanceMachineInfo(ControlFunction *sender);
```
ad3154, line 201: Since these constructors are no longer `default` we should mark them as `explicit`

isobus/src/isobus_guidance_interface.cpp
```diff
@@ -32,14 +32,18 @@
 
 namespace isobus
 {
-	GuidanceInterface::GuidanceInterface(std::shared_ptr<InternalControlFunction> source, std::shared_ptr<ControlFunction> destination) :
+	AgriculturalGuidanceInterface::AgriculturalGuidanceInterface(std::shared_ptr<InternalControlFunction> source,
+	                                                             std::shared_ptr<ControlFunction> destination,
+	                                                             bool sentSystemCommandPeriodically,
```
ad3154, line 37: Maybe `send` instead of `sent`? Or `enableSending`? Might be clearer

isobus/src/isobus_guidance_interface.cpp
```diff
@@ -48,196 +52,240 @@ namespace isobus
...
+	AgriculturalGuidanceInterface::GuidanceMachineInfo::GuidanceMachineInfo(ControlFunction *sender) :
+	  controlFunction(sender)
 	{
-		estimatedCurvature = curvature;
 	}
 
-	float GuidanceInterface::AgriculturalGuidanceMachineInfo::get_estimated_curvature() const
+	bool AgriculturalGuidanceInterface::GuidanceMachineInfo::set_estimated_curvature(float curvature)
+	{
+		if (fabs(estimatedCurvature - curvature) > std::numeric_limits<float>::epsilon())
```
ad3154, line 112:
```suggestion
		if (std::fabs(estimatedCurvature - curvature) > std::numeric_limits<float>::epsilon())
```

isobus/src/isobus_guidance_interface.cpp
```diff
@@ -48,196 +52,240 @@ namespace isobus
...
 	{
-		commandedCurvature = curvature;
+		return commandedStatus;
 	}
 
-	float GuidanceInterface::GuidanceSystemCommand::get_curvature() const
+	bool AgriculturalGuidanceInterface::GuidanceSystemCommand::set_curvature(float curvature)
 	{
-		return commandedCurvature;
+		if (fabs(commandedCurvature - curvature) > std::numeric_limits<float>::epsilon())
```
ad3154, line 77:
```suggestion
		if (std::fabs(commandedCurvature - curvature) > std::numeric_limits<float>::epsilon())
```

ad3154 (changes requested):

GwnDaan:
Applied your suggestions 👍🏽

ad3154 (approved):

PR #255, opened 2023-05-16
[Core]: Improvement to CAN message classes

GwnDaan:
### What changed?
- Renamed `HardwareInterfaceCANFrame` to `CANMessageFrame` for consistensy and editor auto-completions
- Removed `CANLibManagedMessage` in favor of having setters in `CANMessage`
- Removed unused uniqueID in `CANMessage` and resolved small code smells

Do we want these changes? I feel like the result is pretty clean. Though I mostly used 'replace-all' of vscode for renaming and removing parts, so there might still be some artifacts left somewhere.


ad3154:
Well, I guess the reason the other class exists is because I didn't want any application layer consumer of our messages to be able to modify any values in them, since that message may be processed in other callbacks as well. So one solution to that was having zero setters. We could maybe accomplish that by going around and making sure all references to messages the user might touch in callbacks are `const`?

Originally I was going to use `uniqueID` as a way for the network manager to know if something was sent if people didn't use the hardware interface, but never implemented that... and since it's so tightly tied together now I think it makes sense to remove it.

I'm ambivalent about renaming `HardwareInterfaceCANFrame` I guess. It would be more consistent maybe to just have `CANFrame` I think though, otherwise we might want to also rename CANMessage to CANPacket (to imply it can be many frames), but then we are changing sooooooo much.... which is not super exciting.

GwnDaan:
> Well, I guess the reason the other class exists is because I didn't want any application layer consumer of our messages to be able to modify any values in them, since that message may be processed in other callbacks as well. So one solution to that was having zero setters. We could maybe accomplish that by going around and making sure all references to messages the user might touch in callbacks are `const`?

I feel like changing all occurrences to `const` makes more sense here and simplifies it a bit. With the current approach where we have 'protected' variables, it makes it a bit harder to find where they get changed. As theoretically any derived class could be setting them. I don't know, I'm fine with both approaches, but prefer using const references with setters as looks more maintainable and centralized to me.

> I'm ambivalent about renaming `HardwareInterfaceCANFrame` I guess. It would be more consistent maybe to just have `CANFrame` I think though, otherwise we might want to also rename CANMessage to CANPacket (to imply it can be many frames), but then we are changing sooooooo much.... which is not super exciting.

Hmm, I see. My initial thought was a CANMessage basically exists of one or more frames, and so a frame is part of a message, hence the name CANMessageFrame. But CANFrame sounds also fine to me. 
Regarding renaming CANMessage to CANPacket. In the standard (Part 3) they say: 

>5.10.2.1 General
Messages greater than 8 byte in length are too large to fit into a single classical CAN data frame.
Therefore, they are broken into several smaller packets, and those packets are transmitted in separate
classical CAN data frames.

> 5.10.2.4 “Packetization”
A multi-packet message is defined as one whose data does not fit into the data field of a single CAN data
frame (i.e. messages with a data field greater than 8 byte).

I interpret this a 'packet' representing 7 bytes of data when transporting a 'message' larger than 8 bytes. So a 'packet' is basically always a 'frame', but a 'frame' is not necessarily a 'packet.  Hence renaming CANMessage to CANPacket seems not the correct approach. Though I must say that the standard is not really that consistent in their naming.

ad3154:
OK, that all makes sense I think. Still a shame this will be such a disruptive change, but it does make it cleaner and consolidate how messages are used. I suppose it also opens up some ability for us to create another version of `send_can_message` that allows sending of a `CANMessage` (though that would need a lot of validation checking before sending).

GwnDaan:
All occurrences should be changed to use const references. I kept the CANMessageFrame in place, @ad3154 let me know if you want me to change it to CANFrame, now is the time 😄. Otherwise the PR is ready for review

ad3154 (approved):

PR #259, opened 2023-05-19
[VT]: VT server managed working set bugfix

dlips:
Hi, this is my first pull-request and just a small bug fix that prevents a segmentation fault if the vtObjectTree is a nullptr.
Additionally a warning due to an out-of-order initializier list is fixed.

### DIFF
--- a/isobus/src/isobus_virtual_terminal_server_managed_working_set.cpp
+++ b/isobus/src/isobus_virtual_terminal_server_managed_working_set.cpp
@@ -34,7 +34,7 @@ namespace isobus
 	  processingState(ObjectPoolProcessingThreadState::None)
 	{
 		initialize_colour_table();
-		
+
 		if (nullptr != associatedControlFunction)
 		{
 			CANStackLogger::info("[WS]: New VT Server Object Created for CF " + isobus::to_string(static_cast<int>(associatedControlFunction->get_NAME().get_full_name())));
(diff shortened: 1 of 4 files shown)


isobus/src/isobus_virtual_terminal_server_managed_working_set.cpp
```diff
@@ -7,19 +7,20 @@
...
 namespace isobus
 {
 
 	VirtualTerminalServerManagedWorkingSet::VirtualTerminalServerManagedWorkingSet() :
+	  workingSetID(NULL_OBJECT_ID),
 	  objectPoolProcessingThread(nullptr),
 	  workingSetControlFunction(nullptr),
 	  processingState(ObjectPoolProcessingThreadState::None),
 	  workingSetMaintenanceMessageTimestamp_ms(0),
-	  workingSetID(NULL_OBJECT_ID),
```
ad3154, line 22: At some point we will probably want to change all the initializer lists to in-class initializers where possible. I was doing it the sort of old-school way when I wrote this. But it's fine for now.

isobus/src/isobus_virtual_terminal_server_managed_working_set.cpp
```diff
@@ -45,12 +47,15 @@ namespace isobus
...
+		if (nullptr != vtObjectTree)
 		{
-			if (nullptr != (*key).second)
+			for (auto key = vtObjectTree->begin(); key != vtObjectTree->end(); ++key)
 			{
-				delete (*key).second;
-				(*key).second = nullptr;
+				if (nullptr != (*key).second)
+				{
+					delete (*key).second;
```
ad3154, line 56: At some point we will probably want to change the raw pointer being used for the working set to be some kind of smart pointer, but again it's a low priority and doesn't matter too much right now since you're just merging back to a development branch.

ad3154 (approved):

PR #261, opened 2023-05-22
[TP]: Add missing message length for frame chunk callback

GwnDaan:
This PR is bugfix for messages being transported by data chunk callbacks using any transport protocol. Their message length was recognized as 0 due to the removal of CANLibManagedMessage in #255.

### Changes:
- Add `frameChunkCallbackMessageLength` to all transporting protocol session classes
- Add utility function `get_message_data_length` to all transporting protocol session classes to determine the correct message length considering both data chunk callbacks and the normal copied message data.
- Also changed the message length/indexes in CANMessage from `size_t` to `uint32_t` as it can't exceed the maximum message length of ETP: 117440505

### DIFF
--- a/isobus/include/isobus/isobus/can_extended_transport_protocol.hpp
+++ b/isobus/include/isobus/isobus/can_extended_transport_protocol.hpp
@@ -81,6 +81,10 @@ namespace isobus
 			/// @brief A useful way to compare sesson objects to each other for equality
 			bool operator==(const ExtendedTransportProtocolSession &obj);
 
+			/// @brief Get the total number of bytes that will be sent or received in this session
+			/// @return The length of the message in number of bytes
+			std::uint32_t get_message_data_length() const;
+
 		private:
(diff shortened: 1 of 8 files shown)


ad3154:
Seems to throw an assertion for an out of range vector access here:
image

ad3154:
Seems to be because this for loop expects to add all the 0xFF bytes to the received message, but the message size is 9, not 14. 9 Is the correct length, so might need to add an additional check to this loop

ad3154:
[gwndaan-10f89a0-TP_ETPHotfix.patch](https://github.com/ad3154/Isobus-plus-plus/files/11546372/gwndaan-10f89a0-TP_ETPHotfix.patch)
@GwnDaan Here's a patch that should fix the noted issue

ad3154:
If you are busy, since this is a fairly high impact issue, I can merge this in and submit a fix on top of it

PR #267, opened 2023-05-28
[Meta]: Update project name, URL, and sonar organization name

ad3154:
* Updated most references to Isobus++ to be AgIsoStack++

Part of #175


GwnDaan (approved):
LGTM, although I don't really have time to check if all occurrences have been renamed, we can always deal with them later, and for now just make this change.

PR #266, opened 2023-05-25
[Speeds]: Add interface for speed and distance messages

ad3154:
## What's new?

You can optionally command some machines to automatically drive at certain speeds, and/or readout the machine's current speed and distance traveled as broadcast by the TECU. Exciting! (Though you may need some combination of other messages for this to work as well depending on your machine, such as the guidance messages.)

 **Please use extreme caution if you attempt to control a machine with this interface. Machines may move without warning!** 

As always, please review our license before using this, as you are agreeing to it by obtaining this software.

### Changes

* Adds a new interface that parses and allows transmitting:
    - Machine selected speed
    - Ground-based speed
    - Wheel-based speed
    - Machine selected speed commands

Basically there are a few things to know with this change.

* Many machines will broadcast their current speed and distance in the "Machine Selected Speed Message" (MSS), along with other important info for controlling the machine or writing your own guidance controller. The machine selected speed is the tractor's authoritative speed generally. Use the source information in the message to know if the speed is from GNSS, wheel speed, etc.
* Many machines accept the "Machine Selected Speed Command" message to automatically drive at a specified speed.
* The other speeds are more for reference, as some machines have multiple speed sources. For example, the machine selected speed could be using GNSS/Navigation based speed, but you might also want to read the radar based (ground-based) speed for some reason to make sure they are similar, or if the GNSS quality is poor. Or, perhaps you are writing the  TECU software yourself, you might want to blend those speeds together with a filter or something to produce your own machine selected speed.

PR #268, opened 2023-06-02
[#253]: Add maintain power interface

### LINKED ISSUE #253: Key switch state & maintain power functionality
As defined in ISO11783-7 and [here](https://www.isobus.net/isobus/pGNAndSPN/13415?type=PGN), it is possible to make requests to the tractor ECU to maintain (ECU) power for another 2 seconds. This could be helpful when your application needs time to shutdown.

Do we want to implement this on the stack side? E.g. where you can register an (async) listener for when the [key switch](https://www.isobus.net/isobus/pGNAndSPN/13421?type=PGN) is turned off, and the stack makes sure the power stays on for as long as the listener(s) are busy? Probably with a (configurable) timeout of like 30 secs, incase the listener(s) accidently come into an unrecoverable state.

ad3154:
## What's new

* Adds a maintain power interface (for PGN 65095).
    - Allows sending and receiving the maintain power message. Parses out the SPNs within the message for your easy use.
    - Re-implements a tiny subset of the `SpeedMessagesInterface` from #266 to allow parsing the key switch state if the user is not using the `SpeedMessagesInterface` to manage it themselves
    - Allows users to get a callback when receiving the maintain power message (useful if you are programming a tractor ECU)
    - Allows users to get a callback when the key transitions from the "not off" to "off" state

### Description of the message

The message this interface manages can be sent by any CF connected to the implement bus requesting that the Tractor ECU not switch off the power for 2 seconds after it has received the wheel-based speed and distance message indicating that the ignition has been switched off. The message also includes the connected implement(s) operating state.

So basically, if the operator turns the key off while you are doing important implement things, you can ask that the TECU give you a little bit more ECU or actuator power before it shuts it off.

If you set this interface up when you start your program, and tell it how long you need power for in the case of an unexpected key-off, this interface will automatically transmit the maintain power message for that long when it detects the key transitioning to the off state.

This is a user-space interface (rather than being in the network manager) so it needs to have its update routine called periodically from your application to work properly, similar to `SpeedMessagesInterface`.

You can also manually send this message to report the various implement states within it to the rest of the bus. These include the following suspect parameter numbers:

- SPN 7447, Implement in-work state
- SPN 1871, Implement ready-to-work state
- SPN 1870, Implement park state
- SPN 1869, Implement transport state

You can visit [www.isobus.net](https://www.isobus.net) to learn more about those if you are interested, or reference the comments on each in the header file for this new interface.

Closes #253

### DIFF
--- a/isobus/include/isobus/isobus/isobus_maintain_power_interface.hpp
+++ b/isobus/include/isobus/isobus/isobus_maintain_power_interface.hpp
@@ -1,0 +1,174 @@
+//================================================================================================
+/// @file isobus_maintain_power_interface.hpp
+///
+/// @brief Defines an interface for sending and receiving the maintain power message (PGN 65095).
+/// @details This interface is for managing the maintain power message, which is a
+/// message sent by any CF connected to the implement bus requesting that the Tractor
+/// ECU (TECU) not switch off the power for 2s after it has received the wheel-based speed and
+/// distance message indicating that the ignition has been switched off.
+/// The message also includes the connected implement(s) operating state.
+/// You can choose if the TECU maintains actuator power independently of ECU power as well.
+/// You might want to maintain actuator power to ensure your section valves close when keyed off.
+/// @author Adrian Del Grosso
+///
+/// @copyright 2023 Adrian Del Grosso
+//================================================================================================
+#ifndef ISOBUS_MAINTAIN_POWER_INTERFACE_HPP
+#define ISOBUS_MAINTAIN_POWER_INTERFACE_HPP
+
+#include "isobus/isobus/can_internal_control_function.hpp"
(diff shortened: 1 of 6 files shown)


isobus/include/isobus/isobus/isobus_maintain_power_interface.hpp
```diff
@@ -0,0 +1,283 @@
...
+
+			enum class MaintainActuatorPower
+			{
+				NoFurtherRequirementForPWR = 0,
+				RequirementFor2SecondsMoreForPWR = 1,
+				Reserved = 2,
+				DontCare = 3
+			};
+
+			enum class MaintainECUPower
```
GwnDaan, line 84: Missing docs here as well

GwnDaan (approved):

PR #274, opened 2023-06-16
Add missing includes for gcc 13

pbert519:
I tried to compile the stack on fedora with  gcc 13 and had some missing includes.


ad3154 (dismissed):

ad3154:
Hello! Thanks for this contribution. Approved, but just needs clang-format applied to match the repo's style/format.

`sudo apt install clang-format`

Then inside the repo:
`find . -iname *.hpp -o -iname *.cpp | xargs clang-format -i`  and append the changes to your commit. In this case it will just alphabetize cstdint after array.

ad3154 (approved):

PR #282, opened 2023-06-23
[CF]: Remove static lists from control functions

GwnDaan:
This PR moves the static lists located in the InternalControlFunction and PartneredControlFunction classes to a centralized place in the manager class. This should make it easier to refactor away from the singleton pattern for the CANNetworkManager.

### DIFF
--- a/isobus/src/can_internal_control_function.cpp
+++ b/isobus/src/can_internal_control_function.cpp
@@ -11,14 +11,12 @@
 #include "isobus/isobus/can_internal_control_function.hpp"
 
 #include "isobus/isobus/can_constants.hpp"
+#include "isobus/isobus/can_network_manager.hpp"
 
 #include <algorithm>
 
 namespace isobus
 {
-	std::vector<std::shared_ptr<InternalControlFunction>> InternalControlFunction::internalControlFunctionList;
-	bool InternalControlFunction::anyChangedAddress = false;
-
 	InternalControlFunction::InternalControlFunction(NAME desiredName, std::uint8_t preferredAddress, std::uint8_t CANPort) :
 	  ControlFunction(desiredName, NULL_CAN_ADDRESS, CANPort, Type::Internal),
 	  stateMachine(preferredAddress, desiredName, CANPort)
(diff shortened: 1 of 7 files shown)


isobus/src/can_network_manager.cpp
```diff
@@ -320,9 +319,16 @@ namespace isobus
...
 
-	void CANNetworkManager::on_control_function_created(std::shared_ptr<ControlFunction>, CANLibBadge<ControlFunction>)
+	void CANNetworkManager::on_control_function_created(std::shared_ptr<ControlFunction> controlFunction, CANLibBadge<ControlFunction>)
 	{
-		//! @todo implement this when we stop using the dedicated internal/partner control functions lists in their respective classes
+		if (controlFunction->get_type() == ControlFunction::Type::Internal)
+		{
+			internalControlFunctions.push_back(std::static_pointer_cast<InternalControlFunction>(controlFunction));
+		}
+		else if (controlFunction->get_type() == ControlFunction::Type::Partnered)
```
ad3154, line 328: Nit:
```suggestion
		if (ControlFunction::Type::Internal == controlFunction->get_type())
		{
			internalControlFunctions.push_back(std::static_pointer_cast<InternalControlFunction>(controlFunction));
		}
		else if (ControlFunction::Type::Partnered == controlFunction->get_type())
```

isobus/src/can_network_manager.cpp
```diff
@@ -290,6 +280,15 @@ namespace isobus
 
 	void CANNetworkManager::on_control_function_destroyed(std::shared_ptr<ControlFunction> controlFunction, CANLibBadge<ControlFunction>)
 	{
+		if (controlFunction->get_type() == ControlFunction::Type::Internal)
+		{
+			internalControlFunctions.erase(std::remove(internalControlFunctions.begin(), internalControlFunctions.end(), controlFunction), internalControlFunctions.end());
+		}
+		else if (controlFunction->get_type() == ControlFunction::Type::Partnered)
```
ad3154, line 287: Nit:
```suggestion
		if (ControlFunction::Type::Internal == controlFunction->get_type())
		{
			internalControlFunctions.erase(std::remove(internalControlFunctions.begin(), internalControlFunctions.end(), controlFunction), internalControlFunctions.end());
		}
		else if (ControlFunction::Type::Partnered == controlFunction->get_type())
```

isobus/src/can_internal_control_function.cpp
```diff
@@ -28,79 +26,23 @@ namespace isobus
...
-	bool InternalControlFunction::destroy(std::uint32_t expectedRefCount)
-	{
-		std::unique_lock<std::mutex> lock(controlFunctionProcessingMutex);
-		internalControlFunctionList.erase(std::find(internalControlFunctionList.begin(), internalControlFunctionList.end(), shared_from_this()));
-		lock.unlock();
-
-		return ControlFunction::destroy(expectedRefCount);
-	}
-
-	std::shared_ptr<InternalControlFunction> InternalControlFunction::get_internal_control_function(std::uint32_t index)
```
ad3154, line 45: I might have mentioned before, but I think at some point (especially when there's possibly > 1 network manager) I think we will want a way for users to be able to look up their previously created CF objects. Not by index probably, but maybe some other way...? Just thinking out loud, doesn't really affect this PR.

isobus/src/can_network_manager.cpp
```diff
@@ -557,15 +558,14 @@ namespace isobus
...
-					    (PartneredControlFunction::partneredControlFunctionList[i]->check_matches_name(NAME(claimedNAME))))
+					if ((partner->get_can_port() == rxFrame.channel) &&
+					    (partner->check_matches_name(NAME(claimedNAME))))
 					{
-						PartneredControlFunction::partneredControlFunctionList[i]->address = claimedAddress;
-						PartneredControlFunction::partneredControlFunctionList[i]->controlFunctionNAME = NAME(claimedNAME);
-						foundControlFunction = std::shared_ptr<ControlFunction>(PartneredControlFunction::partneredControlFunctionList[i]);
+						partner->address = claimedAddress;
+						partner->controlFunctionNAME = NAME(claimedNAME);
+						foundControlFunction = partner;
```
ad3154, line 568: At some point I think we lost the logging message "A partner with name X has claimed address Y on channel Z" which might be nice to have when reading the debug logs so that users can tell by its absence if their NAME filter's didn't match any ECUs? Or maybe it might make sense to check the list of partners at a low frequency and emit a message if their address is invalid? Open to thoughts on that.

ad3154 (commented):
Made a few comments, overall good stuff - a small step to move us closer to un-static parts of the stack.

Side note, I am really impressed with how performant the library is. I tested this using the VT and TC example code with a real terminal (with a PEAK adapter) and my CPU usage on my laptop was 0.3 percent.

isobus/src/can_internal_control_function.cpp
GwnDaan: Yes, I agree. We expose a const-reference to the vector of CF's I suppose, that maybe is a cleaner approach than limiting to a getter by index?

isobus/src/can_network_manager.cpp
GwnDaan: > Or maybe it might make sense to check the list of partners at a low frequency and emit a message if their address is invalid? 

I'd say that is up to the user to find out. They should be able to make such a function really easy themselves when we expose the list of partners.

> At some point I think we lost the logging message "A partner with name X has claimed address Y on channel Z" which might be nice to have when reading the debug logs so that users can tell by its absence if their NAME filter's didn't match any ECUs?

Hmm, it seems to be at debug level now: 
https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/72a069bf0d7f2584c00f53220768cd1fe3459461/isobus/src/can_network_manager.cpp#L660
It might make sense to bump it to the info level instead, and then use a message along the lines of "A partner with name X has claimed address Y on channel Z" yeah.

isobus/src/can_internal_control_function.cpp
ad3154: That seems like it would probably be ok? Then they get pretty much do whatever they need to. Doesn't need to affect this PR, just something to keep in mind I guess.

GwnDaan: I implemented it in commit https://github.com/Open-Agriculture/AgIsoStack-plus-plus/pull/282/commits/d550dc6c7385f6b63ed9925cdada57c0f3a5f332 of this PR while I was at it

GwnDaan:
Added the getters for the partnered and internal control function lists and improved the logging for address claiming in https://github.com/Open-Agriculture/AgIsoStack-plus-plus/pull/282/commits/f8ea5d440a169f70289d7f0d5fd2be248616486d

GwnDaan:
Though maybe I put too many of the address claiming logs to the "info" level, idk?

ad3154 (approved):

PR #292, opened 2023-07-23
[Examples]: Seeder Example

ad3154:
## What's New:

* Added an advanced example that makes use of many protocols including TC, VT, Diagnostic protocol, and the whole transport layer of course.
* This example simulates a rudimentary planter/seeder.

This is certainly not meant to be the 100% best way to lay out an application using the stack, but hopefully it has some interesting use of the various components and one way a user might choose to organize their app.

It's also probably not 100% bug free due to its size, but I wanted to open the PR to get it out there and into people's hands.

### Features

* More complex VT/UT object pool interactions
    * Automatic unit switching and scaling
    * Useful alarm masks
    * Lots of buttons, object pointers, creative use of transparency, etc
* TC integration with VT's displayed statuses
    * Examples of button presses changing task controller values
    * Auto/Manual mode reporting to TC
* Automatic handling of multiple speed sources
    * Working speedometer
    * Alarm when no speeds are detected
* Displays fun information that the stack provides, such as CAN bus load in the object pool
* Demonstrates a simple way to handle filtering updates to only currently shown VT objects
* Demonstrates how our ProcessingFlags can be used to manage VT object updates
* Fixes a little undefined behavior in IOP file hashing.

### Screenshots

image

image

image

image


examples/seeder_example/vt_application.cpp
```diff
@@ -0,0 +1,1149 @@
...
+	}
+	std::cout << "Loaded object pool from BasePool.iop" << std::endl;
+
+	// Generate a unique version string for this object pool (this is optional, and is entirely application specific behavior)
+	std::string objectPoolHash = isobus::IOPFileInterface::hash_object_pool_to_version(objectPool);
+
+	VTClientInterface.set_object_pool(0, isobus::VirtualTerminalClient::VTVersion::Version4, objectPool.data(), objectPool.size(), objectPoolHash);
+	softkeyEventListener = VTClientInterface.add_vt_soft_key_event_listener([this](const isobus::VirtualTerminalClient::VTKeyEvent &event) { this->handle_vt_key_events(event); });
+	buttonEventListener = VTClientInterface.add_vt_button_event_listener([this](const isobus::VirtualTerminalClient::VTKeyEvent &event) { this->handle_vt_key_events(event); });
+	numericValueEventListener = VTClientInterface.add_vt_change_numeric_value_event_listener([this](const isobus::VirtualTerminalClient::VTChangeNumericValueEvent &event) { this->handle_numeric_value_events(event); });
```
GwnDaan, line 94: I think you can remove the lambda functions here and just pass the functions directly to the `add_..._listener` functions

examples/seeder_example/vt_application.cpp
```diff
@@ -0,0 +1,1149 @@
...
+
+	for (std::uint32_t i = 0; i < static_cast<std::uint32_t>(UpdateVTStateFlags::NumberOfFlags); i++)
+	{
+		txFlags.set_flag(i); // Set all flags to bring the pool up with a known state
+	}
+
+	speedMessages.initialize();
+	machineSelectedSpeedEventHandle = speedMessages.get_machine_selected_speed_data_event_publisher().add_listener([this](const std::shared_ptr<isobus::SpeedMessagesInterface::MachineSelectedSpeedData> mssData, bool changed) { this->handle_machine_selected_speed(mssData, changed); });
+	groundBasedSpeedEventHandle = speedMessages.get_ground_based_machine_speed_data_event_publisher().add_listener([this](const std::shared_ptr<isobus::SpeedMessagesInterface::GroundBasedSpeedData> gbsData, bool changed) { this->handle_ground_based_speed(gbsData, changed); });
+	wheelBasedSpeedEventHandle = speedMessages.get_wheel_based_machine_speed_data_event_publisher().add_listener([this](const std::shared_ptr<isobus::SpeedMessagesInterface::WheelBasedMachineSpeedData> wbsData, bool changed) { this->handle_wheel_based_speed(wbsData, changed); });
```
GwnDaan, line 105: Same for these lambda functions, they seem redundant

examples/seeder_example/vt_application.hpp
```diff
@@ -0,0 +1,255 @@
...
+	isobus::VirtualTerminalClient::VTVersion utVersion = isobus::VirtualTerminalClient::VTVersion::ReservedOrUnknown; ///< Stores the UT's version for display in the pool
+	isobus::TaskControllerClient::Version tcVersion = isobus::TaskControllerClient::Version::Unknown; ///< Stores the TC's version for display in the object pool
+	isobus::SpeedMessagesInterface speedMessages; ///< Interface for reading speed from the bus
+	std::shared_ptr<isobus::DeviceDescriptorObjectPool> ddop = nullptr; ///< Stores our application's DDOP
+	std::shared_ptr<std::function<void(const std::shared_ptr<isobus::SpeedMessagesInterface::MachineSelectedSpeedData> &, const bool &)>> machineSelectedSpeedEventHandle; ///< Handle for MSS events
+	std::shared_ptr<std::function<void(const std::shared_ptr<isobus::SpeedMessagesInterface::GroundBasedSpeedData> &, const bool &)>> groundBasedSpeedEventHandle; ///< Handle for ground based speed events
+	std::shared_ptr<std::function<void(const std::shared_ptr<isobus::SpeedMessagesInterface::WheelBasedMachineSpeedData> &, const bool &)>> wheelBasedSpeedEventHandle; ///< Handle for wheel based speed events
```
GwnDaan, line 242: Maybe use `std::shared_ptr<void>` here for nicer formatting on smaller screen, as it doesn't really matter what the function definition is here

examples/seeder_example/vt_application.cpp
```diff
@@ -0,0 +1,1149 @@
...
+{
+	switch (event.keyEvent)
+	{
+		case isobus::VirtualTerminalClient::KeyActivationCode::ButtonUnlatchedOrReleased:
+		{
+			switch (event.objectID)
+			{
+				case home_Key:
+				{
+					set_currently_active_screen(ActiveScreen::Main);
```
GwnDaan, line 133: I feel like we should prefer the `get_visible_data_mask()` getter for checking if a screen is active instead, there could perhaps be some proprietary method of switching data masks, though unlikely, you never know haha

GwnDaan (changes requested):
This example is dope! Good job! I will try to test out later this week, for now a few comments I had while scanning over the code. Also Sonar seems to have a few recommendations here and there, might be worth looking into :)

ad3154:
> This example is dope! Good job! I will try to test out later this week, for now a few comments I had while scanning over the code. Also Sonar seems to have a few recommendations here and there, might be worth looking into :)

One thing I have learned is that sonar really hates the `#defines` in the object pool generated code 😅I'll take a closer look at its other comments.

examples/seeder_example/vt_application.cpp
ad3154: I could try doing that. I guess the idea was that if we got disconnected from the VT that there'd be a way to go back to the last active screen by setting the flag for updating the data mask instead of always ending up at the default data mask in the pool. Doesn't really matter much for this simple example but in a larger pool it might be nice, IDK

GwnDaan:
> > This example is dope! Good job! I will try to test out later this week, for now a few comments I had while scanning over the code. Also Sonar seems to have a few recommendations here and there, might be worth looking into :)
> 
> One thing I have learned is that sonar really hates the `#defines` in the object pool generated code 😅I'll take a closer look at its other comments.

Hahah I saw, maybe we can add all the object pool generated files to the ignored list somehow?

examples/seeder_example/vt_application.cpp
ad3154: Am I missing something on these? 
image

GwnDaan: Huhh haha, are you able to compile the vt version3_object_pool example? We also use direct calls there: https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/52033f9874c8313557a59bc7bda4c8e39dd04530/examples/virtual_terminal/version3_object_pool/main.cpp#L142-L143

GwnDaan: Maybe try linking `this->handle_vt_key_events` instead? It being a non-static function of the class seems the only difference to me

examples/seeder_example/vt_application.cpp
GwnDaan: Hmm, irl I probably would prefer to being reset to the main page instead. I mean what would be the odds of getting disconnected without it being on purpose? Maybe the use cases on my machines are not that diverse, but still idk if it's worth the hassle of keeping our own `active_screen` tracking system

examples/seeder_example/vt_application.cpp
ad3154: Same results unfortunately. I think the one being static must make all the difference 😣
image

The lambdas might have to stay unfortunately

GwnDaan: Ahh I see, `a pointer to a bound function may only be used to call the function` is the reason why that won't work. The lambda is fine, it would've just made it more readable but this lambda function still is perfectly fine with me :)

GwnDaan:
Quick update, I bumped into an issue where automatic section control by the TC server did not seem to work... I'll have to investigate further

examples/seeder_example/section_control_implement_sim.cpp
```diff
@@ -0,0 +1,452 @@
...
+					{
+						value |= sim->get_section_setpoint_state(i + sectionIndexOffset) << (2 * i);
+					}
+					else
+					{
+						value |= (static_cast<std::uint32_t>(0x03) << (2 * i));
+					}
+				}
+			}
+			break;
```
GwnDaan, line 367: question: @ad3154 is setting the setpoint work state from the client the right thing to do? I have a feeling the server would be the only one determining the setpoint, but can't seem to find the details in the standard.

ad3154: Yeah, so it sounds like in general a TC is both entitled to request all settable values, and is likely to actually based on what I've been hearing. It seems like it would generally request them in case there's some other means of those values being manipulated outside of the TC connection, like manually setting the target rate from the implement's UT is a good example. Seems maybe a little odd for section states but apparently it's pretty common.

GwnDaan (approved):

PR #308, opened 2023-08-27
[Core]: Added a way to get callbacks when any CF goes online or offline

### LINKED ISSUE #210: Callback for partner function status changes
If a partner function takes a long time to come up, or goes down then comes back on a different address, I want to be able to send some messages to it straight away.

A callback to inform when a partner function changes status, either comes up, goes down, changes address, and I'm sure there's more:

```c
enum partner_control_function_status
{
    up,
    down,
    moved,
    // etc.
}

void status_callback(PartnerControlFunction * partner, partner_control_function_status status, void *parentPointer)
{
}

CANNetworkManager::CANNetwork.add_partner_control_function_status_callback(partner, &status_callback, nullptr);
```

**Describe alternatives you've considered**
I did check the code, but the best I could find was polling `get_address_valid`, and that only gives the initial bring-up, not later changes.

ad3154:
### What's New:

* This addresses issue #210 and provides a way to get callbacks from the network manager when a control function changes state.
* Overall, this reduces the need to poll the state of your partners in a consuming application.
* Also fixed some typos.

Closes #210

### DIFF
--- a/isobus/include/isobus/isobus/can_network_manager.hpp
+++ b/isobus/include/isobus/isobus/can_network_manager.hpp
@@ -140,14 +140,12 @@ namespace isobus
 
 		/// @brief Use this to get a callback when a control function goes online or offline.
 		/// This could be useful if you want event driven notifications for when your partners are disconnected from the bus.
-		/// @param[in] controlFunction The control function you want callbacks for
-		/// @param[in] callback The callback you want to be called when the specified control function changes state
-		void add_control_function_status_change_callback(std::shared_ptr<ControlFunction> controlFunction, ControlFunctionStateCallback callback);
+		/// @param[in] callback The callback you want to be called when the any control function changes state
+		void add_control_function_status_change_callback(ControlFunctionStateCallback callback);
 
 		/// @brief Used to remove callbacks added with add_control_function_status_change_callback
-		/// @param[in] controlFunction The control function you want to stop receiving callbacks for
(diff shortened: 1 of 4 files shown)


isobus/include/isobus/isobus/can_network_manager.hpp
```diff
@@ -137,6 +137,17 @@ namespace isobus
 		/// @param[in] controlFunction The control function that was created
 		void on_control_function_created(std::shared_ptr<ControlFunction> controlFunction, CANLibBadge<ControlFunction>);
 
+		/// @brief Use this to get a callback when a control function goes online or offline.
+		/// This could be useful if you want event driven notifications for when your partners are disconnected from the bus.
+		/// @param[in] controlFunction The control function you want callbacks for
+		/// @param[in] callback The callback you want to be called when the specified control function changes state
+		void add_control_function_status_change_callback(std::shared_ptr<ControlFunction> controlFunction, ControlFunctionStateCallback callback);
```
GwnDaan, line 144: Hmm, wouldn't it make more sense to just register a callback for all controlfunctions instead of binding it to a single one. It allows for greater flexability in use cases, and doesn't affect performance I think

ad3154: Hm, I didn't think about that. That would probably be even more performant. I like it - I will change it.

GwnDaan (approved):

PR #317, opened 2023-09-02
[DDOP]: Add validation for DDOP object child references

### LINKED ISSUE #316: [TC]: Validate referable child objects for DETs consists only of DPDs and DPTs
ISO11783-10's Section A.3 defines that the list of referable child objects for a device element object consists only of DPDs and DPTs, which is not being properly validated. 

Need to do the following:
- Correct the TC example code
- Add to our DDOP validation that all children of DET objects are only DPDs or DPTs.

_Originally posted by @ad3154 in https://github.com/Open-Agriculture/AgIsoStack-plus-plus/discussions/307#discussioncomment-6889423_

ad3154:
* Added validation of DET child references' object types. Added sanity checks on other objects' child types even though those are  essentially forced to be correct.
* Reworked a couple things in the unit tests to prevent some tests from interfering with other tests specifically when not using CTest.

Closes #316

### DIFF
--- a/isobus/src/isobus_device_descriptor_object_pool.cpp
+++ b/isobus/src/isobus_device_descriptor_object_pool.cpp
@@ -943,6 +943,16 @@ namespace isobus
 								retVal = false;
 								break;
 							}
+							else if ((task_controller_object::ObjectTypes::DeviceProcessData != child->get_object_type()) &&
+							         (task_controller_object::ObjectTypes::DeviceProperty != child->get_object_type()))
+							{
+								CANStackLogger::error("[DDOP]: Object %u has child %u which is an object type that is not allowed.",
+								                      currentDeviceElement->get_child_object_id(i),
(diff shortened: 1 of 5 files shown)


isobus/src/isobus_device_descriptor_object_pool.cpp
```diff
@@ -943,6 +943,16 @@ namespace isobus
 								retVal = false;
 								break;
 							}
+							else if ((task_controller_object::ObjectTypes::DeviceProcessData != child->get_object_type()) ||
```
GwnDaan, line 946:
```suggestion
							else if ((task_controller_object::ObjectTypes::DeviceProcessData != child->get_object_type()) &&
```

GwnDaan (changes requested):

GwnDaan (approved):

PR #324, opened 2023-09-09
Add support for Arduino, Teensy 4 and 4.1

### LINKED ISSUE #264: Support Arduino IDE
There is some demand for being able to program a microcontroller using Arduino IDE. I think it could be a good way to get started quickly with this stack. Though I must say I don't have much experience using it.

- [x] Verify full functionality through Arduino IDE
- [ ] Write documentation on how to get started
- [x] Adapt an example for 1:1 use with a microcontroller (e.g. ESP32/Teensy) running Arduino Core.

When we commit to support the Arduino IDE, there will be yet another possible way a commit can break something. And since we are constantly making changes at the moment, it could really slowdown progress. Hence waiting with this until a stable version is released may be something to take into considiration.

ad3154:
## What's new

Added lots of changes to allow the stack to build without threading under the Arduino IDE. Added CAN plugin for Teensy hardware using the FlexCAN T4 api.

Changes include:

* Added single threaded CANHardwareInterface option. Changed dynamic casts to static casts to completely eliminate rtti, which shouldn't be an issue since the type can be inferred in those cases where we were using dynamic_pointer_cast.
* Added preprocessor checks to optionally compile out all threading. Added a script to automatically generate an Arduino library from the content of the repo.
* Added Arduino example files.
* Added a python script to automatically generate an Arduino library based on the content of the repo.

Tested with Teensy 4.1 hardware and SN65HVD230 transceiver.

### Follow-up tasks

- [x] Create new repository to store the latest-generated Arduino library version of the stack
- [ ] Submit library to the Arduino [library registry](https://github.com/arduino/library-registry)
- [ ] Add CI pipeline for building with Arduino
- [ ] Investigate adding pipeline that can automatically generate the Arduino library and commit it to the other repo, plus open a PR
- [ ] Create documentation for using the library on Arduino

### How to use/test this right now

Until all the tasks above are done, here's how to try this on Teensy 4.1 on Windows.

1. Install [Arduino IDE](https://www.arduino.cc/en/software)
2. Add Teensy board support to Arduino IDE by following [these instructions](https://www.pjrc.com/teensy/td_download.html)
3. Open the [example file](https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/arduino/examples/arduino_example/arduino_example.ino) in your Arduino IDE.
4. Select your board as shown:
image
5. Ensure optimizations are set to "Smallest Code"
image
6. There's some kind of bug in the Teensy's core, where if you are not building with the default optimization settings, the board fails to start up. Some [attempts](https://github.com/PaulStoffregen/cores/pull/673) appear to have been made to fix this but it remains an issue for me at least. To ensure your board starts up, add `#pragma GCC optimize ("O2")` to the teensy's `startup.c` file which is likely located at the following location: `C:\Users\<your user name>\AppData\Local\Arduino15\packages\teensy\hardware\avr\1.58.1\cores\teensy4`
image
7. Run the `generateArduinoLibrary.py` script with Python in the root of the repo. I usually do this in IDLE because I'm a Python neophyte basically. I know C++ things, what can I say.
image
9. Copy the newly created `arduino_library` folder that the script creates to your Arduino `libraries` directory. You may need to create this directory if it doesn't exist, but it's usually located at `C:\Users\<your user name>\Documents\Arduino\libraries`. Maybe rename the `arduino_library` folder to AgIsoStack or something if you want.
10. Build the sketch, and upload it to your board 🙂

You'll also need a CAN transceiver connected to your Teensy's CAN1 pins.

Closes #264


hardware_integration/include/isobus/hardware_integration/can_hardware_interface_single_thread.hpp
GwnDaan, line 1: Hmm, I guess this approach is good for proof of concept, but the overlap between the multithreaded and this new single threaded file is quite big. I'd argue we should add a "threading" layer on top of this single thread instead for the other one, what are your thoughts? I'm fine with putting it on the backlog for now

GwnDaan (approved):

hardware_integration/include/isobus/hardware_integration/can_hardware_interface_single_thread.hpp
ad3154: Yeah, there is a fair bit of overlap. If we could make a threading layer at some point that extends a common version of `CANHardwareInterface`, I agree that would be cleanest and easiest to maintain.

ad3154:
Added to Arduino library registry here: https://github.com/arduino/library-registry/pull/3320

PR #326, opened 2023-09-16
Add ESP32 + PlatformIO Tutorial and Example

### LINKED ISSUE #263: Create documentation for PlatformIO with ESP-IDF
We get more and more questions about the use of this stack in combination with an ESP (e.g. ESP32). It is known to work with PlatformIO and ESP-IDF, but it is missing some documentation. Therefore it could be useful to have a tutorial page or two dedicated to this. 

- [ ] Documentation on how to use PlatformIO with this stack
- [ ] Example that runs on any ESP with PlatformIO, probably an adapted VT object pool example?

ad3154:
## What's new

* Added tutorial page for using our stack on ESP32 with PlatformIO. 
* Added ESP32 + PIO example files.

I would like some feedback on this new tutorial before merging it, ideally, so you can find a draft of it [here](https://delgrossoengineering.com/isobus-tutorial/Tutorials/ESP32%20PlatformIO)

Any and all comments would be welcome.

Closes #263


AndreaInverardi:
Built "esp32_platformio_object_pool" example, as in the tutorial. Deployed on some ESP32 module. I can see some CAN frames been sent at boot. If I check serial port of ESP32 module it says every 5 seconds:

"E (391530) task_wdt: Task watchdog got triggered. The following tasks/users did not reset the watchdog in time:"

Tried also connected to some tractor's virtual terminal. It do not detect it.

ad3154:
> Built "esp32_platformio_object_pool" example, as in the tutorial. Deployed on some ESP32 module. I can see some CAN frames been sent at boot. If I check serial port of ESP32 module it says every 5 seconds:
> 
> "E (391530) task_wdt: Task watchdog got triggered. The following tasks/users did not reset the watchdog in time:"
> 
> Tried also connected to some tractor's virtual terminal. It do not detect it.

Thanks for this feedback - I will see what I can do. I can sometimes repeat this same symptom, and have seen some successes with replacing some calls to `std::this_thread::sleep_for` with calls directly to FreeRTOS on ESP32 with `vTaskDelay` for example. We might have some success `#define`ing around these when `ESP_PLATFORM` is defined... or maybe we can use CMake and some clever abstraction to optionally compile in different kind of thread delays (this might be the cleanest solution...).

GwnDaan:
Hi @AndreaInverardi @ad3154, sorry for the delay, the fix for your issues should already be in the stack: https://github.com/Open-Agriculture/AgIsoStack-plus-plus/pull/91/commits/29fe04ff5d1c9e66880c9d2fc496c0972c0b5744

Basically the fefault FreeRTOS tick is 10ms, but the stack tries to update internally with 4ms by default. Hence the watchdog gets triggered as the thread is never allowed to switch. 

There are 2 solutions:
- Set the internal tick rate to 10ms in your init: `isobus::CANHardwareInterface::set_number_of_can_channels(10)`
- Increase the tickrate of the your FreeRTOS environment to > 250Hz to match internal tick rate. For PlatformIO with ESP32 this can be done inside the menuconfig: `pio run -t menuconfig`. component config -> FreeRTOS -> Tick rate (hz), IIRC.

I prefer the latter to minimize platform specific issues, but both should work fine I think

ad3154:
Added updates to the ESP32 instructions to run `menuconfig` and adjust both the tick rate and some default stack sizes.
When testing on my end this seemed to resolve all instances of runtime crashes I was seeing.

Check out the new section `Set up your ESP32’s OS and PThread options` in the [draft](https://delgrossoengineering.com/isobus-tutorial/Tutorials/ESP32%20PlatformIO).

GwnDaan (approved):

PR #333, opened 2023-10-07
Switch to debug builds in CI to catch asserts in unit tests

GwnDaan:
Tests were failing in the CI pipeline because it couldn't catch any asserts as it was being build in "Release" mode, which optimizes out the asserts. I think it's reasonable to build the pipeline in debug to pass the unit tests, opinions?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
     types: [opened, synchronize, reopened]
 
 env:
-  BUILD_TYPE: Release
+  BUILD_TYPE: Debug
 
 jobs:
   build_test_analyze:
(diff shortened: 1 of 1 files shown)


ad3154 (approved):
Interesting... I think it's fine. Looking forward to when we might want to do real releases with pre-compiled binary artifacts, maybe it would be nice to have some on-tag action that could compile both the release and debug versions of the library? Just thinking out loud on that, definitely out of scope for this PR.

GwnDaan:
> maybe it would be nice to have some on-tag action that could compile both the release and debug versions of the library?

Yeah, that would be nice to have once we have our first real release

PR #339, opened 2023-10-27
[Examples]: Add NMEA2000 generator/parser

GwnDaan:
Added generator and parser example programs to test the libraries NMEA2000 message interface functionality. Note that the output of the parser is not that neat, so you'd probably want to disable some of the listeners to be able to read your console properly...

Also fixed a few small incorrect units applied, basically sometimes 10E... where it should've been 1E... to get the correct output/input


examples/nmea2000/fast_packet_protocol/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+cmake_minimum_required(VERSION 3.16)
+project(fast_packet_protocol_example)
+
+set(CMAKE_CXX_STANDARD 14)
```
ad3154, line 4: Just a note, I opened #340 because apparently there are nicer or at least more modern ways to do this 🙂

ad3154 (approved):
Huge thanks for an excellent review and testing this for me - I really appreciate it!!!

PR #344, opened 2023-11-09
[TP]: Rewrite to aid maintainability and possiblity to expand with other features

GwnDaan:
This is basically a proof of concept that works as far as the unit tests go. Now I seek some feedback if we actually want these changes or not.

What changes:
- added abstract transport data interface
- reworked the session storage to hold the objects themselves
- a lot of refactorizarion such that a session could in theory be initialized and completed for any ECU on the bus


GwnDaan:
> Part of me is wondering if we might want to make even a different class from TransportProtocolManager, that somewhat duplicates the functionality but could be optionally linked in to add a "transport protocol listener"

Yeah I see what you mean, when implementing I tried to keep in mind a third "SessionDirection" which would be something like "ListenOnly" but thats not yet in there. But an approach like yours might work as good as well, though it introduces some dupplication when maintaining the stack. Something we already experience quite a bit with the current 3 transport protocols being very similar in some places.

ad3154:
> though it introduces some duplication when maintaining the stack

It would of course, and each protocol is juuuust different enough to make sharing code between them troublesome. The good thing is, the transport layer is quite stable from an ISO/J1939 requirements perspective, and probably won't need to be messed with too often once it's working (especially if we can ever hit it with the AEF conformance test tool to get it 100% compliant).

Going back to your (paraphrased) goal with this:

>  I basically [wanted] to try to make the TP class better suited for sniffing TP sessions between other ecu's on the bus.

This strikes me as a, lets say, abnormal thing to want in a "normal" ISOBUS application but something many people will want regardless of course, so, one thing I really like about a separate object is that it makes it clear when you use it that you are doing things outside the "core" functionality of the stack - something extra with a clearly defined but separate data path. Like, you really have to opt-in. Plus, then the fairly substantial extra storage associated with it optional, because sniffing could result in us storing (253 + (253 * 252)) = 63756 sessions!!!! (or, (253!)/(2!*(253-2)!)*2 if you want to be pedantic about your equations).

ad3154:
I'm sure we can control the memory cost without making it optionally linkable of course, it's just something to think over and consider. No matter what path we go, we may want to not sniff anything unless specifically configured to do so one way or another.

GwnDaan:
@ad3154, I still think this is a great addition. I changed a lot and in my opinion it is now more modular and future proof. I can easily implement a message 'sniffer' based on this format that doesn't hurt our performance, and is opt-in without too much extra codebase. Though you might still think otherwise, and hence I'm curious for your thoughts on this :)

isobus/src/can_message.cpp
```diff
@@ -39,11 +39,36 @@ namespace isobus
...
 	{
 		return destination;
 	}
 
+	bool CANMessage::has_valid_destination_control_function() const
+	{
+		return nullptr != destination;
+	}
+
+	bool CANMessage::is_destination_global() const
```
GwnDaan, line 57: We might want to rename this to 
```suggestion
	bool CANMessage::is_broadcast() const
```

ad3154:
> @ad3154, I still think this is a great addition. I changed a lot and in my opinion it is now more modular and future proof. I can easily implement a message 'sniffer' based on this format that doesn't hurt our performance/memory usage, and is opt-in without too much extra codebase. Though you might still think otherwise, and hence I'm curious for your thoughts on this :)
> 
> _edit:_ it most-likely still has some bugs in it that need to be catched

Ah, ok, I can take another look and let you know my thoughts 🙂

isobus/src/can_message.cpp
```diff
@@ -39,11 +39,36 @@ namespace isobus
 		return source;
 	}
 
+	bool CANMessage::has_valid_source_control_function() const
+	{
+		return nullptr != source;
```
ad3154, line 44: "Valid" might be inaccurate, it's just not `nullptr`... might need to also check the CF has claimed since the last address claiming session with the appropriate function?

isobus/include/isobus/isobus/can_transport_protocol.hpp
```diff
@@ -125,115 +245,159 @@ namespace isobus
...
+		bool initialize(CANLibBadge<CANNetworkManager>);
+
+		/// @brief Returns if the protocol has been initialized
+		/// @returns true if the protocol has been initialized, otherwise false
+		bool get_initialized() const;
+
+		/// @brief The protocol's terminate function
+		void terminate(CANLibBadge<CANNetworkManager>);
+
+		/// @brief Updates the diagnostic protocol
```
ad3154, line 264:
```suggestion
		/// @brief Updates the transport protocol
```

isobus/src/can_transport_protocol.cpp
```diff
@@ -11,860 +11,1004 @@
...
-			CANNetworkManager::CANNetwork.add_protocol_parameter_group_number_callback(static_cast<std::uint32_t>(CANLibParameterGroupNumber::TransportProtocolData), process_message, this);
+			CANNetworkManager::CANNetwork.add_protocol_parameter_group_number_callback(static_cast<std::uint32_t>(CANLibParameterGroupNumber::TransportProtocolConnectionManagement), process_message, this);
+			CANNetworkManager::CANNetwork.add_protocol_parameter_group_number_callback(static_cast<std::uint32_t>(CANLibParameterGroupNumber::TransportProtocolDataTransfer), process_message, this);
 		}
+		return initialized;
 	}
 
-	void TransportProtocolManager::process_message(const CANMessage &message)
+	void TransportProtocolManager::process_broadcast_announce_message(const std::shared_ptr<ControlFunction> source,
+	                                                                  std::uint32_t pgn,
```
ad3154, line 203: Nit (several places)
```suggestion
	                                                                  std::uint32_t parameterGroupNumber,
```

isobus/src/can_transport_protocol.cpp
```diff
@@ -11,860 +11,1004 @@
...
-			                                 session->parent);
+			CANStackLogger::warn("[TP]: Received an abort (reason=%hu) with no matching session for PGN 0x%05X", static_cast<uint8_t>(reason), pgn);
 		}
 	}
 
-	bool TransportProtocolManager::send_broadcast_announce_message(TransportProtocolSession *session) const
+	void TransportProtocolManager::process_connection_management_message(const CANMessage &message)
 	{
-		bool retVal = false;
+		const auto pgn = message.get_uint24_at(5);
```
ad3154, line 383: Should maybe validate the message size first, otherwise could throw if the length is 1?

isobus/src/can_transport_protocol.cpp
```diff
@@ -11,860 +11,1004 @@
...
+					              message.get_destination_control_function(),
+					              pgn,
+					              reason);
+				}
+			}
+			break;
+
+			default:
+			{
+				CANStackLogger::CAN_stack_log(CANStackLogger::LoggingLevel::Warning, "[TP]: Bad Mux in Transport Protocol Connection Management message");
```
ad3154, line 479: Since we're changing a bunch in this file anyways we might as well consider replacing some of these with the shorter ones.
```suggestion
				CANStackLogger::warn("[TP]: Bad Mux in Transport Protocol Connection Management message");
```

isobus/src/can_transport_protocol.cpp
```diff
@@ -11,860 +11,1004 @@
...
-		return retVal;
 	}
 
-	bool TransportProtocolManager::send_end_of_session_acknowledgement(TransportProtocolSession *session) const
+	void TransportProtocolManager::process_message(const CANMessage &message, void *parent)
 	{
-		bool retVal = false;
+		if (nullptr != parent)
+		{
+			reinterpret_cast<TransportProtocolManager *>(parent)->process_message(message);
```
ad3154, line 599:
```suggestion
			static_cast<TransportProtocolManager *>(parent)->process_message(message);
```
I think we'd always know it'd be safe to static cast it, especially since there's no polymorphism now with this object

isobus/src/can_transport_protocol.cpp
```diff
@@ -11,860 +11,1004 @@
...
+		if (activeSessions.size() >= CANNetworkManager::CANNetwork.get_configuration().get_max_number_transport_protocol_sessions())
 		{
-			reinterpret_cast<TransportProtocolManager *>(parent)->process_message(message);
+			// TODO: consider using maximum memory instead of maximum number of sessions
+			CANStackLogger::warn("[TP]: Replying with abort to Request To Send (RTS) for 0x%05X, configured maximum number of sessions reached.", pgn);
+			send_abort(std::static_pointer_cast<InternalControlFunction>(destination), source, pgn, ConnectionAbortReason::AlreadyInCMSession);
+		}
+		else
+		{
+			if (auto session = get_session(source, destination))
```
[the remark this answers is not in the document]
ad3154: Here, I will point at the books' rationale

image

image

ad3154 (commented):
I do think there's valuable stuff in here, especially in the TP file and it does make some steps towards multiple goals we wanted.

I would like to do some load testing with it just to make sure it still all works OK with multiple sessions going at once, and of both types of sessions to see if we can provoke any bugs.

My main mental sticking point is still that I don't enjoy too many levels of abstract data types on top of abstract data types. I'm still not sure I grasp the benefit of having things like CANMessageDataVector  which inherits from CANMessageData which inherits from a vector, when the data, really, is just a vector. Or CANMessageDataView which inherits from CANMessageData which inherits from CANDataSpan which is really just a templated version of DataSpan, which is just an array of bytes and a size. 

Was your thinking mainly to consolidate CANMessageDataCallback? It just feels like a lot of misdirection which at least kinda feels more confusing to me than just interacting with the native/standard types.

I dunno... I could be convinced I think, maybe I am just not grasping your whole picture here, or maybe you need this to better enable sniffing? I might just need more explaining

isobus/src/can_transport_protocol.cpp
GwnDaan: Hmm, it had a check for this already one layer up. I moved it and made it an early-return to hopefully give a better overview next time

isobus/src/can_transport_protocol.cpp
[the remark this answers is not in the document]
GwnDaan: Ah yeah that's a bummer. I liked the shorthand haha. I'll change it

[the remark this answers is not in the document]
GwnDaan: Okay, hmm. I have a counter argument for this; if we take the assignment out of the conditional we can have a session that is a nullptr. I therefore liked the idea of the assignment inside the if-statement, it means we can never access a session that is a nullptr :)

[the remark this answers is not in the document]
GwnDaan: Also I don't think the book it is really what the book meant. There is no sub-expression inside the if-statement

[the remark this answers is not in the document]
GwnDaan: Though I see why it can be harder too gasp, hmm...

ad3154:
I appreciate your explanations, they do help me see what you mean, and I do feel a bit better about it now... 

Like I mentioned I'd like to run some tests just because I am paranoid about changing TP, but then I think I'm OK with it and will give it one last look over.

Sorry I'm stalling, haha

GwnDaan:
Yeah no worries, I was planning on trying to cover it with unit tests first as well. But I'm not sure where to start to be completely honest. Since the TP probably won't change in the mean time, i'm fine with leaving it open a little longer to be sure there are no issues.

GwnDaan:
In https://github.com/Open-Agriculture/AgIsoStack-plus-plus/pull/344/commits/b0301bd38441bdde02a35071957a737775f8767f I removed the `TransportProtocolSession` dependency, this should make it a lot easier to unit-test the implementation. 

There's still a list I want to do before thinking about merging this into main:
- rewrite ExtendedTransportProtocol to match the new style
- rewrite FastPacketProtocol to match the new style
- unit-test all implementations

GwnDaan:
Replaced by #391

PR #346, opened 2023-11-10
[VT]: Add Virtual Terminal Object Classes

ad3154:
This is intended to be used with a VT server (coming soon? ) or with an object pool editor program. I wanted to open a PR for the ones I have done before the file gets even larger.

- Added classes to represent most VT objects.

- [x] Todo: Unit tests
- [x] Todo: Get Attribute Value functions

Supported objects so far:
- Working Set
- Data Mask
- Alarm Mask
- Container
- Soft Key mask
- Key
- Key Group
- Button
- Input Boolean
- Input String
- Input Number
- Input List
- Output String
- Output Number
- Output List
- Output Line
- Output Rectangle
- Output Ellipse
- Output Polygon
- Output Meter
- Output Linear Bar Graph
- Output Arched Bar Graph
- Picture Graphic
- Number Variable
- String Variable
- Font Attributes
- Line Attributes
- Fill Attributes
- Input Attributes
- Extended Input Attributes
- Object Pointer
- External Object Pointer
- Macro
- Colour Map
- Window Mask

### DIFF
--- a/isobus/include/isobus/isobus/isobus_virtual_terminal_objects.hpp
+++ b/isobus/include/isobus/isobus/isobus_virtual_terminal_objects.hpp
@@ -148,9 +148,10 @@ namespace isobus
 		};
 
 		/// @brief Constructor for a generic VT object. Sets up default values and the pointer to the member object pool
-		/// @param[in] memberObjectPool a reference to the object tree that this object will be a member of
-		/// @param[in] currentColourTable a reference to the current colour table the working set is using
-		VTObject(std::map<std::uint16_t, std::shared_ptr<VTObject>> &memberObjectPool, VTColourTable &currentColourTable);
+		VTObject() = default;
+
+		/// @brief Virtual destructor for a generic VT object
+		virtual ~VTObject() = default;
 
 		/// @brief Returns the VT object type of the underlying derived object
 		/// @returns The VT object type of the underlying derived object
@@ -162,16 +163,17 @@ namespace isobus
 
 		/// @brief Performs basic error checking on the object and returns if the object is valid
 		/// @returns `true` if the object passed basic error checks
-		virtual bool get_is_valid() const = 0;
+		virtual bool get_is_valid(const std::map<std::uint16_t, std::shared_ptr<VTObject>> &objectPool) const = 0;
 
 		/// @brief Sets an attribute and optionally returns an error code in the last parameter
 		/// @param[in] attributeID The ID of the attribute to change
 		/// @param[in] rawAttributeData The raw data to change the attribute to, as decoded in little endian format with unused
 		/// bytes/bits set to zero.
+		/// @param[in] objectPool A map of all objects in the current object pool, keyed by their object ID. Used to validate some object references.
 		/// @param[out] returnedError If this function returns false, this will be the error code. If the function
 		/// returns true, this value is undefined.
 		/// @returns True if the attribute was changed, otherwise false (check the returnedError in this case to know why).
-		virtual bool set_attribute(std::uint8_t attributeID, std::uint32_t rawAttributeData, AttributeError &returnedError) = 0;
+		virtual bool set_attribute(std::uint8_t attributeID, std::uint32_t rawAttributeData, const std::map<std::uint16_t, std::shared_ptr<VTObject>> &objectPool, AttributeError &returnedError) = 0;
 
 		/// @brief Gets an attribute and returns the raw data in the last parameter
 		/// @param[in] attributeID The ID of the attribute to get
@@ -212,10 +214,6 @@ namespace isobus
 		/// @brief Sets the background color attribute of this object
 		/// @param[in] value The new background color attribute for this object (index to the actual color in the color table)
 		void set_background_color(std::uint8_t value);
-
-		/// @brief Returns a VT object from its member pool by ID, or the null id if it does not exist
-		/// @returns The object with the corresponding ID
-		std::shared_ptr<VTObject> get_object_by_id(std::uint16_t objectID) const;
 
 		/// @brief Returns the number of child objects within this object
 		std::uint16_t get_number_children() const;
@@ -268,6 +266,12 @@ namespace isobus
 		/// This is meant to be a faster way to deal with objects that only have a max of 1 child.
 		void pop_child();
 
+		/// @brief Returns a VT object from its member pool by ID, or the null id if it does not exist
+		/// @param[in] objectID The object ID to search for
+		/// @param[in] objectPool The object pool to search in
+		/// @returns The object with the corresponding ID
+		static std::shared_ptr<VTObject> get_object_by_id(std::uint16_t objectID, const std::map<std::uint16_t, std::shared_ptr<VTObject>> &objectPool);
+
 	protected:
 		/// @brief Storage for child object data
 		class ChildObjectData
@@ -288,20 +292,6 @@ namespace isobus
 			std::int16_t yLocation = 0; ///< Relative Y location of the top left corner of the object
 		};
 
-		/// @brief A helper function to swap a child object of specified type with a new object.
-		/// This is meant to be used to replace attributes of which there can be only 1, such as a string variable in an output string.
-		/// @param[in] newID The new child object's ID
-		/// @param[in] typeToRemove The object type of both the current child, and the new child (they must match)
-		/// @returns True if the replacement succeeded, otherwise false
-		bool replace_only_child_of_type(std::uint16_t newID, VirtualTerminalObjectType typeToRemove);
-
-		/// @brief A helper function to get the first child object of the specified type
-		/// @param[in] typeToFind The object type to find
-		/// @returns The object ID of the first child object of the specified type, or NULL_OBJECT_ID if none were found
-		std::uint16_t get_first_child_of_type(VirtualTerminalObjectType typeToFind) const;
-
-		VTColourTable &colourTable; ///< A reference to the current object pool's colour table. Useful for rendering most objects.
-		std::map<std::uint16_t, std::shared_ptr<VTObject>> &thisObjectPool; ///< A pointer to the rest of the object pool. Convenient for lookups by object ID.
 		std::vector<ChildObjectData> children; ///< List of child objects
 		std::uint16_t objectID = NULL_OBJECT_ID; ///< Object identifier. Shall be unique within the object pool.
 		std::uint16_t width = 0; ///< The width of the object. Not always applicable, but often used.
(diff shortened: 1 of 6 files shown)


ad3154:
@GwnDaan I've finished my todos on this for now, so it's ready to review at your convenience.

Despite the amount of code, it's mostly just setters and getters for the various object types, so there's not really that much functionality to worry about. The object pool deserializer and whatnot will come in another PR.

GwnDaan:
> @GwnDaan I've finished my todos on this for now, so it's ready to review at your convenience.
> 
> Despite the amount of code, it's mostly just setters and getters for the various object types, so there's not really that much functionality to worry about. The object pool deserializer and whatnot will come in another PR.

Sounds good! I'll try to review it in the coming days, and then I think I'll just mostly look at the tests and not so much at the rest, as you suggested👍

isobus/include/isobus/isobus/isobus_virtual_terminal_objects.hpp
```diff
@@ -76,6 +86,3886 @@ namespace isobus
...
+			std::int16_t xLocation = 0; ///< Relative X location of the top left corner of the object
+			std::int16_t yLocation = 0; ///< Relative Y location of the top left corner of the object
+		};
+
+		/// @brief A helper function to swap a child object of specified type with a new object.
+		/// This is meant to be used to replace attributes of which there can be only 1, such as a string variable in an output string.
+		/// @param[in] newID The new child object's ID
+		/// @param[in] typeToRemove The object type of both the current child, and the new child (they must match)
+		/// @returns True if the replacement succeeded, otherwise false
+		bool replace_only_child_of_type(std::uint16_t newID, VirtualTerminalObjectType typeToRemove);
```
GwnDaan, line 296: I'm not sure how this can work? Shouldn't you want to simply overwrite the attribute of let's say an output string to achieve this? 

Consider the following (incomplete) pool: 
- Output string (id=20)
- String variable (id=30)
- String variable (id=31)

Then you'd simply want to change the variable reference of the "Output string (id=20)" from "30" to "31"? Instead of replacing things in the object pool like done here?

isobus/include/isobus/isobus/isobus_virtual_terminal_objects.hpp
```diff
@@ -76,6 +86,3886 @@ namespace isobus
...
+
+		/// @brief A helper function to get the first child object of the specified type
+		/// @param[in] typeToFind The object type to find
+		/// @returns The object ID of the first child object of the specified type, or NULL_OBJECT_ID if none were found
+		std::uint16_t get_first_child_of_type(VirtualTerminalObjectType typeToFind) const;
+
+		VTColourTable &colourTable; ///< A reference to the current object pool's colour table. Useful for rendering most objects.
+		std::map<std::uint16_t, std::shared_ptr<VTObject>> &thisObjectPool; ///< A pointer to the rest of the object pool. Convenient for lookups by object ID.
```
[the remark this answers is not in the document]
ad3154: That's fair - the way I'm consuming these classes myself is safe I think but that doesn't mean it would be for others. Might be better to store a weak pointer or something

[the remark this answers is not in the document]
GwnDaan: Or maybe just require an argument in the functions that use it? It doesn't seem to be used that often IIRC

isobus/include/isobus/isobus/isobus_virtual_terminal_objects.hpp
ad3154: Well, it's not actually replacing an object, per-se, so maybe the name could be clearer... let me explain from the beginning.

So, some of the reason why this exists was to try and reduce the amount of functions and variables in each object. The base object has a `std::vector<ChildObjectData>`, so I've tried to make use of that as much as possible. So if an object like an input number can only ever have 1 of certain kinds of references (number variable, and font attributes in this case) I just put them in that vector instead of making special setters and getters for them. I don't know if this is a good thing or a bad thing, but since I have to do a lot of iterating over objects' children recursively in the VT server, it's maybe more consistent? 

Of course then I still had to support the case where someone wanted to change that child, especially something like a `FontAttributes`, so I made this function to do it for me. It replaces the object ID of the old object in the child vector with the object id of the new child in the vector. So it doesn't actually "replace" something in the pool, if you know what I mean, it's just changing the object ID of a child to another object's ID.

So, I guess I think this function is doing as you said: `change the variable reference of the "Output string (id=20)" from "30" to "31"` it's just a little less than obvious.

If you think it would be clearer to rename it, I could do that. Or if you would prefer concrete child variables with named setters and getters I could do that too, it'll just mean a longer file.

GwnDaan: Okay hmm I think I get it? Correct me if I'm wrong.

It now uses the `children` list of an object to store objects 'references' as if they were an actual child of the object. But then you don't actually use it as a child, but instead as a reference? I guess this would work yeah, though that also means we have to filter this `children` list when we want to serialize the object. Also a small downside is it requires at least 4 bytes extra storage as `ChildObjectData` also has the locations which are not used when acting as an objectId reference only. IMO not that big of a deal, but will it be worth the hassle?

To be completely honest, I think for _us in the future_, and also other new maintainers, it might be more consistent and clear if we just have for each attribute a dedicated member field of the class. Then no-one has to guess how it's stored and I think it's the most straight forward. 

I do like the way you are thinking of making the file smaller (and hence less code to maintain), but I think there might be too many edge cases attached to this, for it to worth it hmmm... What do you think?

isobus/include/isobus/isobus/isobus_virtual_terminal_objects.hpp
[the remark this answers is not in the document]
ad3154: It's mostly there for `std::shared_ptr<VTObject> get_object_by_id(std::uint16_t objectID) const;` In my VT server I call this hundreds of times haha. My GUI objects actually inherit from each of these objects, and then also inherit from the GUI framework, so inside their `paint` routine I can get objects from the parent object pool easily.

If we could force people to interact with creating these objects only through some other class, then we could ensure the reference is valid... like, we could make the constructors badged, or private with a `friend` class of the object pool or something. IDK 🤔

isobus/include/isobus/isobus/isobus_virtual_terminal_objects.hpp
ad3154: To be honest I hadn't really given a lot of thought to serializing the pool yet, so that is a good thing to point out. And yes, there is definitely a 4 byte RAM cost per reference.

I'm not particularly attached to my weird child object thing here, so I'm inclined to agree that maybe it'll be less weird/confusing if we just use dedicated fields for non-child references. I can make those changes

vbenso: Gentleman, 

I'm following the progress of this project, awesome work by the way.
About the VTOP, here goes my take. In the test's I've done here (this is a kind of weekend project for experimentation about the subject), the results using a serialized structure to store all the object in RAM, seem to be working. 
It's not as fast I wished it to be, but perhaps can be of some use for you. 
Also some feedback would be appreciated.
I'm using the following to parse a VTOP and, also, as a base to another class that can draw the elements based on the attributes of each into an LCD screen (actually at this development stage using an ESP32.

[opparser.zip](https://github.com/Open-Agriculture/AgIsoStack-plus-plus/files/13429535/opparser.zip)

PS.: Sorry for not using such a modern C++ implementation, old habits die hard. I'll get there some day.

isobus/include/isobus/isobus/isobus_virtual_terminal_objects.hpp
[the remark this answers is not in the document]
GwnDaan: Ohhh I see! I looked a bit in the VT server, and noticed in almost all places where you call `get_object_by_id` you also have a `VTServerManagerWorkingSet` to your disposal. Maybe it would be an idea to move this function to there? Since that object should also have access to the object pool. 

Also theoretically, one might want to handle objects independent from the complete pool. I think that is a little difficult right now by requiring an object pool for an object. 

> If we could force people to interact with creating these objects only through some other class, then we could ensure the reference is valid... like, we could make the constructors badged, or private with a friend class of the object pool or something.

I think instead of limiting the interaction with the objects, the approach above, where specific operations can only happen with some kind of middle-man that holds the reference as value/smart_pointer instead, might be a good tradeoff?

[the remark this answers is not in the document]
GwnDaan: > If we could force people to interact with creating these objects only through some other class, then we could ensure the reference is valid

I mean, that would only hold if the other class holds the pool as a value **and** lives longer than the vt-object. The last one you can't really guarantee without an extra dependency I think

isobus/include/isobus/isobus/isobus_virtual_terminal_objects.hpp
ad3154: Hey there!

Not to get too off-topic on this PR, but here, take a look at [our parser](https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/2cbc71eee5beedb184be0ac4188fba515a4e14f0/isobus/src/isobus_virtual_terminal_server_managed_working_set.cpp#L145) as well for how we've been approaching it. It's maybe a bit heavy-weight for an ESP32 admittedly.

Some feedback I guess would be that it's mostly actually C! So, best-practices aside (such as having `using namespace` in a header file), I worry about how you're essentially just casting the pool data to the different types which would result in very strange things happening if the pool schema is not valid, possibly a hardfault/segfault. There might also be concerns between machines with different memory endianness? Though it's clearly working for ESP32 at least. Casting is efficient, when the pool is properly constructed yes, but intolerant of many OP issues which a server is obligated to report to a client I think unless I am missing something. `appendObject` seems like it could be the slowest part of parsing, lots of memcpy, malloc and free.

I think we are fairly close to making our VT GUI for PC available to the public, so be sure to keep a close eye on things! It has a few issues yet to work out, but I'd say it's maybe 80% complete. I've done some testing with 2 simultaneous working sets each with 10k+ objects and it seemed to hold up fairly well, I think.

isobus/include/isobus/isobus/isobus_virtual_terminal_objects.hpp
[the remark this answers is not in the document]
ad3154: Removed stored references to both the object tree and colour table.

isobus/include/isobus/isobus/isobus_virtual_terminal_objects.hpp
ad3154: Removed the function(s) mentioned, and refactored objects so that attributes of concern have dedicated setters and getters

GwnDaan (approved):

PR #353, opened 2023-11-22
[Core]: add control function references

GwnDaan:
Hmm, this basically should allow us to use weak_ptr's more easily throughout the stack for (Partnered)ControlFunction classes. Because of the way we assign `null_ptr`s to shared_ptr's when we want to indicate no specific control function, this also meant that you don't know if the weak_ptr expired or that it has the "no specific control function" assigned.

This is only needed if we still want to continue the journey of destroying a ControlFunction in one place, and then automatically stopping all functionality that depended upon that CF (I think we do want that? not sure tbh)

### DIFF
--- a/isobus/CMakeLists.txt
+++ b/isobus/CMakeLists.txt
@@ -39,7 +39,8 @@ set(ISOBUS_SRC
     "isobus_speed_distance_messages.cpp"
     "isobus_maintain_power_interface.cpp"
     "nmea2000_message_definitions.cpp"
-    "nmea2000_message_interface.cpp")
+    "nmea2000_message_interface.cpp"
+    "can_control_function_reference.cpp")
 
 # Prepend the source directory path to all the source files
 prepend(ISOBUS_SRC ${ISOBUS_SRC_DIR} ${ISOBUS_SRC})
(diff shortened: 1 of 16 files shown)


GwnDaan:
Curious on your thoughts @ad3154, I took a few shortcuts here and there to get it working such that I could get your feedback, such as implicit conversions

ad3154:
> I think we do want that?

Hmm, I'm not sure... I'm not really sure what to say I guess. I'm not sure how much people are creating and deleting partners during the lifetime of their application. I would think this would be a relatively rare thing to do, as if you want to dynamically lock onto and remove numerous partners at runtime, you'd probably not want to even use a partner for most of those cases, and instead just register for PGNs from everyone, and use the regular ControlFunction instead? That's how our VT server class works I guess. Perhaps making the user clean up all references to the shared pointer of a partner instead forces them to not have stale client/interface objects as well, encouraging them to delete their VT client and whatnot instead of letting them hang around with a dead ControlFunctionReference, IDK

GwnDaan:
Yeah well f*ck hahah. I agree, I jumped in too soon I think. I feel like many things are not absolutely perfect yet, but then when I try to change something, I guess it starts out really promising but in the end it's also not optimal 😦

PR #356, opened 2023-11-23
[CI/Meta]: Add documentation validation checking

ad3154:
- Added an action to validate doxygen builds with no warnings. 
- Slightly increased the warning level of doxygen to catch more issues. 
- Resolved all doxygen warnings.
- Updated doxyfile schema.


GwnDaan (approved):
LGTM!

Issue #357, opened 2023-11-25
CXX_STANDARD 14 Problems in Relation to esp-idf-v5.1.1


franz-ms-muc:
**Describe the bug**
CXX_STANDARD 14 Problems in Relation to esp-idf-v5.1.1

**Supporting Documentation**
https://github.com/Meisterschulen-am-Ostbahnhof-Munchen/AgIsoStack-plus-plus-sample/tree/cxx_20
--> does Compile fine. 

with the Setting 14 it brings following Error: 

```
In file included from /home/franz/esp-idf-v5.1.1/components/freertos/FreeRTOS-Kernel/portable/xtensa/include/freertos/portmacro.h:76,
                 from /home/franz/esp-idf-v5.1.1/components/freertos/FreeRTOS-Kernel/include/freertos/portable.h:59,
                 from /home/franz/esp-idf-v5.1.1/components/freertos/FreeRTOS-Kernel/include/freertos/FreeRTOS.h:71,
                 from /home/franz/esp-idf-v5.1.1/components/driver/twai/include/driver/twai.h:9,
                 from /home/franz/git/ms/AgIsoStack-plus-plus-sample/sample/components/AgIsoStack-plus-plus/AgIsoStack-plus-plus/hardware_integration/include/isobus/hardware_integration/twai_plugin.hpp:17,
                 from /home/franz/git/ms/AgIsoStack-plus-plus-sample/sample/components/AgIsoStack-plus-plus/AgIsoStack-plus-plus/hardware_integration/src/twai_plugin.cpp:10:
/home/franz/esp-idf-v5.1.1/components/esp_common/include/esp_macros.h:87:25: error: static assertion failed: CHOOSE_MACRO_VA_ARG() result does not match for 0 arguments
   87 | ESP_STATIC_ASSERT(foo() == 2, "CHOOSE_MACRO_VA_ARG() result does not match for 0 arguments");
      |                         ^
/home/franz/esp-idf-v5.1.1/components/esp_common/include/esp_macros.h:87:25: note: the comparison reduces to '(1 == 2)'
```

The reason might be: 

`-std=gnu++2b -fno-exceptions -fno-rtti -std=c++14`
The command line then contains 2 non-compatible Cxx Standards. 

**Environment (please complete the following information):**
 - OS: [esp-idf-v5.1.1]
 - Compiler [xtensa-esp32-elf-g++ (crosstool-NG esp-12.2.0_20230208) 12.2.0]
[...]

PR #358, opened 2023-11-25
Update CMakeLists.txt to use C++ standard 20 for unit_tests

franz-ms-muc:
- Update the CXX_STANDARD property from 14 to 20 in the unit_tests target of CMakeLists.txt.
- fix https://github.com/Open-Agriculture/AgIsoStack-plus-plus/issues/357

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,7 +119,7 @@ if(BUILD_TESTING)
   add_executable(unit_tests ${TEST_SRC})
   set_target_properties(
     unit_tests
-    PROPERTIES CXX_STANDARD 14
+    PROPERTIES CXX_STANDARD 20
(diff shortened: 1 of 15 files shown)


GwnDaan:
Hi, thanks for taking your time for this contribution! Unfortunately as stated in our [CONTRIBUTING.md](https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/main/CONTRIBUTING.md) guideline; 
> Only features present in C++14 can be used to help provide the maximum compatibility to all "modern" compilers

by enforcing C++14 we can support a wider range of compilers (even the older, not updated to the latest standard ones). Hence updating the `CXX_STANDARD` to anything above 14 will violate that guideline.

Let's look for another way to solve #357

PR #363, opened 2023-11-30
[TC]: Add restart and reupload capabilities to TC client

### LINKED ISSUE #354: Restart routine for TC
> What I've tried thus far is calling the `terminate()`, then `configure()` and  `initialize()` again. However the doesn't work as the TC client doesn't enter the "Disconnected" state, and I get an error from the stack indicating it cannot be reconfigured. 
I noticed in the VT client `terminate()` call  is explicitly calls the `set_state(Disconnected)` so I tried adding that to the TC client, but that seems to break other things. 

_Originally posted by @trm01 in https://github.com/Open-Agriculture/AgIsoStack-plus-plus/discussions/313#discussioncomment-7614770_

### LINKED ISSUE #355: Add a way to request re-upload of DDOP
> Changing DPTs' values at runtime or adding/removing objects in the DDOP will always require completely re-uploading the DDOP, which we could and should allow I suppose, without deleting the interface. It should be fairly simple to add this functionality.

_Originally posted by @ad3154 in https://github.com/Open-Agriculture/AgIsoStack-plus-plus/discussions/313#discussioncomment-7644164_

ad3154:
## What's new?

- Added a way to reset a connection with a task controller server manually. 
- Added a way to trigger an update (and reupload) of a DDOP. 
- Fixed a crash/exception when trying to call the TC client terminate function from within the worker thread.

Closes #354 

Closes #355

### DIFF
--- a/isobus/include/isobus/isobus/isobus_task_controller_client.hpp
+++ b/isobus/include/isobus/isobus/isobus_task_controller_client.hpp
@@ -211,6 +211,10 @@ namespace isobus
 		               bool reportToTCSupportsPeerControlAssignment,
 		               bool reportToTCSupportsImplementSectionControl);
 
+		/// @brief Calling this function will reset the task controller client's connection
+		/// with the TC server, and cause it to reconnect after a short delay.
+		void restart();
+
 		// Calling this will stop the worker thread if it exists
 		/// @brief Terminates the client and joins the worker thread if applicable
 		void terminate();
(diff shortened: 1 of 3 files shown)


ad3154:
Formatting check depends on #364, and I'll try and throw in a couple trivial unit tests but it's still essentially ready for review.

trm01:
I just tried this branch with 2 different TCs and saw the same behavior.  The re-upload works well, much quicker than the tear-down and re-initialization. Great so far, however after the re-upload I start to see that the TC is requesting values for what I think is the incorrect elements. My 2nd DDOP re-uses element numbers (maybe that is not allowed? ) and in my  `RequestValueCommandCallback`  is getting DDIs requests for  a element number that never presented it as an option in the 2nd DDOP.  

Am I doing something wrong, or are these element numbers reset within the stack or TC server?

ad3154:
Oh hey, good catch, it's probably because we call that function based on the content of previous requests from the TC, and I'm probably not clearing them on a reset - I will push an update to this branch that will probably fix that

trm01:
Awesome, thanks! 
I won't be able to retest tonight, but will retest tomorrow and get back to you.

trm01:
FYI:  I gave this a try and I am still seeing the callbacks. I'm currently digging into the code right now to ensure it isn't something that I am messing up with  my DDOP changes.

trm01:
Okay, we're good. 
I'm confirming it was something I am doing in my code... I was calling the re-upload twice in a row very quickly with 2 different DDOPs. The second one was not being uploaded as the TC client state machine had already started the 1st  deactivation.

GwnDaan (approved):

PR #366, opened 2023-11-30
[VT]: Add VT objects for Aux inputs and Aux functions

ad3154:
- Added VT objects for Aux Input type 1 and 2.
- Added VT objects for Aux Functions type 1 and 2.
- Added Auxiliary Control Designator Type 2 Object

### DIFF
--- a/test/vt_object_tests.cpp
+++ b/test/vt_object_tests.cpp
@@ -3063,6 +3063,32 @@ TEST(VIRTUAL_TERMINAL_OBJECT_TESTS, AuxiliaryInputType2Tests)
 	EXPECT_TRUE(auxiliaryInput->get_function_attribute(AuxiliaryInputType2::FunctionAttribute::AssignmentRestriction));
 	EXPECT_FALSE(auxiliaryInput->get_function_attribute(AuxiliaryInputType2::FunctionAttribute::CriticalControl));
 	EXPECT_EQ(AuxiliaryFunctionType2::FunctionType::AnalougeReturnTo50Percent, auxiliaryInput->get_function_type());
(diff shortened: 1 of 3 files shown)


test/vt_object_tests.cpp
```diff
@@ -2975,3 +2975,265 @@ TEST(VIRTUAL_TERMINAL_OBJECT_TESTS, ObjectPointerTests)
...
+	run_baseline_tests(auxiliaryInput.get());
+	EXPECT_EQ(auxiliaryInput->get_object_type(), VirtualTerminalObjectType::AuxiliaryInputType1);
+
+	auxiliaryInput->set_function_type(AuxiliaryInputType1::FunctionType::Analogue);
+	EXPECT_EQ(AuxiliaryInputType1::FunctionType::Analogue, auxiliaryInput->get_function_type());
+
+	auxiliaryInput->set_input_id(200);
+	EXPECT_EQ(200, auxiliaryInput->get_input_id());
+
+	auxiliaryInput->set_input_id(254); // Max is 250
```
GwnDaan, line 2993:
```suggestion
	ASSERT_FALSE(auxiliaryInput->set_input_id(254)); // Max is 250
```
I think returning true/false would be nice, the docs currently doesn't state it caps at 250, eventhough it's in the standard it's nice to have here as well to avoid confusion

GwnDaan (changes requested):

test/vt_object_tests.cpp
ad3154: Added! 🙂

GwnDaan (approved):

PR #368, opened 2023-12-01
[DDOP]: Add an ISO XML DDOP serializer

ad3154:
## What's New:

- Added a way to export our DDOP class contents as ISO XML (TASKDATA.xml)

This is a useful feature for task controller servers.


GwnDaan (approved):
No experience with ISO XML files, but it looks well defined 😄

ad3154:
> No experience with ISO XML files, but it looks well defined 😄

It's very strange and magical, almost as if they made the fields unreadable nonsense on purpose so that you have to purchase a copy of the standard to interoperate with it, haha.

PR #373, opened 2023-12-04
Removed references to std::make_unique

ad3154:
This is an attempt to expand compatibility with the consumers of the Arduino library, one of whom reported that `std::make_unique` was causing them to be unable to compile the library.

https://github.com/Open-Agriculture/AgIsoStack-Arduino/issues/7

Interested to hear thoughts on this... I know we are targeting at least C++14 (which is when make_unique was introduced) but this seems like a fairly minor change to bring the Arduino lib back down to C++11

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,7 +120,7 @@ if(BUILD_TESTING)
   add_executable(unit_tests ${TEST_SRC})
   set_target_properties(
     unit_tests
-    PROPERTIES CXX_STANDARD 14
+    PROPERTIES CXX_STANDARD 11
                CXX_EXTENSIONS OFF
                CXX_STANDARD_REQUIRED ON)
   target_link_libraries(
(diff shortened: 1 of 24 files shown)


GwnDaan:
While it hurts me to downgrade the supported C++ version to 11, I feel like it's the only reasonable option here. The arduino-ide is widely used among hobbyists and I don't want them to have to follow some cumbersome procedure to be able to compile the stack. I had a shot at it myself, and managed to succeed in changing the version, but only after following a dozen links on google 😒. 

Though I'm happy it seems like a pretty quick fix to go back to C++11 with the stack. Can you update the CMakeLists.txt as well, so Github Actions can notify us if we happen to make changes not compliant with C++11?

ad3154:
> > Can you update the CMakeLists.txt as well
> 
> Will do

I know I said will do, but then you did it in #382 😆

Sonar seems mad about C++ 17 things like `std::scoped_lock` and template type deduction so I'll see if I can modify those rules...

GwnDaan (approved):

Issue #374, opened 2023-12-05
Update supported version in working set maintenance message


GwnDaan:
Currently we ask the application to provide the "supported" VT version for a pool. This is **not** what the standard want us to provide. It instead requires us to set the VT version in the maintenance message to the version of the standard that the application is designed to. 

I think the correct thing to do is provide the standard version the application is designed to comply with. Since the stack is basically the one doing communication defined in the standard, that means that we can put it to a constant version. Namely we aim to be complaint with VT version 6, i.e. ISO11783-6:2018.

Reference to applicable code:
https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/495195584f1a8a30e9474d7aa8ded6657dcddc3e/isobus/src/isobus_virtual_terminal_client.cpp#L1767-L1819

ad3154:
We will have to monitor what happens when we do this. Although I agree with it, I _have_ seen some VT servers do some pretty funky stuff, and we should recall that clients are required to change their behavior based on the server version, so it may be less obvious to consumers now that version is something they need to be thinking about. We should be sure that our logging statements are high quality for situations where the server version < client.

ad3154:
> if it follows the standard

I'm not going to name any names... but I have seen some... things in my days... It is not reasonable to try to cover non-conformant terminals probably.

>we should be fully backwards compatible 

The main thing we cannot control is what the user puts in the object pool, and what functions they call, both of which may result in non-conformant behavior, so we may need to add some additional validation as we `implement the standard more thoroughly in the coming months` to minimize this possibility?

ad3154:
> It is not reasonable to try to cover non-conformant terminals probably

(Except AgIsoVirtualTerminal hehe)

ad3154:
I think it's OK to make the change now, but we'll then want to prioritize follow-up refactoring in the client for anything we're not conformant on, such as TAN processing for version 6 servers

GwnDaan:
Yeah sounds good, let's make some issues to try to upgrade our compatibility

GwnDaan:
> and what functions they call

We can limit the outgoing messages based on the VT version connected right? E.g. we can make functions like `send_lock_unlock_mask` (Introduced in VT version 4)  always return `false` if used when connected to a VT version 3 or prior.

GwnDaan:
> The main thing we cannot control is what the user puts in the object pool
[...]

PR #375, opened 2023-12-05
changes standard version to VT version 6

davidfiliks:
should resolve issue #374 if I understood correctly.

### DIFF
--- a/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
+++ b/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
@@ -1338,7 +1338,6 @@ namespace isobus
 			std::uint32_t objectPoolSize; ///< The size of the object pool
 			std::uint32_t autoScaleDataMaskOriginalDimension; ///< The original length or width of this object pool's data mask area (in pixels)
 			std::uint32_t autoScaleSoftKeyDesignatorOriginalHeight; ///< The original height of a soft key designator as designed in the pool (in pixels)
-			VTVersion version; ///< The version of the object pool. Must be the same for all pools!
 			bool useDataCallback; ///< Determines if the client will use callbacks to get the data in chunks.
 			bool uploaded; ///< The upload state of this pool
 		};
--- a/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
+++ b/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
@@ -1370,6 +1370,8 @@ namespace isobus
 		/// @returns true if the message was sent
 		bool send_delete_object_pool() const;
 
+		bool send_working_set_maintenance(bool initializing) const;
+
 		/// @brief Sends the working set maintenance message
 		/// @param[in] initializing Used to set the initializing bit
 		/// @param[in] workingSetVersion The version supported by the working set
--- a/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
+++ b/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
@@ -1370,13 +1370,10 @@ namespace isobus
 		/// @returns true if the message was sent
 		bool send_delete_object_pool() const;
 
-		bool send_working_set_maintenance(bool initializing) const;
-
 		/// @brief Sends the working set maintenance message
 		/// @param[in] initializing Used to set the initializing bit
-		/// @param[in] workingSetVersion The version supported by the working set
 		/// @returns true if the message was sent
-		bool send_working_set_maintenance(bool initializing, VTVersion workingSetVersion) const;
+		bool send_working_set_maintenance(bool initializing) const;
 
 		/// @brief Sends the get memory message
 		/// @details This message checks to see if the VT has enough memory available to store your object pool(s)
--- a/isobus/src/isobus_virtual_terminal_client.cpp
+++ b/isobus/src/isobus_virtual_terminal_client.cpp
@@ -1376,7 +1376,7 @@ namespace isobus
 					if (0 != objectPools.size())
 					{
 						set_state(StateMachineState::SendGetMemory);
-						send_working_set_maintenance(true, objectPools[0].version);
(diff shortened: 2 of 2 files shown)


isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
```diff
@@ -1370,6 +1370,8 @@ namespace isobus
 		/// @returns true if the message was sent
 		bool send_delete_object_pool() const;
 
+		bool send_working_set_maintenance(bool initializing) const;
```
GwnDaan, line 1373: issue: there are now two send_working_set_maintenance functions defined in the header file. 

Try to modify the old one instead of adding a new one

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -1764,47 +1764,14 @@ namespace isobus
...
-			{
-				versionByte = 0xFF;
-			}
-			break;
-		}
-
 		const std::array<std::uint8_t, CAN_DATA_LENGTH> buffer = { static_cast<std::uint8_t>(Function::WorkingSetMaintenanceMessage),
 			                                                         bitmask,
-			                                                         versionByte,
+			                                                         0x06,
```
GwnDaan, line 1774: Nitpick(non-blocking): instead of specifying it directly inside the function, try to define it using a constexpr before so it is clear what it is for later on. 

E.g. "static constexpr std::uint8_t SUPPORTED_VT_VERSION = 0x06"

GwnDaan (changes requested):
Thanks for your PR, really appreciate it! I have a few comments, could you look at those?

GwnDaan:
Also the version passed into the function is saved as a class member of the VirtualTerminalClient class, it'll be unused after these changes. So could you remove it aswell?

isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
davidfiliks: Should I just remove the one which takes VTVersion workingSetVersion as a parameter? I thought there would be no need for this one since you want the version to be 6 all the time.

isobus/src/isobus_virtual_terminal_client.cpp
davidfiliks: Will do! Thanks.

davidfiliks:
I'll try to do all these, thank you!

davidfiliks:
I have made another commit, check it out and let me know any feedback. Thanks.

isobus/src/isobus_virtual_terminal_client.cpp
GwnDaan: Almost there, only instead of putting it inside the global namespace, I recommend putting it inside the function itself, since that would be the only place this constant will be used. We try to minimize the global namespace usage

davidfiliks: Alright, will do that and submit a commit shortly.

GwnDaan:
The build still seems to fail because where this function is used, the parameter "version" is still passed in: https://github.com/davidfiliks/AgIsoStack-plus-plus/blob/9f5e841d470eb57f8713a1e24accf06ecd5227f2/isobus/src/isobus_virtual_terminal_client.cpp#L1381

Furthermore, I was referring to this class member being unused after we make the changes in this PR: https://github.com/davidfiliks/AgIsoStack-plus-plus/blob/9f5e841d470eb57f8713a1e24accf06ecd5227f2/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp#L1341
Would you be able to remove it, together with all places it is accessed/modified? That should only be inside the three `set_object_pool` functions.

I enabled the Github Actions checks to automatically start, try to get all checks to succeed. Again thanks for taking your time to make a contribution, highly appreciated :)

davidfiliks:
Thank you for helping like this, it is my first ever contribution on github. Check it out and give any feedback. Thanks.

ad3154:
> Thank you for helping like this, it is my first ever contribution on github. Check it out and give any feedback. Thanks.

Well, it's cool that your first contribution is here with us! Welcome! 

It's looking pretty good to me, my only comment would be that it would be nice to squash or fixup all your commits into one commit with a nice tile, something like your first one "Change VT Client version to VT version 6" or something similar.

GwnDaan:
> It's looking pretty good to me, my only comment would be that it would be nice to squash or fixup all your commits into one commit with a nice tile, something like your first one "Change VT Client version to VT version 6" or something similar.

@ad3154 FYI, that's also something GitHub can do for us :). It's the `Squash and merge` option

davidfiliks:
How would I merge all these commits into one?

ad3154:
> How would I merge all these commits into one?

It depends on how you interact with Git. I use [fork](https://git-fork.com/) but for raw Git, it would be something like what is [described here](https://www.freecodecamp.org/news/git-squash-commits/) with a `git rebase -i HEAD~5` followed by force pushing the result to your branch.

This is a very common thing in Git, and fairly critical part of most git workflows, so it's something to look into as you gain experience with it, but no worries if it's not something you want to deal with right now, we can just have GitHub auto-squash it.

davidfiliks:
> _fatal: It seems that there is already a rebase-merge directory, and
> I wonder if you are in the middle of another rebase.  If that is the
> case, please try
>         git rebase (--continue | --abort | --skip)
> If that is not the case, please
>         rm -fr ".git/rebase-merge"
> and run me again.  I am stopping in case you still have something
> valuable there._

I am not sure what to do here. This is the output after git rebase -i HEAD~5

ad3154:
> I am not sure what to do here. This is the output after git rebase -i HEAD~5

This seems like the rebase exited while it was happening, like maybe it couldn't open your selected default text editor or something, or there was some conflict... Don't worry about it for now, you can cancel it with `git rebase --abort`. We can just squash it on here.

ad3154 (approved):

davidfiliks:
Alright, thanks.

Issue #378, opened 2023-12-06
Update all VT client messages to priority 5


ad3154:
Once we merge in #375 we'll need to adjust the priorities of all VT client messages.

ISO11783-6 Version 5 and prior required this to be 7, but VT version 6 suggests priority 5 to follow recommendations provided in ISO 11783-3.

PR #379, opened 2023-12-06
update all VT client messages to priority 5 done

davidfiliks:
updates all VT client messages to priority 5 as stated in #378

### DIFF
--- a/isobus/src/isobus_virtual_terminal_client.cpp
+++ b/isobus/src/isobus_virtual_terminal_client.cpp
@@ -1237,7 +1237,6 @@ namespace isobus
 			tempData.objectPoolSize = size;
 			tempData.autoScaleDataMaskOriginalDimension = 0;
 			tempData.autoScaleSoftKeyDesignatorOriginalHeight = 0;
-			tempData.version = poolSupportedVTVersion;
 			tempData.useDataCallback = false;
 			tempData.uploaded = false;
 			tempData.versionLabel = version;
@@ -1267,7 +1266,6 @@ namespace isobus
 			tempData.objectPoolSize = pool->size();
 			tempData.autoScaleDataMaskOriginalDimension = 0;
 			tempData.autoScaleSoftKeyDesignatorOriginalHeight = 0;
-			tempData.version = poolSupportedVTVersion;
 			tempData.useDataCallback = false;
 			tempData.uploaded = false;
 			tempData.versionLabel = version;
@@ -1305,7 +1303,6 @@ namespace isobus
 			tempData.objectPoolVectorPointer = nullptr;
 			tempData.dataCallback = value;
 			tempData.objectPoolSize = poolTotalSize;
-			tempData.version = poolSupportedVTVersion;
 			tempData.useDataCallback = true;
 			tempData.uploaded = false;
 			tempData.autoScaleSoftKeyDesignatorOriginalHeight = 0;
@@ -1376,7 +1373,7 @@ namespace isobus
 					if (0 != objectPools.size())
 					{
 						set_state(StateMachineState::SendGetMemory);
-						send_working_set_maintenance(true, objectPools[0].version);
+						send_working_set_maintenance(true);
 						lastWorkingSetMaintenanceTimestamp_ms = SystemTiming::get_timestamp_ms();
 						sendWorkingSetMaintenance = true;
 						sendAuxiliaryMaintenance = true;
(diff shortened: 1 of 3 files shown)


ad3154:
> I don't know why it's failing...

It's failing because the unit tests for the VT client are comparing the whole CAN ID of the expected message to the actual message that gets sent.

image

Here you can see that it's checking to see that the ID of this message begins with 0x1C, which is priority 7, but now we're sending it with a leading 0x14, which is priority 5. So, in order to get it to pass, you'd need to update the file `test\vt_client_tests.cpp` to change the expected ID of the failing messages.

ad3154:
You can check the output [here](https://github.com/Open-Agriculture/AgIsoStack-plus-plus/actions/runs/7118012733/job/19379997766#step:4:321) to see the exact lines you'll have to modify: 919, 936, 952, 968, 979, and 992.

davidfiliks:
Thanks

test/vt_client_tests.cpp
```diff
@@ -965,7 +965,7 @@ TEST(VIRTUAL_TERMINAL_TESTS, MessageConstruction)
 	EXPECT_EQ(0, testFrame.channel);
 	EXPECT_EQ(CAN_DATA_LENGTH, testFrame.dataLength);
 	EXPECT_TRUE(testFrame.isExtendedFrame);
-	EXPECT_EQ(0x1CE72637, testFrame.identifier);
+	EXPECT_EQ(0x1CE72637, 0x1CE72637);
```
ad3154, line 968: Not quite what I meant... now we're just comparing two hard-coded numbers.

Here's what I meant:
```suggestion
	EXPECT_EQ(0x14E72637, testFrame.identifier);
```

davidfiliks: Sorry, I understood it wrong. I thought that 0x1C was the code for priority 5. My bad.

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -4335,12 +4301,12 @@ namespace isobus
 		                                                      data.size(),
 		                                                      myControlFunction,
 		                                                      partnerControlFunction,
-		                                                      priority);
+		                                                      CANIdentifier::Priority5);
 	}
 
 	bool VirtualTerminalClient::queue_command(const std::vector<std::uint8_t> &data, CANIdentifier::CANPriority priority, bool replace)
```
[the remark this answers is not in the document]
GwnDaan: Also further down the line it can be removed, so that would be inside the `replace_command` function and `commandQueue` field

davidfiliks:
Sorry, I'll do this when I come back from school.

davidfiliks:
Please check if this is what you wanted.

GwnDaan:
Good job! It'a almost there I think, only thing still left would be the line:

https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/main/isobus%2Finclude%2Fisobus%2Fisobus%2Fisobus_virtual_terminal_client.hpp#L1657

Here again priority is not used anymore, so the `std::pair` can be removed, and you'd end up with the nested `std::vector<std::vector<std::uint8_t>>`

Also I saw you managed to squash your commits, nice!

davidfiliks:
Hey, I don't know why you would want to remove this, because we want the priority to still exist but it just to be set to 5? Anyways, I did this and the it->first and it->second statements stopped working. I tried to fix this but its most likely wrong. Check it out. Thanks!

GwnDaan:
> Hey, I don't know why you would want to remove this, because we want the priority to still exist but it just to be set to 5? Anyways, I did this and the it->first and it->second statements stopped working. I tried to fix this but its most likely wrong. Check it out. Thanks!

Yeah, so basically the  `commandQueue` was the link between the `queue_command` function and `send_command` function. When that link first was created, I thought it could have different priorities, but with the current standard there we're no different priorities for VT commands. 

However, that doesn't mean that messages (something different than the VT commands) can't have other priorities, it's just that the `commandQueue` is only for VT commands and not for messages in general. And since you already changes the `send_command` to have a fixed priority of 5, I figured we can just remove the priority for the VT commands further up the road, since after that fixed priority, they we're basically unused.

I hope this explanation makes sense to you hahah. I'm not sure about your knowledge of the CAN/J1939/ISO11783 protocols in general, but I understand it could be a bit hard to gasp otherwise.

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -4350,26 +4316,26 @@ namespace isobus
...
-	bool VirtualTerminalClient::replace_command(const std::vector<std::uint8_t> &data, CANIdentifier::CANPriority priority)
+	bool VirtualTerminalClient::replace_command(const std::vector<std::uint8_t> &data)
 	{
 		bool alreadyReplaced = false;
 		for (auto it = commandQueue.begin(); it != commandQueue.end();)
 		{
-			bool matchesFunctionCode = (it->first[0] == data[0]);
-			if (matchesFunctionCode && (it->second == priority))
+			bool matchesFunctionCode = (it[0][0] == data[0]);
+			if (matchesFunctionCode)
```
GwnDaan, line 4334:
```suggestion
			bool matchesFunctionCode = (it[0] == data[0]);
			if (matchesFunctionCode)
```
I think this is better? Not sure if it makes a difference

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -4350,26 +4316,26 @@ namespace isobus
...
 			{
 				if (!alreadyReplaced)
 				{
-					it->first = data;
+					it[0] = data;
```
GwnDaan, line 4338:
```suggestion
					*it = data;
```
Same here, not sure if there's a difference. But I feel like the `*` notation makes it easier to gasp that we're still talking about a complete vector of data here

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -4386,7 +4352,7 @@ namespace isobus
 	{
 		for (auto it = commandQueue.begin(); it != commandQueue.end();)
 		{
-			if (send_command(it->first, it->second))
+			if (send_command(it[0]))
```
GwnDaan, line 4355:
```suggestion
			if (send_command(*it))
```
Same here

isobus/src/isobus_virtual_terminal_client.cpp
GwnDaan: Or maybe *it[0]?

davidfiliks: Nah, this doesn't work. I've tried and the only thing that seems to work is it[0][0] since it is a vector of a vector and data is just a vector.

GwnDaan: Oh yeah, I can totally see why that's confusing. Let me explain; basically a `std::vector<X>::iterator` means it a iterator/pointer to an element of it's suffix. So essentially the whole `std::vector<X>::iterator` points to an element X (which lies inside the vector), and hence in our case, it point to an inner `std::vector<uint8_t>` element. Does that help?

GwnDaan:
Thanks again for taking the time to contribute to the project. Don't take the feedback personally, I like to keep the project up to a high standard haha. Your efforts are greatly appreciated!

PR #380, opened 2023-12-07
[VT] Wait on response before processing next command

GwnDaan:
To comply with the standard, we must wait for responses from the VT server before sending the next command. This PR implements that behaviour

### DIFF
--- a/isobus/src/isobus_virtual_terminal_client.cpp
+++ b/isobus/src/isobus_virtual_terminal_client.cpp
@@ -250,11 +250,6 @@ namespace isobus
 		}
 	}
 
-	void VirtualTerminalClient::set_should_queue_commands(const bool shouldQueueCommands)
-	{
-		commandQueueEnabled = shouldQueueCommands;
-	}
-
 	isobus::VirtualTerminalClient::AssignedAuxiliaryFunction::AssignedAuxiliaryFunction(const std::uint16_t functionObjectID, const std::uint16_t inputObjectID, const AuxiliaryTypeTwoFunctionType functionType) :
 	  functionObjectID(functionObjectID), inputObjectID(inputObjectID), functionType(functionType)
 	{
@@ -1765,7 +1760,6 @@ namespace isobus
 	{
 		static constexpr std::uint8_t SUPPORTED_VT_VERSION = 0x06;
 
-		std::uint8_t versionByte;
 		std::uint8_t bitmask = (initializing ? 0x01 : 0x00);
 
 		const std::array<std::uint8_t, CAN_DATA_LENGTH> buffer = { static_cast<std::uint8_t>(Function::WorkingSetMaintenanceMessage),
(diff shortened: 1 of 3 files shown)


isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -4322,20 +4360,30 @@ namespace isobus
 		return retVal;
 	}
 
-	bool VirtualTerminalClient::send_command(const std::vector<std::uint8_t> &data, CANIdentifier::CANPriority priority) const
+	bool VirtualTerminalClient::send_command(const std::vector<std::uint8_t> &data, CANIdentifier::CANPriority priority)
 	{
+		if (commandAwaitingResponse)
```
ad3154, line 4365: We should probably add a timeout and check that as well in here? If we're waiting more than 1.5s (I think that's what the standard said?) then we could process the transmit as normal and log what happened

GwnDaan: Oh yeah of course, totally forgot it

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -4288,20 +4325,31 @@ namespace isobus
 		return retVal;
 	}
 
-	bool VirtualTerminalClient::send_command(const std::vector<std::uint8_t> &data) const
+	bool VirtualTerminalClient::send_command(const std::vector<std::uint8_t> &data)
 	{
+		if (commandAwaitingResponse && (!SystemTiming::time_expired_ms(lastCommandTimestamp_ms, 1500)))
+		{
+			return false;
+		}
```
ad3154, line 4333: I'd like suggest maybe something like this?

```suggestion
		bool isTimedOut = SystemTiming::time_expired_ms(lastCommandTimestamp_ms, 1500);
		
		if (commandAwaitingResponse && (!isTimedOut ))
		{
			return false;
		}
		else if ((0 != lastCommandTimestamp_ms) && isTimedOut)
		{
			CANStackLogger::warn("[VT]: Server response to a command timed out");
		}
```

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -3033,6 +3027,49 @@ namespace isobus
...
+						case static_cast<std::uint8_t>(Function::SelectColourMapCommand):
+						case static_cast<std::uint8_t>(Function::ExecuteExtendedMacroCommand):
+						case static_cast<std::uint8_t>(Function::SelectActiveWorkingSet):
+						{
+							// By checking if it's a response with our control functions, we verify that it's a response to a request we sent.
+							// This because we only support Working Set Masters at the moment.
+							if ((parentVT->myControlFunction == message.get_destination_control_function()) &&
+							    (parentVT->partnerControlFunction == message.get_source_control_function()))
+							{
+								parentVT->commandAwaitingResponse = false;
```
ad3154, line 3069: If you take my other suggestion we may also want to reset the timestamp to 0 here

isobus/src/isobus_virtual_terminal_client.cpp
GwnDaan: Logging is indeed desired here, I changed up your suggestion a little bit. I feel like it's easier to read:
```c++
if (commandAwaitingResponse)
{
	if (SystemTiming::time_expired_ms(lastCommandTimestamp_ms, 1500))
	{
		CANStackLogger::warn("[VT]: Server response to a command timed out");
		commandAwaitingResponse = false;
	}
	else
	{
		// We're still waiting for a response to the last command, so we can't send another one yet
		return false;
	}
}
```

isobus/src/isobus_virtual_terminal_client.cpp
GwnDaan: With my edited solution we don't need to set timestamp to 0, I think it's valid to always leave it at the timestamp the last command was sent.

GwnDaan:
The main branch is currently failing to build on PlatformIO: https://github.com/Open-Agriculture/AgIsoStack-plus-plus/actions/runs/7147251833/job/19466433156. This PR will also fix that

ad3154:
Tested this by re-basing the seeder example on your branch, and... it's a little slow, only around 20 messages per second. Our update rate of the interface means that after a response is received we have to wait a fair bit before sending the next message which causes some noticeable lag. I am wondering if we should explore options to send the next message on-response.

Some ideas:

- If using the worker thread, mutex the queue and initiate the transmit of the next message from the receiving thread
- Or, Use a condition variable to wake up the VT worker thread when a message response is received?
- Update faster?

GwnDaan:
I opted for the mutex because I feel like it's the most efficient one, good that you tested this. I didn't notice it during testing

isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -4350,6 +4406,9 @@ namespace isobus
 
 	void VirtualTerminalClient::process_command_queue()
 	{
+#if !defined CAN_STACK_DISABLE_THREADS && !defined ARDUINO
+		std::lock_guard<std::mutex> lock(commandQueueMutex);
+#endif
 		for (auto it = commandQueue.begin(); it != commandQueue.end();)
 		{
 			if (send_command(*it))
```
ad3154, line 4414: Rather than returning false on a bunch of the queue, we could maybe change this if-else to be 

```
if (!send_command(*it))
{
    break;
}
```
Otherwise we're wasting time with the mutex locked I think

GwnDaan: Hmm, I'm not sure. My thought of not breaking out of the loop is mainly that there can be commands 'single-framed' and others using transport protocols. Hence some might still succeed if others fail to send. Also since the mutex is solely around the CommandQueue, and in other places not used very often, I feel like the mutex being locked won't be really be a problem?

ad3154: Ah OK, I see, that makes sense

ad3154 (approved):

PR #387, opened 2023-12-12
Remove implicit conversion for identifier priority

GwnDaan:
With this change, **all** compilers will throw errors when conversion is done implicitly on the identitier enums. It prevents us from making subtle mistakes where conversion shouldn't have happened.

### DIFF
--- a/isobus/src/isobus_virtual_terminal_client.cpp
+++ b/isobus/src/isobus_virtual_terminal_client.cpp
@@ -275,7 +275,7 @@ namespace isobus
 			                                         0xFF,
 			                                         0xFF,
 			                                         0xFF };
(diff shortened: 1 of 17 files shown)


isobus/src/isobus_virtual_terminal_client.cpp
```diff
@@ -275,7 +275,7 @@ namespace isobus
 			                                         0xFF,
 			                                         0xFF,
 			                                         0xFF };
-		return queue_command(buffer, CANIdentifier::CANPriority::PriorityLowest7);
+		return queue_command(buffer);
```
GwnDaan, line 278: This is one of those implicit conversions that is simply incorrect. The second parameter is a `boolean`, while the priority enum represents `uint8_t`

ad3154 (approved):

PR #396, opened 2023-12-22
Less duplication for macro execute/process commands

GwnDaan:
Also renamed made `processMacro` to `process_macro` to follow snake_case convention of repository

### DIFF
--- a/isobus/src/isobus_virtual_terminal_server.cpp
+++ b/isobus/src/isobus_virtual_terminal_server.cpp
@@ -171,11 +171,11 @@ namespace isobus
 				CANStackLogger::debug("[VT Server]: Executing macro %u", macro->get_id());
 				retVal = true;
 
-				for (std::uint16_t j = 0; j < macro->get_number_of_commands(); j++)
+				for (std::uint8_t j = 0; j < macro->get_number_of_commands(); j++)
 				{
 					std::array<std::uint8_t, 8> commandPacket = { 0 };
 
-					if (macro->get_command_packet(static_cast<std::uint8_t>(j), commandPacket))
+					if (macro->get_command_packet(j, commandPacket))
 					{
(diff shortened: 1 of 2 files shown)


isobus/src/isobus_virtual_terminal_server.cpp
```diff
@@ -171,11 +171,11 @@ namespace isobus
 				CANStackLogger::debug("[VT Server]: Executing macro %u", macro->get_id());
 				retVal = true;
 
-				for (std::uint16_t j = 0; j < macro->get_number_of_commands(); j++)
```
ad3154, line 174: I actually had this as uint16 to avoid an infinite loop in the case where exactly 255 macros were included

GwnDaan: Hmm, that's strange. The `get_number_of_macros` function returns an `uint8`, so a loop of `uint16` doesnt make sense to me as a fix for that issue

ad3154: 255 commands, rather. Like if get_number_of_commands returns 255, our counter here would roll over

GwnDaan: Sorry I meant get_number_of_commands indeed. That is still a uint8_t, so max would be 255 commands as you said. 

Here we are speaking about a **strictly less** loop, which would mean the loop stops at `i=254` since `(i=255) < (255=max)` doesn't hold (which is also the index of the command 255). You can see the following example function in assembly, if that helps:
image

If this was about a **less or equal** operator or something, then it would be another story; 
image
here the loop can never reach it's end condition since no uint8_t can be higher than 255

ad3154: Ah ok

ad3154 (approved):

PR #400, opened 2023-12-29
Add VT Server

ad3154:
This is a draft for cleaning up and eventually pulling in the VT server files.

- Created a base class for VT client & server so they can share common enums
- Added the concept of a "managed working set" which stores client specific data for a VT server
    - This also provides the object pool deserializer. Deserialization is handled in a worker thread to avoid tying up a server while it's parsing an object pool.
- Added VT Server (somewhat WIP, needs lots of Aux-N stuff for example)
    - The actual server interface is abstract, though I provide an implementation for some app specific things like `get_supported_small_fonts_bitfield`... probably need to re-evaluate some abstraction before merging.
- Fixed a number of warnings in various places, which I got when building with JUCE


ad3154:
Now that I can work on the VT properly, I think it makes sense to squash all this and merge it in, and then make incremental reviews as we iterate on the VT.

You okay with that @GwnDaan

GwnDaan (approved):
Hi, yes, that sounds good to me. Sorry for being inactive lately, busy times. I hope to have a bit of spare time this winter to pick up some tasks because I genuinely enjoyed working on the stack

PR #401, opened 2024-01-06
Bootstrap public api documentation in sphinx

GwnDaan:
To enhance the user experience for finding the right parts of the public api of the stack, I think adding them to the sphinx site is a good idea. Note that this doesn't replace the doxygen documentation, instead it provides a more structured alternative for less experienced users.

It is automatically build as well: https://agisostack-plus-plus.readthedocs.io/en/latest/api/index.html. 

Also I claimed the `agisostack-plus-plus` subdomain, so we probably want the `isobus-plus-plus` subdomain to redirect to it?


ad3154:
It seems like redirecting the whole thing is not really supported without some hacky workarounds... I wonder if we merge it if it'll build identical docs for both RTD projects?

GwnDaan:
> I wonder if we merge it if it'll build identical docs for both RTD projects?

Hmm, yeah that could work I think. Though I feel like that stresses RTD servers for no good reason hahah.

I found [this](https://stackoverflow.com/a/69928404) stackoverflow post which we can follow instead; https://github.com/GwnDaan/AgIsoStackpp-docs-redirect. I setup a test redirect https://agisostackplusplus.readthedocs.io and it seems to work fine

ad3154 (approved):

PR #404, opened 2024-01-20
[VT]: Add handling of unsupported VT function

### LINKED ISSUE #377: Add (VT) Unsupported VT Function message
VT version 5 and higher Working Sets are required to send the `Unsupported VT Function` message when receiving unsupported functions from the VT server. We should also incorporate listening to the `VT Unsupported VT Function` to know when the VT doesn't support a particular command we sent.
It is allowed for VTs and Working Sets that were designed for VT version 4 and earlier to also incorporate these messages.

GwnDaan:
Closes #377

### DIFF
--- a/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
+++ b/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
@@ -1562,6 +1562,16 @@ namespace isobus
 		/// @returns true if the object was resized, otherwise false
 		bool resize_object(std::uint8_t *buffer, float scaleFactor, VirtualTerminalObjectType type);
 
+		/// @brief Extract from the cache whether a VT does not support a specific function
+		/// @param[in] function The function to check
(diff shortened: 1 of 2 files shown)


ad3154 (approved):

ad3154:
Do you think it would be valuable to also have a way to get a list if all previously cached unsupported functions? Or not really?

GwnDaan:
> Do you think it would be valuable to also have a way to get a list if all previously cached unsupported functions?

Hmm, I'm not sure. What do you think it can be used for? You currently only know that a function is unsupport after you tried to send it, so we can't really use that list at startup or something I think

ad3154:
> Hmm, I'm not sure. What do you think it can be used for? You currently only know that a function is unsupported after you tried to send it, so we can't really use that list at startup or something I think

Maybe unit tests? Haha. I'm not sure, it was just a thought, but I think that doing something on the client side based on some known list of unsupported functions could be challenging, so maybe it's not particularly helpful

Issue #403, opened 2024-01-10
Provide out-of-the-box solution for tracking and updating object pool values


GwnDaan:
Almost all applications using the VT client want to update values in their displayed pool. Since both the VT server and your application (VT client) track these values, syncing and/or tracking the values of these variables can be cumbersome and time-consuming. Worry no more, with a easy-to-use and out-of-the-box helper class we should be able to solve this issue.

My initial idea for the system flow behind the helper:
- when you want to change a tracked value from the VT client:
![AgIsoStack++ (4)](https://github.com/Open-Agriculture/AgIsoStack-plus-plus/assets/29043784/5ef09083-72a7-4997-b233-54bf61373217)
- when we receive a response from the VT server:
![AgIsoStack++ (3)](https://github.com/Open-Agriculture/AgIsoStack-plus-plus/assets/29043784/93d720ca-65fc-4b77-a8bf-1da7c646400e)

Does this make sense and would it work? The changes in tracked value coming from macros, still needs to be worked out

PR #405, opened 2024-01-21
[VT]: add a numeric value tracker and updater for working sets of a VT client

GwnDaan:
This is the first of several PR's to address #403, this one adds the fundamentals, as well as the tracking and updating of numeric values.


isobus/src/isobus_virtual_terminal_client_state_tracker.cpp
```diff
@@ -0,0 +1,107 @@
...
+				case static_cast<std::uint8_t>(VirtualTerminalClient::Function::VTChangeNumericValueMessage):
+				{
+					std::uint16_t objectId = message.get_uint16_at(1);
+					if (numericValueStates.find(objectId) != numericValueStates.end())
+					{
+						std::uint32_t value = message.get_uint32_at(4);
+						numericValueStates[objectId] = value;
+					}
+					break;
+				}
```
ad3154, line 101: The VT's not obligated to change the displayed value when it sends the ChangeNumericValueCommand message, since it relies on the client somewhat to ensure that the value which was set is valid, so when this happens, if we really want to be helpful, we may want to also transmit our version of the command back to the VT

isobus/include/isobus/isobus/isobus_virtual_terminal_client_update_helper.hpp
```diff
@@ -0,0 +1,49 @@
...
+		/// @param[in] step The step size to decrease the numeric value with.
+		/// @return True if the value was decreased successfully, false otherwise.
+		bool decrease_numeric_value(std::uint16_t objectId, std::uint32_t step = 1);
+
+	private:
+		std::shared_ptr<VirtualTerminalClient> client; ///< Holds the vt client.
+	};
+}; // namespace isobus
+
+#endif // ISOBUS_VIRTUAL_TERMINAL_CLIENT_UPDATE_HELPER_HPP
```
ad3154, line 49: Nit, missing newline at end of file

isobus/src/isobus_virtual_terminal_client_state_tracker.cpp
GwnDaan: Hmm I agree, though we currently already send a response back in the virtual terminal client class: https://github.com/Open-Agriculture/AgIsoStack-plus-plus//blob/daan/vt-track-and-update-states/isobus/src/isobus_virtual_terminal_client.cpp#L2559-L2578.

Do you think we should change that? I tried to made the tracker class standalone such that it in theory can track any VT client, not just ours. Hence sending responses in here is a little off I think

GwnDaan: Also the response value will be identical to the value per definition of the standard, and if there's no response from the client the VT will (for version 5 and later) act as if an unexpected shutdown occurred. Maybe the latter is not really incorporated yet into the state tracker, but it's a bit out-of-scope for this PR I feel like. I'll add a TODO for it :)

GwnDaan: Okay I delved a bit deeper into this, we might instead need to add a way of entering a so called "safe state" in the `VirtualTerminalClient` itself. When an unexpected shutdown occurs, the standard requires us to put all functions dependant on the VT operator interface into a known state such that it will not put the operator or machine at risk. I think we should definitely add proper implementation for that instead of just restarting the connection as we do now

ad3154: > though we currently already send a response back

Right, but that's not the same thing as commanding the value. I think sending the response there is appropriate.

> I tried to made the tracker class standalone such that it in theory can track any VT client, not just ours. Hence sending responses in here is a little off I think

To clarify - I'm not talking about the response, I mean we should send the non-VT version of the change numeric value command in the helper. This is important with older VTs, like version 4, which may not care about the responses.

> we might instead need to add a way of entering a so called "safe state"

Only if entering a safe state is a callback to the user's app in some way. If a shutdown occurs, the VT client state we have is meaningless - rather, it's up to the main app (the person integrating our library) to know what their implement is doing and take some action accordingly. This is not something we can control outside of some kind of notification to the user.

GwnDaan: > I mean we should send the non-VT version of the change numeric value command in the helper

Oh hmm yeah, you mean we should send a reply to the VT with a command message (0xA8)? Is it concerning the race condition that can occur where both the VT and working set apply a change at the same time? Would that be something we can also do in the VT client interface, I'm still unsure if the tracker is the right place for that

> it's up to the main app (the person integrating our library) to know what their implement is doing and take some action accordingly. This is not something we can control outside of some kind of notification to the user.

For sure, I indeed meant it in a way where the application indeed get's notified and can take action accordingly. Restarting the connection blindly like we do now can introduce some concerning side effects I imagine

ad3154: > you mean we should send a reply to the VT with a command message (0xA8)

Yeah, that one! We could maybe make these proposed A8 commands optional? But it seems like it could be a nice quality of life feature in the general sense.

> Would that be something we can also do in the VT client interface, I'm still unsure if the tracker is the right place for that

I don't think the client is the ideal place, because it's up to the application/the user to take the value set by the VT and figure out if that value is "valid" or even if they want to do some other behavior. That's kind of why I was think the helper made more sense, it's like a way to make the default behavior to accept whatever the VT says is happening. What do you think?

> Is it concerning the race condition that can occur where both the VT and working set apply a change at the same time?

Somewhat? We should think about that a bit probably, though I am not sure how I would prevent it... 

> For sure, I indeed meant it in a way where the application indeed get's notified and can take action accordingly. 

OK perfect, I can add some thoughts to that issue as well

GwnDaan: > it's up to the application/the user to take the value set by the VT and figure out if that value is "valid" or even if they want to do some other behavior. That's kind of why I was think the helper made more sense, it's like a way to make the default behavior to accept whatever the VT says is happening. What do you think?

Ahh okay now I understand. I agree that the client interface is not the best place providing a complete interface for that. Though the implementation you are trying to achieve is almost there already; We have the `add_vt_change_numeric_value_event_listener` in the client interface, and using that in combination with the `send_change_numeric_value` should yield your result. For simplicity this is already quite good, what do you think?

Let's take our seeder app as an example - I'm in the process of converting it to the helper as well:
```c++
void SeederVtApplication::handle_numeric_value_events(const isobus::VirtualTerminalClient::VTChangeNumericValueEvent &event)
{
	switch (event.objectID)
	{
		case statisticsSelection_VarNum:
		{
			auto newSelectedStatistic = static_cast<Statistics>(event.value);
			if (newSelectedStatistic != currentlySelectedStatistic)
			{
				currentlySelectedStatistic = newSelectedStatistic;
				txFlags.set_flag(static_cast<std::uint32_t>(UpdateVTStateFlags::UpdateStatisticsSelection_VarNum));
				txFlags.set_flag(static_cast<std::uint32_t>(UpdateVTStateFlags::UpdateSelectedStatisticsContainer_ObjPtr));
			}
		}
		break;

		case enableAlarms_VarNum:
		{
			alarmsEnabled = event.value;
			txFlags.set_flag(static_cast<std::uint32_t>(UpdateVTStateFlags::UpdateEnableAlarms_VarNum));
		}
		break;

		default:
			break;
	}
}

``` 

can be reduced to - while providing the same functionality:

```c++
void SeederVtApplication::handle_numeric_value_events(const isobus::VirtualTerminalClient::VTChangeNumericValueEvent &event)
{
	switch (event.objectID)
	{
		case statisticsSelection_VarNum:
		{
			// Acknowledge the selection change by sending it back to the VT
			VTClientInterface->send_change_numeric_value(statisticsSelection_VarNum, event.value);

			// Update the frame to show the new selected statistic
			std::uint16_t newContainer = isobus::VirtualTerminalClient::NULL_OBJECT_ID;
			auto correspondingContainer = STATISTICS_CONTAINER_MAP.find(static_cast<Statistics>(event.value));
			if (STATISTICS_CONTAINER_MAP.end() != correspondingContainer)
			{
				newContainer = correspondingContainer->second;
			}
			VTClientUpdateHelper.set_numeric_value(selectedStatisticsContainer_ObjPtr, newContainer);
		}
		break;

		case enableAlarms_VarNum:
		{
			// Acknowledge the selection change by sending it back to the VT
			VTClientInterface->send_change_numeric_value(enableAlarms_VarNum, event.value);
		}
		break;

		default:
			break;
	}
}
```

GwnDaan: Or do you reckon registering an extra callback inside the tracker class is more clear? I'm not sure what the default behavior should then be if one would not provide the callback. hmm...

ad3154: Yeah I guess I'm not really sure... the example you posted above is probably simple enough? But like, if we did register an extra callback in the tracker, I'd expect the entirety of `case enableAlarms_VarNum:` could even be done away with. IDK if that's better, or weirder haha

GwnDaan: I agree. It is probably better to provide it inside the tracker indeed, I can add it. Should be pretty easy now I think about it hahah

isobus/src/isobus_virtual_terminal_client_update_helper.cpp
```diff
@@ -0,0 +1,82 @@
...
+//================================================================================================
+
+#include "isobus/isobus/isobus_virtual_terminal_client_update_helper.hpp"
+
+#include "isobus/isobus/can_stack_logger.hpp"
+
+namespace isobus
+{
+	VirtualTerminalClientUpdateHelper::VirtualTerminalClientUpdateHelper(std::shared_ptr<VirtualTerminalClient> client) :
+	  VirtualTerminalClientStateTracker(client->get_internal_control_function()),
```
ad3154, line 17: Possibility to crash here if the shared_ptr is invalid... would be nice if we could catch that with an assert or something, or maybe a `nullptr == client ? nullptr :  client->get_internal_control_function()`

isobus/src/isobus_virtual_terminal_client_state_tracker.cpp
```diff
@@ -0,0 +1,107 @@
...
+					if (errorCode == 0)
+					{
+						std::uint16_t objectId = message.get_uint16_at(1);
+						if (numericValueStates.find(objectId) != numericValueStates.end())
+						{
+							std::uint32_t value = message.get_uint32_at(4);
+							numericValueStates[objectId] = value;
+						}
+					}
+					break;
```
ad3154, line 90: This is a nitpick but I prefer
```
case x:
{
}
break;
```

Over 
```
case x:
{
break;
}
```

if only for consistency with most of the rest of the repo

isobus/src/isobus_virtual_terminal_client_state_tracker.cpp
```diff
@@ -0,0 +1,107 @@
...
+		if (message.has_valid_source_control_function() &&
+		    message.is_destination(client) &&
+		    message.is_parameter_group_number(CANLibParameterGroupNumber::VirtualTerminalToECU))
+		{
+			std::uint8_t function = message.get_uint8_at(0);
+			switch (function)
+			{
+				case static_cast<std::uint8_t>(VirtualTerminalClient::Function::ChangeNumericValueCommand):
+				{
+					auto errorCode = message.get_uint8_at(3);
```
ad3154, line 80: Possibility to crash a few places in here. Should probably check the data length to ensure it's >= 1 for ` function = message.get_uint8_at(0);` and >= 8 for change numeric value?

isobus/src/isobus_virtual_terminal_client_state_tracker.cpp
```diff
@@ -0,0 +1,116 @@
...
+		    message.is_destination(client) &&
+		    message.is_parameter_group_number(CANLibParameterGroupNumber::VirtualTerminalToECU) &&
+		    (message.get_data_length() >= 1))
+		{
+			std::uint8_t function = message.get_uint8_at(0);
+			switch (function)
+			{
+				case static_cast<std::uint8_t>(VirtualTerminalClient::Function::ChangeNumericValueCommand):
+				{
+					if (message.get_data_length() == 8)
```
ad3154, line 81: Nit
```suggestion
					if (CAN_DATA_LENGTH == message.get_data_length())
```

isobus/src/isobus_virtual_terminal_client_state_tracker.cpp
```diff
@@ -0,0 +1,116 @@
...
+								numericValueStates[objectId] = value;
+							}
+						}
+					}
+				}
+				break;
+
+				case static_cast<std::uint8_t>(VirtualTerminalClient::Function::VTChangeNumericValueMessage):
+				{
+					if (message.get_data_length() == 8)
```
ad3154, line 99: Same here
```suggestion
					if (CAN_DATA_LENGTH == message.get_data_length())
```

isobus/src/isobus_virtual_terminal_client_state_tracker.cpp
```diff
@@ -0,0 +1,116 @@
...
+namespace isobus
+{
+	VirtualTerminalClientStateTracker::VirtualTerminalClientStateTracker(std::shared_ptr<ControlFunction> client) :
+	  client(client)
+	{
+	}
+
+	void VirtualTerminalClientStateTracker::initialize()
+	{
+		CANNetworkManager::CANNetwork.add_any_control_function_parameter_group_number_callback(static_cast<std::uint32_t>(CANLibParameterGroupNumber::VirtualTerminalToECU), process_rx_message, this);
```
ad3154, line 27: Probably want some way to remove this? Otherwise it could get called after this object is destructed?

GwnDaan: Oh yaeh definitely!

isobus/include/isobus/isobus/isobus_virtual_terminal_client_update_helper.hpp
```diff
@@ -0,0 +1,63 @@
...
+		/// be rejected by sending the current value back to the VT.
+		/// @param[in] callback The callback function to register, or nullptr to unregister.
+		void set_callback_validate_numeric_value(const std::function<bool(std::uint16_t, std::uint32_t)> &callback);
+
+	private:
+		/// @brief Processes a numeric value change event
+		/// @param[in] event The numeric value change event to process.
+		void process_numeric_value_change_event(const VirtualTerminalClient::VTChangeNumericValueEvent &event);
+
+		std::shared_ptr<VirtualTerminalClient> client; ///< Holds the vt client.
```
ad3154, line 56: Last comment I swear, haha.

`Field "client" shadows a field of the same name in base class "VirtualTerminalClientStateTracker".`

Probably want to fix that one sonar issue

GwnDaan: Yeah no worries, I initially shadowed intentionally but I guess it can be confusing for others, so i have renamed it 👍🏽. There are probably more things that are not optimal / not working. I've been changing quite a bit to get the advanced sample to work properly so will be a few more PR's coming hahah

ad3154 (approved):

PR #406, opened 2024-01-21
[VT]: Add state tracker and updater for active masks

GwnDaan:
Sequel of #405 and part of #403. With this addition you can update active masks with the helper class. The benefit is that you don't have to do any preliminary checks whether or not the update is necessary, the helper class will do that for you


isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
```diff
@@ -1606,9 +1594,6 @@ namespace isobus
 
 		// Status message contents from the VT
 		std::uint32_t lastVTStatusTimestamp_ms = 0; ///< The timestamp of the last VT status message
-		std::uint16_t activeWorkingSetDataMaskObjectID = NULL_OBJECT_ID; ///< The active working set data mask object ID
-		std::uint16_t activeWorkingSetSoftKeyMaskObjectID = NULL_OBJECT_ID; ///< The active working set's softkey mask object ID
-		std::uint8_t activeWorkingSetMasterAddress = NULL_CAN_ADDRESS; ///< The active working set master address
```
ad3154, line 1611: These still seem useful to me in the VTClient, since I think using the helper should be optional

ad3154:
Since you've got this one set to merge into [daan/vt-track-and-update-states](https://github.com/Open-Agriculture/AgIsoStack-plus-plus/tree/daan/vt-track-and-update-states) should we do reviews on this one before merging that one to main?

GwnDaan:
> should we do reviews on this one before merging that one to main?

Hmm, I'd prefer the opposite. If we merge the other one first, the this will simply be based upon main. Will make things a bit less complicated I think

ad3154 (approved):

Issue #350, opened 2023-11-21
Task Controller Server


Eric-Bwr:
Good day,

is there something planned in this regard?
Nice work btw!

GwnDaan:
Hey there, welcome to the community! Ultimately we would want to support everything inside the ISO11783 standard with this stack, including the task controller server. Though the task controller server has not yet been touched upon, my wild guess is that it will still take quite a while for us to get to. But maybe @ad3154 has some plans already? Also we welcome any contributions you can make to help drive progress 😄

Eric-Bwr:
Hello,
unfortunately, I dont have much experience with the behind the scenes stuff of a tc server. We are using a commercial library which is... lacking some features... And the developers are.... lets say they are too busy or whatever... 
So this library is an absolut dream. But I fear due to my limited time and lack of deep knowledge I cant help you guys out. 
Im more than happy to help in testing though. Ive got several implements at my disposal.

Eric-Bwr:
Wonderfull! Sure! Tomorrow I will provide a list of functionality which we have right now, and would still be nice to have, and a list which we are missing out on. Thank you very much!
And of cause no pressure. But lets get this party started :)

Eric-Bwr:
Too much work today, sry guys, next week :)

Eric-Bwr:
So, lets see what would be nice to have in regards to the TC-Server:

- A way to read in TASKDATA.XML Files and set the store dir path for the coverage
- Setting the task status (Start, Paused, Stop)
- Callbacks for task status 
- Callbacks for AddDevice / RemoveDevice (when Implement connects to can, could handle things like pivot type (trailed or fixed), and getting the implement (booms (section)) geometry aswell as adding a x and y offset to them)
- Callbacks and setter within those for manipulating the section control state 
- Same thing for rate control
- Possibility to set the pre on off time
- Setter for section overlap
- Setter for Products (rate control related)
- Setter to apply general gps offset (not identical to the AddDevice Boom Offset)
- Being able to clear the coverage and set custom values at X and Y index

The lib we use has its own direction api, so they try to compute the working direction based on position data.
We would like to use our own heading information to resolve some latency issues.
But here are some features in that regards:
- Callbacks from direction api (optional -> gps loss, gps update)
- Setter / Getter for machine direction
- Setter / Getter for heading information (or even parse the NMEA2000 cog & sog pgn and other various protocols)

So yeah, this is the ruff idea I guess.
[...]

PR #407, opened 2024-01-21
Task controller server

ad3154:
## Task Controller Server

This is an abstract interface for an ISO11783-10 version 4 task controller or data logger server. This portion of a TC server handles communication with the implement, receives the DDOP, and provides hooks into your application so that you can command the implements' sections, change triggers, set values, and generally manage each implement.

### Features

- Automatic DDOP upload/handshaking with clients
- Settable measurement commands and triggers for getting data back from the implement
- Value commands and acknowledgements
- Integrates well with our DeviceDescriptorObjectPool class for parsing received DDOPs.
- Has a minimum example to help 

### Todo/things to consider

I made this PR a draft while I consider or improve the following:

- [x] Improve unit test coverage
- [x] Consider abstraction/helper for gathering geometry data from an implement as mentioned in #350 **Edit:** see #427
- [x] Consider abstraction/helper for "products" and rate control in general as mentioned in #350 **Edit:** see #427
- [x] ~~Similar to how we have an ISOXML serializer, add an ISOXML reader/deserializer~~ Edit: It would not be too bad to only parse out DDOP information, but to accurately do this would require > 50 classes! So I think we can put that in a different PR.
- [x] Documentation? ~~Consider additions to the sphinx docs.~~ Edit: Added sphinx docs

### Screenshot
image


examples/task_controller_server/main.cpp
```diff
@@ -0,0 +1,173 @@
...
+	TestDeviceNAME.set_device_class(0);
+	TestDeviceNAME.set_function_code(static_cast<std::uint8_t>(isobus::NAME::Function::TaskController));
+	TestDeviceNAME.set_identity_number(20);
+	TestDeviceNAME.set_ecu_instance(0);
+	TestDeviceNAME.set_function_instance(0); // TC #1. If you want to change the TC number, change this.
+	TestDeviceNAME.set_device_class_instance(0);
+	TestDeviceNAME.set_manufacturer_code(1407);
+
+	auto TestInternalECU = isobus::InternalControlFunction::create(TestDeviceNAME, 247, 0); // The preferred address for a TC is defined in ISO 11783
+	MyTCServer server(TestInternalECU, 4, 255, 16, 0x17); // 4 booms, 255 sections, 16 channels, some options configured using isobus::TaskControllerServer::ServerOptions, such as "Supports Documentation"
```
GwnDaan, line 156: Hmm, might be good to have some way to construct the 0x17 options more describtively to set a good example. not sure how though... maybe a simple ServerOptionsBuilder class/struct would suffice?

isobus/include/isobus/isobus/isobus_task_controller_server.hpp
```diff
@@ -0,0 +1,534 @@
...
+		virtual ~TaskControllerServer();
+
+		/// @brief Deleted copy constructor
+		TaskControllerServer(TaskControllerServer &) = delete;
+
+		/// @brief Deleted assignment operator
+		TaskControllerServer &operator=(const TaskControllerServer &) = delete;
+
+		/// @brief Shuts down the TC server, unregisters PGN callbacks
+		void terminate();
```
GwnDaan, line 128: Nit: would be more clear if we move this function down to the (de)init functions

GwnDaan (commented):
Nicee! I took a first look at the few first files, and it looks really promising. Left a few comments here and there that could be valuable :)

isobus/CMakeLists.txt
```diff
@@ -83,7 +84,7 @@ set(ISOBUS_INCLUDE
     "isobus_functionalities.hpp"
     "isobus_speed_distance_messages.hpp"
     "isobus_maintain_power_interface.hpp"
-    "isobus_virtual_terminal_objects.hpp"
```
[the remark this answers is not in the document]
ad3154: Actually, removed a duplicate, since `isobus_virtual_terminal_objects.hpp` was already included up around line 77

examples/task_controller_server/main.cpp
ad3154: Yeah bit fields have always felt weird/tricky to construct nicely.  A tiny helper class could work better than some arcane hex value I think

GwnDaan:
I did some testing with the current interface - I know it's still draft - it's working exceptionally well from the start, just a few lines of code on top of the boilerplate and I had section control over ISOBUS working! Here's a simple demonstration using AOG as the guidance system providing the section states: 

https://github.com/Open-Agriculture/AgIsoStack-plus-plus/assets/29043784/95a5b102-8b5d-4041-8acf-4957b765447e

Great stuff @ad3154!

isobus/src/isobus_task_controller_server.cpp
```diff
@@ -0,0 +1,1057 @@
...
+		                                                      serverControlFunction,
+		                                                      nullptr,
+		                                                      CANIdentifier::CANPriority::Priority3);
+	}
+
+	bool TaskControllerServer::send_version(std::shared_ptr<ControlFunction> clientControlFunction) const
+	{
+		std::array<std::uint8_t, CAN_DATA_LENGTH> payload = {
+			static_cast<std::uint8_t>(TechnicalDataCommandParameters::ParameterVersion) << 4,
+			static_cast<std::uint8_t>(TaskControllerVersion::SecondPublishedEdition),
```
GwnDaan, line 778: issue: the version needs to be (optionally) configurable. If a TC were to support only the "TC-SC generation 1" it needs to send the version number 3, instead of the 4 here.

isobus/src/isobus_task_controller_server.cpp
```diff
@@ -0,0 +1,1057 @@
...
+							if (nullptr != get_active_client(rxMessage.get_source_control_function()))
+							{
+								if (get_active_client(rxMessage.get_source_control_function())->isDDOPActive)
+								{
+									std::uint16_t DDI = static_cast<std::uint16_t>(rxData[2]) | (static_cast<std::uint16_t>(rxData[3]) << 8);
+									std::uint16_t elementNumber = static_cast<std::uint16_t>(rxData[0] >> 4) | (static_cast<std::uint16_t>(rxData[1]) << 4);
+									std::int32_t processVariableValue = static_cast<std::int32_t>(static_cast<std::uint32_t>(rxData[4]) |
+									                                                              (static_cast<std::uint32_t>(rxData[5]) << 8) |
+									                                                              (static_cast<std::uint32_t>(rxData[6]) << 16) |
+									                                                              (static_cast<std::uint32_t>(rxData[7]) << 24));
```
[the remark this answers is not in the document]
ad3154: Hmm, the standard defines "process variable value" (SPN 5202) as "signed long integer type" so it may make more sense to change it in the client

isobus/include/isobus/isobus/isobus_task_controller_server.hpp
```diff
@@ -0,0 +1,534 @@
...
+			bool isDDOPActive = false; ///< Whether or not the client's DDOP is active.
+		};
+
+		/// @brief Stores messages received from task controller clients for processing later.
+		/// @param message The message received from the client.
+		/// @param parentPointer A context variable that can be used to find this class's instance.
+		static void store_rx_message(const CANMessage &message, void *parentPointer);
+
+		/// @brief Processes messages received from task controller clients.
+		void process_rx_messages();
```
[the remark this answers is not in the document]
ad3154: I wasn't really sure what the best way to handle this would be... if we handle the messages on the CAN stack thread, the user needs to know to use aggressive mutexing in their callbacks, but if we do it when `update` is called, then it all happens on their app's thread, which maybe seems nicer? With this approach I only need to mutex internally between `store_rx_message` and `process_rx_messages` (adding this in next commit).

If that seems reasonable to you, I'll add additional comments explaining this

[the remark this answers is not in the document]
GwnDaan: Yeah sounds good to me. I think it's a good trade-off to save complexity in the user' application. We can also expose a condition variable to let the application know there are messages available to process, for when one wants to respond asap to received messages without busy-waiting

examples/task_controller_server/main.cpp
ad3154: See if you like the way I have it now. It looks like this:
```
MyTCServer server(TestInternalECU,
                  4, // Booms
                  255, // Sections
                  16, // Channels
                  isobus::TaskControllerOptions()
                    .with_documentation()
                    .with_implement_section_control()
                    .with_tc_geo_with_position_based_control());
```

isobus/src/isobus_task_controller_server.cpp
ad3154: Added as an optional parameter to the constructor of the server

examples/task_controller_server/main.cpp
GwnDaan: Perfect!

Eric-Bwr:
Oh my days, this is literally amazing! Looking forward to this :) I would be up for some tests if needed. We have various implements if needed.

isobus/src/isobus_task_controller_server.cpp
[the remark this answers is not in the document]
ad3154: Addressed this comment in another PR #464

ad3154:
OK, I think is is ready for re-review or approval at your convenience @GwnDaan

GwnDaan (approved):

PR #411, opened 2024-01-23
Github PR template

0x-a6:
## Describe your changes

Adds a PR template to github.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,20 @@
+## Describe your changes
+
+<!-- Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List any dependencies that are required for this change. -->
+
+Fixes # (issue)
+
+## How has this been tested?
+
+<!-- Please describe the tests that you ran to verify your changes.
+Provide instructions so we can reproduce. Also list any relevant details for your test configuration. -->
+
+```bash
+TODO
+```
+
+## Checklist before requesting a review
+- [ ] I have performed a self-review of my code
+- [ ] If it is a core feature, I have added thorough tests.
+- [ ] Do we need to implement analytics?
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,20 @@
...
+<!-- Please describe the tests that you ran to verify your changes.
+Provide instructions so we can reproduce. Also list any relevant details for your test configuration. -->
+
+```bash
+TODO
+```
+
+## Checklist before requesting a review
+- [ ] I have performed a self-review of my code
+- [ ] If it is a core feature, I have added thorough tests.
```
ad3154, line 18: Not sure about these bottom items. We can see if they did tests or not in the Sonar quality gate, don't really have a way to implement any kind of analytics, and since it's a library, everything is sort of a product update? What do you think @GwnDaan

ad3154: Specifically
```
If it is a core feature, I have added thorough tests.
Do we need to implement analytics?
Will this be part of a product update? If yes, please write one phrase about this update.
```

0x-a6: That makes sense, is there anything sonar doesn't catch - we have it at work but can't recall if it will check for appropriate documentation checks/comments? I think it does. 

Do you think there is no need for a checklist due to the review process?

ad3154: Well, I don't think we have it set to check for documentation - instead we have a github action that makes sure the doxygen builds with no warnings [like this](https://github.com/Open-Agriculture/AgIsoStack-plus-plus/actions/runs/7624463723/job/20766638223)

Overall I think our build pipelines are pretty robust at catching serious things, I think we could probably omit the checklist

0x-a6:
@ad3154 or @GwnDaan or someone with write access needs to squash merge/merge. :pray:

PR #412, opened 2024-01-24
[Core]: Add a reference for ISO and SAE preferred addresses

ad3154:
## Describe your changes

This was inspired by a comment on #407 and adds a reference for standard preferred control function addresses. This can be used as a reference for developers looking to assign a control function's preferred SA if they use a "well-known" standard function code.

As usual, users should not expect that the preferred address will be used if their NAME fails to arbitrate against another NAME on the bus that also wants that address. In those cases the stack will automatically select an appropriate address in the self configurable range if the NAME allows arbitrary address capabilities.

## How has this been tested?

This information was directly sourced from the publically available ISO11783-11 database [here](https://www.isobus.net/isobus/attachments/isoExport_xlsx.zip). Since the change is header-only, no tests were required.

### DIFF
--- a/isobus/include/isobus/isobus/isobus_preferred_addresses.hpp
+++ b/isobus/include/isobus/isobus/isobus_preferred_addresses.hpp
@@ -1,0 +1,225 @@
+//================================================================================================
+/// @file isobus_preferred_addresses.hpp
+///
+/// @brief This is a reference for control function's preferred addresses as defined by
+/// ISO 11783-11 and/or SAE. Preferred addresses are industry group specific.
+/// You should use these when your are creating a control function that is a well known
+/// function type, but if your control function doesn't arbitrate for that address, the
+/// stack will claim for you in the dynamic address range.
+///
+/// @author Adrian Del Grosso
+///
+/// @copyright 2024 The Open-Agriculture Developers
+//================================================================================================
+#ifndef ISOBUS_PREFERRED_ADDRESSES_HPP
+#define ISOBUS_PREFERRED_ADDRESSES_HPP
+
+#include <cstdint>
+
+namespace isobus
+{
+	/// @brief This namespace contains all defined preferred addresses for control functions, which
+	/// should be used when creating a control function that is a well known function type.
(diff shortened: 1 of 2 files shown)


isobus/include/isobus/isobus/isobus_preferred_addresses.hpp
```diff
@@ -0,0 +1,210 @@
...
+
+namespace isobus
+{
+	/// @brief This namespace contains all defined preferred addresses for control functions, which
+	/// should be used when creating a control function that is a well known function type.
+	namespace preferred_addresses
+	{
+		/// @brief Industry Group 1 applies to on-highway equipment. This
+		/// This enumerates all preferred addresses for this industry group.
+		enum class IndustryGroup1 : std::uint8_t
```
GwnDaan, line 27: thought: since I'd imagine this only to be used as a lookup table, could it be more user friendly to implicitly convert to `std::uint8_t` here? So we can just use 

```c++
	auto TestInternalECU = isobus::InternalControlFunction::create(TestDeviceNAME, isobus::preferred_address::IndustryGroup2::TaskController_MappingComputer, 0);
```

instead of 

```c++
	auto TestInternalECU = isobus::InternalControlFunction::create(TestDeviceNAME, static_cast<std::uint8_t>(isobus::preferred_address::IndustryGroup2::TaskController_MappingComputer), 0);
```

I know Sonar won't like it, but this might be an exception?

ad3154: Yep, I agree that does seem nicer (sorry sonar). I will make them regular enums instead.

ad3154: To get this behavior to work without redefinition errors (there are multiple enumerators with the same name) I had to do:

```
namespace preferred_addresses
{
	namespace IndustryGroup1
	{
		enum PreferredAddress : std::uint8_t
		{
```
For example

GwnDaan (approved):

PR #413, opened 2024-01-24
[TC]: Add ISO11783-11 database lookup functionality

ad3154:
## Describe your changes

- Added the the main content of the ISOBUS DDI database to use in task controllers when processing DDOPs.
- Added a python script to retrieve the DDI database from [isobus.net](https://www.isobus.net/isobus/dDEntity) and generate associated C++ files.

@GwnDaan I'd like your help assessing the format/schema of this "database". I currently have it implemented as a dumb C style array because that seemed to be the most non-c++14+ way of ensuring that on microcontrollers the data ends up in flash instead of in RAM.

## How has this been tested?

Added unit tests to validate a couple random DDIs return expected values, plus one undefined DDI returns a known, default value.

You can run the code generation yourself by running the included python script. `python3 code_gen_ddi_database.py`


GwnDaan:
> I currently have it implemented as a dumb C style array

I really like this "dumb" way, no need to overcomplicate things there. Though I do think there is a few things we could improve upon:

1. Instead of replacing the entire file with the python script, we can use a bit of regex and replace just the parts we're interested in (for [inspiration](https://stackoverflow.com/a/59021728)). This will save us some confusion with having the file duplicated in two places
2. Generating an enum for the DDI name mapping to the index number. I imagine we should be interested in the meaning of the DDI rather than the number behind it
3. Generating an enum for the units, as comparing a string can be more user-error prone than comparing it with another enum

ad3154:
> we can use a bit of regex

That could make sense, as long as it also adds in new ones as well, though I would like to let someone else tackle that if possible, I despise regex, and I'm not sure if matching on any part of a definition would result in a perfect result depending on what they change.

> Generating an enum for the DDI name mapping to the index number. I imagine we should be interested in the meaning of the DDI rather than the number behind it

So, I think it depends on if you're a server or a client, so we may want both. Servers will care about the DDI's number, not the name. Clients will care about the name, not the number, generally. I think it probably makes sense to add this enum based approach additionally I guess.

> Generating an enum for the units, as comparing a string can be more user-error prone than comparing it with another enum

I think the use case is again maybe a bit different based on different applications' requirements. I think if you want to display the units, having the string is nice, but if you want to process the DDI in some way, you'll want the enum to know what to do with it in an efficient way. So I think once again, maybe having both is ideal

GwnDaan:
> I would like to let someone else tackle that if possible

I proposed the changes in #415. Seems to be working quite well as long as the file is in a valid state. 

> So I think once again, maybe having both is ideal

I agree, including both seems to be the best of both worlds here for the reasons you provided

GwnDaan (approved):
LGTM. I think my requested enums is probably something we need to tackle once we bump into it. And since it will be able to happily live alongside to the current implementation, we can just make a PR for that later.

PR #414, opened 2024-01-24
Add vscode dev container

0x-a6:
## Describe your changes

Add's a MS dev container for ubuntu 22.04 - gcc version 11.3.0-1ubuntu1~22.04.1 

## How has this been tested?

Tested on Ubuntu 23.10 desktop, 
`Docker version 24.0.7, build afdd53b`, 
vscode 1.85.2, 
Dev containers version `2023-12-09, 06:02:20`

Ran build as per [CONTRIBUTING.md](https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/main/CONTRIBUTING.md#setting-up-a-development-environment-with-gui) without examples.

### update 2024-01-25
Added can-utils. Ran the `./scripts/setup_vcan.sh` on the host. Opened two teminal tty's inside vscode, worked sending and seeing message on candump.
image

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -13,5 +13,5 @@ RUN if [ "${REINSTALL_CMAKE_VERSION_FROM_SOURCE}" != "none" ]; then \
 # RUN su vscode -c "${VCPKG_ROOT}/vcpkg install <your-port-name-here>"
 
 # [Optional] Uncomment this section to install additional packages.
-# RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
-#     && apt-get -y install --no-install-recommends <your-package-list-here>
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends can-utils
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,17 @@
+FROM mcr.microsoft.com/devcontainers/cpp:0-ubuntu-22.04
+ARG REINSTALL_CMAKE_VERSION_FROM_SOURCE="none"
+
+# Optionally install the cmake for vcpkg
+COPY ./reinstall-cmake.sh /tmp/
+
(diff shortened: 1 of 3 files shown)


GwnDaan:
Hmm this is cool, I tested it on Windows and without too much hassle I got a working ubuntu image running. I'm wondering if it would be possible to automatically setup a isolated vcan bus inside the container. Do you have any experience with that?

GwnDaan:
Out-of-scope for this PR, but on the topic of vscode, would it be nice to provide a list of [recommended extensions](https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions) to get new users up and running?

0x-a6:
I wanted to submit a separate PR for a vxcan setup - but basically, as you know first hand, vcan is a host operating system feature where the networking needs to be `--network host`, which reminds me, that needs to be in this PR :joy: 

The reason it's on the host networking is because it uses a networking interface (on Linux), I also think it's not great because that's privilaged system access (to your host network) to just to run virtual CAN. So I'd like to explore a docker container option, which keeps it all in the user space. The option I've found so far is to do it with vxcan connections and gateways. Very easy for one container as it would be just xvcan to eth bridge, but I'm imagining you want to bring up two containers, one is a vt server and the other has vt client for example. And that's a bit more tricky. Obviously this would be an example code that could be inside a tutorial for the library, or developer options. Here's a great article I found on the issue: https://www.systec-electronic.com/en/demo/blog/article/news-socketcan-docker-the-solution

I'll add networking host to this PR.

0x-a6:
I've tested this on Apple M2 macOS 14.3, docker desktop 4.26.1, which meant I had to run `softwareupdate --install-rosetta` from terminal before installing docker following this installation guide: https://docs.docker.com/desktop/install/mac-install/

It built against the gcc 11.3 from the container fine, again without examples nor a can test.

GwnDaan:
Nice! I assume the vxcan / vcan won't work on host systems other than linux?

GwnDaan (approved):

ad3154 (approved):

PR #415, opened 2024-01-25
[TC]: Add regex for updating values in DDI lookup table

GwnDaan:
## Describe your changes

As discussed in #413, it was probably nice to have the update script only modify the parts it should be interested in. This preserves the possibility of continuing to edit the file manually where it is not auto-generated. 

## How has this been tested?

Run the update script after manually modifying the `DDI_ENTRIES` table, and testing whether it will return the same content as before. And ran the unit-test afterwards just to be sure


ad3154 (approved):
Big thanks 😁

PR #419, opened 2024-01-26
[ETP]: Fixed issues where we could CTS too many packets

### LINKED ISSUE #418: [ETP Rx]: ECTS requested packets exceeds message size
**Describe the bug**
When receiving ETP sessions, we can sometimes incorrectly issue a clear to send to the source control function with an incorrectly limited packet count, specifically on the final group of packets.

I have specifically seen a situation where we were receiving a session of 2291 bytes, and when the final group of packets was to be sent, we cleared the sender for 16 packets when only 8 were needed, causing them to abort the transfer with the ISO11783-3 error code `ECTS requested packets exceeds message size`

We should limit the CTS to be only as many frames as are required to complete the session in this case.

**Supporting Documentation**

image

can_extended_transport_protocol.cpp line 818
image

- OS: Windows
- Compiler MSVC
- CAN Driver: SysTec

ad3154:
## Describe your changes

This change limits the ETP CTS to be only as many frames as are required to complete a session when receiving an ETP session. This change was required because in #418 I was seeing a situation where a control function was aborting our ETP Rx session because we CTS'd too many packets with the ISO 11783-3 reason "ECTS requested packets exceeds message size" as described in ISO11783-3 Table 9.

Fixes #418 

## How has this been tested?

I used the TC server branch and the same control function to recreate the failing transfer, and it was successful.
Also, validated unit tests continued to pass.

### DIFF
--- a/isobus/src/can_extended_transport_protocol.cpp
+++ b/isobus/src/can_extended_transport_protocol.cpp
@@ -799,9 +799,16 @@ namespace isobus
 	bool ExtendedTransportProtocolManager::send_clear_to_send(std::shared_ptr<ExtendedTransportProtocolSession> &session) const
 	{
 		std::uint32_t nextPacketNumber = session->get_last_packet_number() + 1;
+		std::uint8_t packetLimit = session->get_cts_number_of_packet_limit();
+
+		if (packetLimit > session->get_number_of_remaining_packets())
+		{
+			packetLimit = static_cast<std::uint8_t>(session->get_number_of_remaining_packets());
+		}
+
 		const std::array<std::uint8_t, CAN_DATA_LENGTH> buffer{
 			CLEAR_TO_SEND_MULTIPLEXOR,
-			session->get_cts_number_of_packet_limit(),
+			packetLimit,
 			static_cast<std::uint8_t>(nextPacketNumber & 0xFF),
 			static_cast<std::uint8_t>((nextPacketNumber >> 8) & 0xFF),
 			static_cast<std::uint8_t>((nextPacketNumber >> 16) & 0xFF),


isobus/src/can_extended_transport_protocol.cpp
```diff
@@ -799,9 +799,18 @@ namespace isobus
...
 		std::uint32_t nextPacketNumber = session->get_last_packet_number() + 1;
+		std::uint32_t bytesProcessed = (PROTOCOL_BYTES_PER_FRAME * session->get_last_packet_number());
+		std::uint8_t numberOfPacketsToClear = session->get_cts_number_of_packet_limit();
+
+		if (((PROTOCOL_BYTES_PER_FRAME * numberOfPacketsToClear) + bytesProcessed) > (session->get_message_length()))
+		{
+			std::uint32_t bytesRemaining = session->get_message_length() - bytesProcessed;
+			numberOfPacketsToClear = (bytesRemaining / 7) + static_cast<std::uint8_t>(0 != (bytesRemaining % 7));
+		}
+
```
GwnDaan, line 810:
```suggestion
		std::uint8_t packetLimit = session->get_cts_number_of_packet_limit();
		if (packetLimit > session->get_number_of_remaining_packets())
		{
			packetLimit = static_cast<std::uint8_t>(session->get_number_of_remaining_packets());
		}
```

This was my fix haha, might be a bit friendlier to look at?

GwnDaan: Not sure if it will actually fix it though, haven't tested it. But it should do the same

ad3154: I'll test it real quick - it does look nicer, I just didn't know that get_number_of_remaining_packets existed haha

ad3154: Yep, that works! Suggestion shamelessly stolen

GwnDaan (approved):

GwnDaan:
Good catch! I totally missed it due to no unit tests hahah

PR #420, opened 2024-01-26
[Core]: Require complete message representation at construction time

GwnDaan:
## Describe your changes

This should make it a bit more self-explanatory what the message needs to be functional, instead of needing to know which loosely-required functions you have to call before you get a functional message. Also did a few cleanups of the CANNetworkManager where older things were still present or unclear

## How has this been tested?

Ran the virtual terminal examples to ensure core communication is still correct

### DIFF
--- a/isobus/include/isobus/isobus/can_message.hpp
+++ b/isobus/include/isobus/isobus/can_message.hpp
@@ -36,7 +36,6 @@ namespace isobus
 		{
 			Transmit, ///< Message is to be transmitted from the stack
 			Receive, ///< Message is being received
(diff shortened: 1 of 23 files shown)


ad3154 (approved):
Seems OK, probably makes sense to require these values up front anyways. Sonar's complaint about `new` we should maybe look at if/when we refactor fast packet protocol.

ad3154:
One note... removing `receive_can_message(const CANMessage &message);` made it impossible to receive long messages in the unit tests without actually sending all the transport messages I think, which is kind of unfortunate. I had a unit test in the TC server for receiving the entire DDOP but now have to leave it out I think.

PR #421, opened 2024-01-26
[Logger]: Made it possible to compile out the logger

### LINKED ISSUE #416: Ability to compile away CANStackLogger
The strings needed for the log messages require quite a lot of space in the binary. For releasing purposes it could be useful to be able to compile them away, especially on (embedded) platforms with low flash sizes.

ad3154:
## Describe your changes

I'm not sure if this is the route we want to go with this, but this change allows compiling out all logging strings.

- Added a cmake option for disabling the logger to save ROM space. Disabling the logger was accomplished by adding a macro to remove the content of logging operations when the cmake option is set.

Sonar is not gonna like this one I suspect. I am open to other approaches - just wanted to submit this one for opinions.

Fixes #416 

## How has this been tested?

Testing still in progress...

Initial results are that isobus.lib was reduced by 7ish percent with MSVC (debug) and the unit test executable (which generates many log statements) only decreased size by about 1 percent.

### DIFF
--- a/hardware_integration/src/innomaker_usb2can_windows_plugin.cpp
+++ b/hardware_integration/src/innomaker_usb2can_windows_plugin.cpp
@@ -66,7 +66,7 @@ namespace isobus
 			}
 			if (allChannelsClosed)
 			{
-				isobus::CANStackLogger::info("[InnoMaker-Windows] All channels closed, closing driver instance");
+				LOG_INFO("[InnoMaker-Windows] All channels closed, closing driver instance");
 				driverInstance->setdown();
 				driverInstance = nullptr;
 			}
@@ -238,7 +238,7 @@ namespace isobus
 
 				default:
 				{
-					isobus::CANStackLogger::error("[InnoMaker-Windows] Unsupported baudrate with index " + isobus::to_string(baudrate) + " in InnoMakerUSB2CANWindowsPlugin::Baudrate enum.");
+					LOG_ERROR("[InnoMaker-Windows] Unsupported baudrate with index " + isobus::to_string(baudrate) + " in InnoMakerUSB2CANWindowsPlugin::Baudrate enum.");
 					return;
 				}
 				break;
@@ -248,7 +248,7 @@ namespace isobus
 		}
 		else
 		{
-			isobus::CANStackLogger::error("[InnoMaker-Windows] No device found on channel " + isobus::to_string(channel));
+			LOG_ERROR("[InnoMaker-Windows] No device found on channel " + isobus::to_string(channel));
 		}
 	}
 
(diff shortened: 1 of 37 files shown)


hardware_integration/src/innomaker_usb2can_windows_plugin.cpp
```diff
@@ -238,7 +238,7 @@ namespace isobus
 
 				default:
 				{
-					isobus::CANStackLogger::error("[InnoMaker-Windows] Unsupported baudrate with index " + isobus::to_string(baudrate) + " in InnoMakerUSB2CANWindowsPlugin::Baudrate enum.");
+					LOG_IF_ENABLED(isobus::CANStackLogger::error("[InnoMaker-Windows] Unsupported baudrate with index " + isobus::to_string(baudrate) + " in InnoMakerUSB2CANWindowsPlugin::Baudrate enum."));
```
GwnDaan, line 241:
```suggestion
					LOG_ERROR("[InnoMaker-Windows] Unsupported baudrate with index " + isobus::to_string(baudrate) + " in InnoMakerUSB2CANWindowsPlugin::Baudrate enum."));
```
thought: could we achieve something like this? And then for each level a different define

ad3154: Probably, that would be much shorter which is nice. I can work on that, probably easy enough to find and replace them all

ad3154: Though, I will say that lots of libraries use `LOG_<level>` as macros, so we may conflict with them... we may want a more unique prefix

ad3154: I made it `LOG_<LEVEL>` but to accommodate variadic arguments it's maybe a bit odd in that you need an internal set of parentheses... It's probably fine I guess, since it's internal logging.

GwnDaan: Hmm okay I see, probably not ideal indeed. This can be a solution for it:

```c++
#define LOG_CRITICAL(...) isobus::CANStackLogger::critical(__VA_ARGS__)
```

Only thing is that you can't explicitly state the needed arguments in the macro itself, we'd have to rely on doxygen for that. Maybe something like:
```c++
/// @param[in] logString A log statement
/// @param[in] args (optional) A list of printf style arguments to format the logString with
```

ad3154: Added your suggestion, though doxygen was mad about documenting `__VA_ARGS__`. I like having the `param` documentation so I surrounded that definition with `//! @cond Doxygen_Suppress`

isobus/include/isobus/isobus/can_stack_logger.hpp
```diff
@@ -172,4 +176,45 @@ namespace isobus
...
+#ifdef DISABLE_CAN_STACK_LOGGER
+/// @brief A macro which removes a "critical" log statement depending on the state of DISABLE_CAN_STACK_LOGGER
+/// @param logString A log statement
+#define LOG_CRITICAL(logString)
+/// @brief A macro which removes a "error" log statement depending on the state of DISABLE_CAN_STACK_LOGGER
+/// @param logString A log statement
+#define LOG_ERROR(logString)
+/// @brief A macro which removes a "warning" log statement depending on the state of DISABLE_CAN_STACK_LOGGER
+/// @param logString A log statement
+#define LOG_WARNING(logString)
```
GwnDaan, line 189: image
I get the error now that "macro "LOG_WARNING" passed 4 arguments, but takes just 1" when I disable the can stack logger. I think adding the `...` here as well will fix that

GwnDaan (dismissed):

GwnDaan (approved):

PR #422, opened 2024-01-26
[Misc]: Fix -Wpedantic issues

ad3154:
## Describe your changes

Removed random extra semicolons which showed up as warnings with -Wpedantic

## How has this been tested?

Tested that unit tests compile, which should be enough for this one.


GwnDaan (approved):
My bad, maybe we can make Github Actions catch these warnings somehow?

ad3154:
> My bad, maybe we can make Github Actions catch these warnings somehow?

Yeah, I think we could add -Wpedantic to one of our builds to catch them as long as warnings are also caught as errors

PR #427, opened 2024-02-04
[TC]: Add a helper for getting implement geometry from a DDOP

ad3154:
## Describe your changes

Added a helper to easily get implement geometry from a DDOP, which would be nice to have in a TC server.

This currently provides a static function `get_implement_geometry` which returns a vector of `Boom`s. Each boom then contains descriptions of that boom's geometry like this:

- Booms
    - Sub Booms (if applicable)
        - Sections
            * X Offset
            * Y Offset
            * Z Offset
            * Width
       * X Offset
       * Y Offset
       * Z Offset
    - Sections (if applicable)
        * X Offset
        * Y Offset
        * Z Offset
        * Width
    * X Offset
    * Y Offset
    * Z Offset

Each value itself contains metadata so the user knows if the value is settable or not, and a way to check if the value is known or not (based on it being in a property versus being in a DPD for example).

You can check if values were provided by doing things like `if (booms.at(0).xOffset_mm)` for example, then check the value itself `booms.at(0).xOffset_mm.get()`

Take a look at the little unit test I added to see more.

### Questions

I would like some feedback on what specific additional info would be desirable from this helper!

For example, if we want to provide _product_ information, what's important to know about products? There are so many different kinds of rates and things that it is difficult to abstract to some extent.

Do we want these values to be settable by the users so that they can track value updates on DPDs with them?

## How has this been tested?

~~Added a unit test for this parsing based on our seeder example. I intend to also add one additional DDOP to test against before merging so that the sub-boom geometry gets tested.~~

Added three unit tests which validate all 3 "boom" configurations.

### DIFF
--- a/isobus/include/isobus/isobus/isobus_device_descriptor_object_pool_helpers.hpp
+++ b/isobus/include/isobus/isobus/isobus_device_descriptor_object_pool_helpers.hpp
@@ -1,0 +1,158 @@
+//================================================================================================
+/// @file isobus_device_descriptor_object_pool_helpers.hpp
+///
+/// @brief Defines helpers for getting commonly needed information out of a DDOP.
+/// These are provided so that you don't have to do quite as much manual parsing of
+/// the DDOP.
+/// @author Adrian Del Grosso
+///
+/// @copyright 2024 The Open-Agriculture Developers
+//================================================================================================
+
+#ifndef ISOBUS_DEVICE_DESCRIPTOR_OBJECT_POOL_HELPERS_HPP
+#define ISOBUS_DEVICE_DESCRIPTOR_OBJECT_POOL_HELPERS_HPP
+
+#include "isobus/isobus/isobus_device_descriptor_object_pool.hpp"
+#include "isobus_standard_data_description_indices.hpp"
+
+namespace isobus
+{
+	/// @brief Helper object for parsing DDOPs
+	/// @attention Getting this data from the DDOP requires traversing the
+	/// entire DDOP several times, so you should treat these as O(n^2) and try
+	/// not to call them too many times.
+	class DeviceDescriptorObjectPoolHelper
+	{
+	public:
+		/// @brief A wrapper for a DDOP value which tells you if the value
+		/// was actually supplied by the DDOP.
+		class ObjectPoolValue
+		{
+		public:
+			/// @brief Default constructor for ObjectPoolValue which
+			/// defaults the value to being non-existant and not settable
+			ObjectPoolValue() = default;
+
+			/// @brief overloads the bool operator so that you can check for
+			/// this value's existence by doing if(thisObject)
+			/// @returns true if the value was in the DDOP or has been manually set, otherwise false
+			explicit operator bool() const;
+
+			/// @brief Returns if this variable exists.
+			/// A variable exists if it was either provided in the DDOP, or has been set manually
+			/// as part of a DPD value command.
+			/// @returns true if the value has ever been set, otherwise false
+			bool exists() const;
+
+			/// @brief Returns if this value is editable. DPDs are editable. DPTs are not.
+			/// @returns true if the value can be set/edited
+			bool editable() const;
+
+			/// @brief Returns the value. If the value doesn't exist this will return 0.
+			/// @returns The value if it exists, otherwise 0.
+			std::int32_t get() const;
+
+		private:
+			friend class DeviceDescriptorObjectPoolHelper; ///< Allow our helper to change the values
+
+			std::int32_t value = 0; ///< The value being wrapped by this object
+			bool isValuePresent = false; ///< Stores if the value has ever been set
+			bool isSettable = false; ///< Stores if the value can be set, such as on a DPD's value
+		};
+
+		/// @brief A helper class that describes an individual section of a boom.
+		/// This is used to describe the sections of a boom. Units are defined in mm as specified
+		/// in the ISO 11783-10 standard. X offsets are fore/aft. Y offsets are left/right again as
+		/// defined in the ISO 11783-10 standard.
+		class Section
+		{
+		public:
+			/// @brief Default constructor for a helper class that describes an individual section of a boom
+			Section();
+
+			ObjectPoolValue xOffset_mm; ///< The x offset of the section in mm. X offsets are fore+/aft-.
+			ObjectPoolValue yOffset_mm; ///< The y offset of the section in mm. Y offsets are left-/right+.
+			ObjectPoolValue zOffset_mm; ///< The z offset of the section in mm. Z offsets are up+/down-.
+			ObjectPoolValue width_mm; ///< The width of the section in mm.
+		};
+
+		/// @brief A helper class that describes a sub boom (not all devices support this)
+		class SubBoom
+		{
+		public:
+			/// @brief Default constructor for a helper class that describes a sub boom
+			SubBoom();
+
+			std::vector<Section> sections; ///< The sections of the sub boom
+			ObjectPoolValue xOffset_mm; ///< The x offset of the sub boom in mm. X offsets are fore+/aft-.
+			ObjectPoolValue yOffset_mm; ///< The y offset of the sub boom in mm. Y offsets are left-/right+.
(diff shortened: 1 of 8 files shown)


isobus/include/isobus/isobus/isobus_device_descriptor_object_pool_helpers.hpp
```diff
@@ -0,0 +1,151 @@
...
+		/// @brief A helper class that describes a sub boom (not all devices support this)
+		class SubBoom
+		{
+		public:
+			/// @brief Default constructor for a helper class that describes a sub boom
+			SubBoom();
+
+			std::vector<Section> sections; ///< The sections of the sub boom
+			ObjectPoolValue xOffset_mm; ///< The x offset of the sub boom in mm
+			ObjectPoolValue yOffset_mm; ///< The y offset of the sub boom in mm
```
GwnDaan, line 97:
```suggestion
			ObjectPoolValue xOffset_mm; ///< The x offset of the sub boom in mm. X offsets are fore+/aft-.
			ObjectPoolValue yOffset_mm; ///< The y offset of the sub boom in mm. Y offsets are left-/right+.
```

isobus/include/isobus/isobus/isobus_device_descriptor_object_pool_helpers.hpp
```diff
@@ -0,0 +1,151 @@
...
+		/// @brief A helper class that describes a boom
+		/// This is used to describe a boom, or more generally, an ISO11783-10 function element.
+		class Boom
+		{
+		public:
+			std::vector<Section> sections; ///< The sections of the boom
+			std::vector<SubBoom> subBooms; ///< The sub booms of the boom
+			ObjectPoolValue xOffset_mm; ///< The x offset of the sub boom in mm
+			ObjectPoolValue yOffset_mm; ///< The y offset of the sub boom in mm
+			ObjectPoolValue zOffset_mm; ///< The y offset of the sub boom in mm
```
GwnDaan, line 111:
```suggestion
			ObjectPoolValue xOffset_mm; ///< The x offset of the sub boom in mm. X offsets are fore+/aft-.
			ObjectPoolValue yOffset_mm; ///< The y offset of the sub boom in mm. Y offsets are left-/right+.
			ObjectPoolValue zOffset_mm; ///< The z offset of the sub boom in mm. Z offsets are up+/down-.
```

isobus/src/isobus_device_descriptor_object_pool_helpers.cpp
```diff
@@ -0,0 +1,419 @@
...
+					{
+						// All the things we care about are likely in here, under the device element.
+						for (std::uint32_t k = 0; k < ddop.size(); k++)
+						{
+							auto potentialFunction = ddop.get_object_by_index(k);
+
+							if ((nullptr != potentialFunction) &&
+							    (task_controller_object::ObjectTypes::DeviceElement == potentialFunction->get_object_type()) &&
+							    (std::static_pointer_cast<task_controller_object::DeviceElementObject>(potentialFunction)->get_parent_object() == currentElementObject->get_object_id()) &&
+							    (task_controller_object::DeviceElementObject::Type::Function == std::static_pointer_cast<task_controller_object::DeviceElementObject>(potentialFunction)->get_type()))
```
GwnDaan, line 51: See ISO11783-10 part F.3.1, a boom can both be either a function OR a device. So I think requiring an object to be a function is not always correct?

test/tc_server_tests.cpp
```diff
@@ -1175,3 +1176,30 @@ TEST(TASK_CONTROLLER_SERVER_TESTS, MessageEncoding)
...
+
+	for (std::size_t i = 0; i < 16; i++)
+	{
+		EXPECT_TRUE(implement.booms.at(0).sections.at(i).width_mm);
+		EXPECT_TRUE(implement.booms.at(0).sections.at(i).xOffset_mm);
+		EXPECT_TRUE(implement.booms.at(0).sections.at(i).yOffset_mm);
+		EXPECT_FALSE(implement.booms.at(0).sections.at(i).zOffset_mm);
+
+		EXPECT_NE(0, implement.booms.at(0).sections.at(i).width_mm.get());
+		EXPECT_NE(0, implement.booms.at(0).sections.at(i).yOffset_mm.get());
```
GwnDaan, line 1203: suggestion(non-blocking): probably better to match the actual expected values here rather than an arbitrary number not equal to 0

GwnDaan (commented):
Good job! I really like helpers like this. Simple and straight to the point and not that intertwined with other classes 😄

isobus/src/isobus_device_descriptor_object_pool_helpers.cpp
```diff
@@ -0,0 +1,419 @@
...
+		Implement retVal;
+
+		if (0 == ddop.size())
+		{
+			CANStackLogger::error("[DDOP Helper]: No objects in the pool.");
+			return retVal; // Return empty object
+		}
+
+		// First, find the device object
+		for (std::uint32_t i = 0; i < ddop.size(); i++)
```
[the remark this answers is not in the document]
ad3154: It is true that there can only be 65535 objects in the DDOP (though that would be a very ridiculous number of objects). I'd be OK with casting it down in the DDOP object instead.

isobus/src/isobus_device_descriptor_object_pool_helpers.cpp
ad3154: Without a device, object the DDOP wouldn't be very functional, so it maybe makes sense to check for a function first, then if none is found, consider the device to be the root of the boom?

ad3154: Added this functionality and added a unit test specific to this case.

isobus/src/isobus_device_descriptor_object_pool_helpers.cpp
```diff
@@ -0,0 +1,278 @@
...
+	DeviceDescriptorObjectPoolHelper::SubBoom DeviceDescriptorObjectPoolHelper::parse_sub_boom(DeviceDescriptorObjectPool &ddop,
+	                                                                                           std::shared_ptr<task_controller_object::DeviceElementObject> elementObject)
+	{
+		SubBoom retVal;
+
+		// Find all sections in this sub boom
+		// We again have to search the whole pool because elements have parent links, not child links
+		for (std::uint16_t i = 0; i < ddop.size(); i++)
+		{
+			auto section = ddop.get_object_by_index(static_cast<std::uint16_t>(i));
```
GwnDaan, line 219: nit(non-blocking): redundant cast

GwnDaan (approved):

GwnDaan:
> For example, if we want to provide product information, what's important to know about products? There are so many different kinds of rates and things that it is difficult to abstract to some extent.

I'm unsure what specific product information you are targeting here. But I think making a helper class for abstracting the rate control can be a good start here. Implementing the setpoint, actual, default, minimum and maximum application rate for all the common yet [different types](https://www.isobus.net/isobus/dDEntity/index?DDEntity%5BDDIdentifier%5D=&DDEntity%5BDeviceClassID%5D=&DDEntity%5BShowRequests%5D=0&DDEntity%5BsysUnitID%5D=&DDEntity%5BShowAttachmentsOnly%5D=0&DDEntity%5BDefinition%5D=application+rate). Then based on like an enum or something we can differentiate them. Ideally we have it automatically track the state by listening to changes to the respective DDEntities, similar to the VT tracker helper. And then also for the other way around, we can have some setter functions that abstracts the setting of these application rates, similar to the VT update helper... Just thinking out loud here

ad3154:
I think I'll merge for now, and follow-up to fix your nits, as well as to add some extra data for sections, like I think it would be nice to also return the element numbers of each section so a TC knows quickly what section == a specific element to command. Reduces the need for further O(n^2) searches.

As for products, it's just very complicated, because you can have a boom level rate, for example, with sub-rates in each sub-boom, with sub-rates in each section, and with booms having different kinds of rates, like count/area, volume/area, mass/area etc, so abstracting it seems like it will be tough to represent well.. I will experiment around...

PR #436, opened 2024-02-18
[Docs]: add standardized alarm ack handling to examples

### LINKED ISSUE #424: Listen to alarm ACK messages in example
When an alarm mask is activated, most of the time the VT will present an extra button for closing it. (e.g. the VT proprietary acknowledge button). You expect the alarm to be acknowledged when you press it. However, currently we do not listen for it in our examples. Could be a nice quality of life thing to add.

edit(31-01): it seems to be the Soft Key Activation message with `soft key code = 0` that is used for such a message

GwnDaan:
## Describe your changes

- Added the standardized way of closing alarm masks with the VT proprietary `0` key number.
- Updated tutorials to match latest changes in the stack

Closes #424 

## How has this been tested?

Ran both examples, one with SocketCAN and the other with platformio on ESP32-S3. 
For the tutorials, I followed them and at the same time made the necessary adjustments. I also removed a bit of the ESP-IDF setup since PlatformIO will handle that for you. The ESP32 tutorial I tested on a fresh dev container, to avoid dependencies being already installed.


ad3154:
Hmm.. I am having some troubles building the docs recently which I don't really understand...

```
writing output... [  3%] Concepts
Extension error (sphinxext.opengraph):
Handler <function html_page_context at 0x7f48d1d37ca0> for event 'html-page-context' threw an exception (exception: First argument must be a path or binary-mode file object)
```

Does this mean anything to you? Either way, could you add a version/branch build of this on readthedocs so I can check it out?

GwnDaan:
Hmm, that doesn't ring a bell for me. I'd suggest doing a clean build (i.e. removing all build folders first), but you probably already tried that. Also make sure to build doxygen before sphinx, although that still should give a different error compared to yours. Lastly, you could try removing some python packages with pip, it seems it has trouble with `sphinxext.opengraph` which I don't think we actively use?

Either way, here is a link to the readthedocs documentation for this PR: https://agisostack-plus-plus.readthedocs.io/en/daan-add-esc-button/

ad3154 (approved):
Added a 1 line PR to this branch you can merge in if you want #438 but looks good!

PR #443, opened 2024-02-28
[TECU]: Add ISOBUS Heartbeat Message

ad3154:
## Describe your changes

Added an interface which manages sending and receiving the ISOBUS heartbeat.

The heartbeat message is used to determine the integrity of the communication of messages and parameters being transmitted by a control function. There may be multiple instances of the heartbeat message on the network, and CFs are required transmit the message on request. As long as the heartbeat message is transmitted at the regular time interval and the sequence number increases through the valid range, then the heartbeat message indicates that the data source CF is operational and provides correct data in all its messages.

This interface is enabled by default, but can be disabled if you want to stop your heartbeat(s) or don't care about the safety-critical path of the machine.

https://www.isobus.net/isobus/pGNAndSPN/2507?type=PGN

## How has this been tested?

- [X] Added unit test for heartbeat interface
- [x] Tested with seeder example

### DIFF
--- a/isobus/include/isobus/isobus/isobus_heartbeat.hpp
+++ b/isobus/include/isobus/isobus/isobus_heartbeat.hpp
@@ -1,0 +1,154 @@
+//================================================================================================
+/// @file isobus_heartbeat.hpp
+///
+/// @brief Defines an interface for sending and receiving ISOBUS heartbeats.
+/// The heartbeat message is used to determine the integrity of the communication of messages and
+/// parameters being transmitted by a control function. There may be multiple instances of the
+/// heartbeat message on the network, and CFs are required transmit the message on request.
+/// As long as the heartbeat message is transmitted at the regular
+/// time interval and the sequence number increases through the valid range, then the
+/// heartbeat message indicates that the data source CF is operational and provides
+/// correct data in all its messages.
+///
+/// @author Adrian Del Grosso
+///
+/// @copyright 2024 The Open-Agriculture Developers
+//================================================================================================
+#ifndef ISOBUS_HEARTBEAT_HPP
+#define ISOBUS_HEARTBEAT_HPP
+
+#include "isobus/isobus/can_callbacks.hpp"
+#include "isobus/isobus/can_internal_control_function.hpp"
+#include "isobus/isobus/can_message.hpp"
+#include "isobus/utility/event_dispatcher.hpp"
+
+#include <list>
+
+namespace isobus
+{
+	/// @brief This class is used to send and receive ISOBUS heartbeats.
+	class HeartbeatInterface
+	{
+	public:
+		/// @brief This enum is used to define the possible errors that can occur when receiving a heartbeat.
+		enum class HeartBeatError
+		{
+			InvalidSequenceCounter, ///< The sequence counter is not valid
+			TimedOut ///< The heartbeat message has not been received within the repetition rate
+		};
+
+		/// @brief Constructor for a HeartbeatInterface
+		/// @param[in] sendCANFrameCallback A callback used to send CAN frames
+		HeartbeatInterface(const CANMessageFrameCallback &sendCANFrameCallback);
+
+		/// @brief This can be used to disable or enable this heartbeat functionality.
+		/// It's probably best to leave it enabled for most applications, but it's not
+		/// strictly needed.
+		/// @note The interface is enabled by default.
+		/// @param[in] enable Set to true to enable the interface, or false to disable it.
+		void set_enabled(bool enable);
+
+		/// @brief Returns if the interface is currently enabled or not.
+		/// @note The interface is enabled by default.
+		/// @returns true if the interface is enabled, false if the interface is disabled
+		bool is_enabled() const;
+
+		/// @brief This method can be used to request that another control function on the bus
+		/// start sending the heartbeat message. This does not mean the request will be honored.
+		/// In order to know if your request was accepted, you will need to either
+		/// register for timeout events, register for heartbeat events, or check to see if your
+		/// destination control function ever responded at some later time using the various methods
+		/// available to you on this class' public interface.
+		/// @note CFs may take up to 250ms to begin sending the heartbeat.
+		/// @param[in] sourceControlFunction The internal control function to use when sending the request
+		/// @param[in] destinationControlFunction The destination for the request
+		/// @returns true if the request was transmitted, otherwise false.
+		bool request_heartbeat(std::shared_ptr<InternalControlFunction> sourceControlFunction,
+		                       std::shared_ptr<ControlFunction> destinationControlFunction) const;
+
+		/// @brief Called by the internal control function class when a new internal control function is added.
+		/// This allows us to respond to requests for heartbeats from other control functions.
+		/// @param[in] newControlFunction The new internal control function
+		void on_new_internal_control_function(std::shared_ptr<InternalControlFunction> newControlFunction);
+
+		/// @brief Called when an internal control function is deleted. Cleans up stale registrations with
+		/// PGN request protocol.
+		/// @param[in] destroyedControlFunction The destroyed internal control function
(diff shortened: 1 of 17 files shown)


isobus/src/isobus_heartbeat.cpp
```diff
@@ -0,0 +1,287 @@
...
+/// correct data in all its messages
+///
+/// @author Adrian Del Grosso
+///
+/// @copyright 2024 The Open-Agriculture Developers
+//================================================================================================
+#include "isobus/isobus/isobus_heartbeat.hpp"
+
+#include "isobus/isobus/can_general_parameter_group_numbers.hpp"
+#include "isobus/isobus/can_network_manager.hpp"
```
GwnDaan, line 20: thought: I think we have a circular dependency currently, correct me if I'm wrong. Maintaining (and testing) this interface should be easier if we can remove it. I'll leave a few more comments on how I would approach it

isobus/src/isobus_heartbeat.cpp
```diff
@@ -0,0 +1,287 @@
...
+
+	void HeartbeatInterface::set_enabled(bool enable)
+	{
+		if ((!enable) && (enable != enabled))
+		{
+			LOG_DEBUG("[HB]: Disabling ISOBUS heartbeat interface.");
+			disable();
+		}
+		enabled = enable;
+	}
```
GwnDaan, line 60: resolving circular-dependency: remove call to `disable()` here, and instead use an early return of some kind in the `update()` and `process_request_for_heartbeat()`. Although not ideal, it simplifies the states disabling/enabling the interface and helps us remove the circular dependency
```c++
if (!enabled) {
    return;
}
```

isobus/src/isobus_heartbeat.cpp
```diff
@@ -0,0 +1,287 @@
...
+			for (auto &cf : internalControlFunctions)
+			{
+				auto pgnRequestProtocol = cf->get_pgn_request_protocol().lock();
+
+				if (nullptr != pgnRequestProtocol)
+				{
+					pgnRequestProtocol->register_request_for_repetition_rate_callback(static_cast<std::uint32_t>(CANLibParameterGroupNumber::HeartbeatMessage), process_request_for_heartbeat, this);
+				}
+			}
+			CANNetworkManager::CANNetwork.add_any_control_function_parameter_group_number_callback(static_cast<std::uint32_t>(CANLibParameterGroupNumber::HeartbeatMessage), process_rx_message, this);
```
GwnDaan, line 42: resolving circular-dependency: Instead of adding a "any" callback for this PGN, we can call `process_rx_message()` directly from the CANNetworkManager

isobus/src/isobus_heartbeat.cpp
```diff
@@ -0,0 +1,287 @@
...
+			initialized = true;
+		}
+	}
```
GwnDaan, line 45: thought: theoretically there can't be any internal control function that `on_new_internal_control_function` doesn't catch? If so then this initialize function can be removed

isobus/src/isobus_heartbeat.cpp
```diff
@@ -0,0 +1,287 @@
...
+	  timestamp_ms(SystemTiming::get_timestamp_ms())
+	{
+	}
+
+	bool HeartbeatInterface::Heartbeat::send()
+	{
+		bool retVal = false;
+		const std::array<std::uint8_t, 1> buffer = { sequenceCounter };
+
+		retVal = CANNetworkManager::CANNetwork.send_can_message(static_cast<std::uint32_t>(CANLibParameterGroupNumber::HeartbeatMessage),
```
GwnDaan, line 160: resolving circular-dependency: this is the interesting part, to make this interface completely free from the network manager then we probably need a callback for sending messages like in the recently refactored transport protocols? I still don't really like that approach but I haven't found anything nicer yet

isobus/src/isobus_heartbeat.cpp
```diff
@@ -0,0 +1,287 @@
...
+				{
+					LOG_ERROR("[HB]: Invalid sequence counter received in heartbeat.");
+					interface->heartbeatErrorEventDispatcher.call(HeartBeatError::InvalidSequenceCounter, message.get_source_control_function());
+				}
+				else if ((250 == managedHeartbeat->sequenceCounter) &&
+				         (0 != message.get_uint8_at(0)))
+				{
+					LOG_ERROR("[HB]: Invalid sequence counter received in heartbeat.");
+					interface->heartbeatErrorEventDispatcher.call(HeartBeatError::InvalidSequenceCounter, message.get_source_control_function());
+				}
```
GwnDaan, line 208:
```suggestion
				else if (message.get_uint8_at(0) != ((managedHeartbeat->sequenceCounter + 1) % 250))
				{
					LOG_ERROR("[HB]: Invalid sequence counter received in heartbeat.");
					interface->heartbeatErrorEventDispatcher.call(HeartBeatError::InvalidSequenceCounter, message.get_source_control_function());
				}
```

GwnDaan (changes requested):
Some initial feedback, I haven't been able to test it yet but it looks promising

isobus/include/isobus/isobus/isobus_heartbeat.hpp
```diff
@@ -0,0 +1,158 @@
...
+		/// This allows us to respond to requests for heartbeats from other control functions.
+		/// @param[in] newControlFunction The new internal control function
+		void on_new_internal_control_function(std::shared_ptr<InternalControlFunction> newControlFunction);
+
+		/// @brief Returns an event dispatcher which can be used to register for heartbeat errors.
+		/// Heartbeat errors are generated when a heartbeat message is not received within the
+		/// repetition rate, or when the sequence counter is not valid.
+		/// The control function that generated the error is passed as an argument to the event.
+		/// @returns An event dispatcher for heartbeat errors
+		EventDispatcher<HeartBeatError, std::shared_ptr<ControlFunction>> &GetHeartbeatErrorEventDispatcher();
```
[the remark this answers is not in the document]
ad3154: My work habits rubbing off on me... at work we use camel pascal case....

isobus/src/isobus_heartbeat.cpp
ad3154: Yeah, I dunno if I super love how we're doing the callback thing with those, but I agree I don't have an alterative to propose, so for the time being we might as well keep it consistent and do something similar...

isobus/src/isobus_heartbeat.cpp
ad3154: I prefer adding to the top of the enclosing `if` checks personally, which accomplishes the same thing basically. Added where needed.

isobus/src/isobus_heartbeat.cpp
ad3154: I think so? Though this reminds me that I need to implement the reverse, so when an internal control function is deleted I should clean up these registrations probably.

isobus/src/isobus_heartbeat.cpp
ad3154: Circular dependency now removed

ad3154:
> Some initial feedback, I haven't been able to test it yet but it looks promising

I think I addressed all your comments/suggestions in the latest diff, and I tested with the seeder example and it appeared to start sending a compliant heartbeat on request, which is great! The counter appeared to start at 251, roll over to 0, and reset once 250 was crossed as expected.

I also fixed the seeder example's manufacturer code (was 64) to be our code.

isobus/src/isobus_heartbeat.cpp
GwnDaan: Seems you forgot to implement `sendCANFrameCallback` here?

GwnDaan:
Sounds good, I re-opened one of them, but other than that looks good! I'm thinking it might be

isobus/src/isobus_heartbeat.cpp
ad3154: Quite right, fixed.

GwnDaan (approved):
Awesome!

PR #444, opened 2024-03-02
[TC]: Add product abstraction to DDOP helper

ad3154:
## Describe your changes

This provides a way to get rate/product information from a DDOP. Generally it considers any `Bin` to be the root of a product, which seems to be fairly standard.

This is intended to be used in a TC server to quickly learn about what products the implement might have.

The information provided includes data about:

- Default rate
- Setpoint rate
- Actual rate
- Minimum rate
- Maximum rate

## How has this been tested?

Expanded TC server unit tests to include parsing for product information.


GwnDaan (approved):

ad3154:
Hmmmmm... the more I think about it, the more I think some of these things need their parent element number included in the returned data, that way the TC knows what element to use when commanding or requesting the value without needing to search the whole DDOP again. I will merge this, but I think I will follow up with another PR that expands the data returned

PR #448, opened 2024-03-13
[TC]: Add options for falling back to other language data sources when TC doesn't respond

ad3154:
## Describe your changes

This change alters how we deal with the language command PGN in the TC client to be more reliable and more tolerant of TCs which don't fully comply with version 4 of the standard when they claim they're version 4.

- Allow the TC client to fall back to other language sources as needed.
- Fixed a condition where the TC client would become stuck while waiting for a language response if a response never arrived. 
- Added a getter to the language command interface to allow retrieving the assigned partner CF. 
- Prevent a crash in the language command interface which could occur if you send the command with some fields left unconfigured.
- Removed TC client's dependency on the VT client.

## How has this been tested?

~~This is pretty challenging to test in a unit test without a large amount of boiler plate code, so it has minimal testing at the moment.~~ Ideally it would be nice to try this against a TC which doesn't send the language command. ~~I would like to get some kind of automated test added though, so I've made this a draft until I can cover at least the VT fallback scenario.~~

Added unit test coverage of the fallback scenario and the 6 second complete timeout scenario.

### DIFF
--- a/isobus/include/isobus/isobus/isobus_language_command_interface.hpp
+++ b/isobus/include/isobus/isobus/isobus_language_command_interface.hpp
@@ -166,6 +166,10 @@ namespace isobus
 		/// @param[in] filteredControlFunction The new partner to communicate with
 		void set_partner(std::shared_ptr<PartneredControlFunction> filteredControlFunction);
 
+		/// @brief Returns the current partner being used by the interface
+		/// @return The current partner being used by the interface, or nullptr if none
+		std::shared_ptr<PartneredControlFunction> get_partner() const;
+
 		/// @brief Returns if initialize has been called yet
 		/// @return `true` if initialize has been called, otherwise false
 		bool get_initialized() const;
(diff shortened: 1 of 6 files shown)


GwnDaan (approved):
LGTM. I like how you separated the "languageCommandWaitingTimestamp_ms" and the "stateMachineTimestamp_ms" then there is no way of getting a livelock.

One thing to note, when we fallback onto the VT language interface, we don't store the result in the TCClient's languageCommandInterface. Might lead to inconsistencies in the application. 

One thing I see how we could combat it; just set the partner of the languageCommandInterface of the TCClient to the VT control function, instead of doing requests via the commandInterface of the VTClient 

EDIT: that could also remove the dependency on the VT, since it was only being used for the languageInterface inside the TC. Now we can just pass the partneredControlFunction of the VT to the TC client?

ad3154:
> EDIT: that could also remove the dependency on the VT, since it was only being used for the languageInterface inside the TC. Now we can just pass the partneredControlFunction of the VT to the TC client?

Oh hey, that sounds great!

I'll change that. Removing the dependency sounds really nice, I never really liked how that was....

GwnDaan (approved):

PR #451, opened 2024-03-15
[HW] Combine single- and multithreaded hardware interfaces

GwnDaan:
## Describe your changes

Attempted to combine the hardware interfaces for the single-threaded and multi-threaded variants into one class. For this too be somewhat easier, I switched the messages queues to a ring-buffer which can have a lockfree implementation. For the single threaded variant I just use the standard queue as it doesn't have to worry about concurrent access.

I also did some more moving around, especially moving some of the channel specific functionality stuff to its struct (now class).

## How has this been tested?

- Ran a slightly modified version of the VT3 pool example to test single-threading. 
- Ran the seeder example for multi-threaded testing
- Made sure unit-tests are still speed-wise the same, so I hopefully didn't break the threading stuff


utility/include/isobus/utility/thread_synchronization.hpp
```diff
@@ -38,6 +106,95 @@ namespace isobus
...
+private:
+	std::vector<T> buffer; ///< The buffer for the circular buffer.
+	std::atomic<size_t> readIndex = { 0 }; ///< The read index for the circular buffer.
+	std::atomic<size_t> writeIndex = { 0 }; ///< The write index for the circular buffer.
+	const size_t capacity; ///< The capacity of the circular buffer.
+
+	/// @brief Get the next index in the circular buffer.
+	/// @param current The current index.
+	/// @return The next index in the circular buffer.
+	size_t nextIndex(size_t current) const
```
ad3154, line 192: There's some size_t's in here that look like they need `std::`

utility/include/isobus/utility/thread_synchronization.hpp
```diff
@@ -38,6 +106,95 @@ namespace isobus
...
 
+/// @brief A template class for a lock free queue.
+/// @tparam T The item type for the queue.
+template<typename T>
+class LockFreeQueue
+{
+public:
+	/// @brief Constructor for the lock free queue.
+	explicit LockFreeQueue(std::size_t size) :
+	  buffer(size), capacity(size) {}
```
ad3154, line 117: Might sound a little insane, but we might want to zero-check the size... I see the possibility of `% 0` later and my UB alarms start going off

utility/include/isobus/utility/thread_synchronization.hpp
```diff
@@ -25,9 +26,76 @@ namespace isobus
...
+	{
+		return false;
+	}
+
+	/// @brief Clear the queue.
+	void clear()
+	{
+		while (!queue.empty())
+		{
+			queue.pop();
```
ad3154, line 86: I think this might actually be more efficient, though it probably doesn't matter

`queue = { };`

hardware_integration/src/can_hardware_interface.cpp
```diff
@@ -16,48 +17,160 @@
...
+		LOCK_GUARD(Mutex, hardwareChannelsMutex);
 
-		if (threadsStarted)
+		if (started)
 		{
 			LOG_ERROR("[HardwareInterface] Cannot set number of channels after interface is started.");
 			return false;
 		}
 
 		while (value > hardwareChannels.size())
```
ad3154, line 171: While we're in this file we might want to clean up these implicit casts between uint8_t and size_t

utility/include/isobus/utility/thread_synchronization.hpp
```diff
@@ -38,6 +106,95 @@ namespace isobus
 /// @param x The mutex to lock.
```
ad3154, line 106: Nit: extra unneeded `;` on line 130. It's not part of the diff but it makes a lot of angry warnings with pedantic warnings enabled

hardware_integration/include/isobus/hardware_integration/can_hardware_interface.hpp
```diff
@@ -84,88 +83,114 @@ namespace isobus
...
+			void start_threads();
+
+			/// @brief Stops the receiving thread for this CAN channel
+			void stop_threads();
+
+			/// @brief The receiving thread loop for this CAN channel
+			void receive_thread_function();
 
 			std::unique_ptr<std::thread> receiveMessageThread; ///< Thread to manage getting messages from a CAN channel
+			bool receiveThreadRunning; ///< Flag to indicate if the receive thread is running
```
ad3154, line 152: MSVC complains that this is not always initialized. Might want a `= false` on there

utility/include/isobus/utility/thread_synchronization.hpp
```diff
@@ -38,6 +106,95 @@ namespace isobus
...
 #define LOCK_GUARD(type, x) const std::lock_guard<type> x##Lock(x)
 
+/// @brief A template class for a lock free queue.
+/// @tparam T The item type for the queue.
+template<typename T>
+class LockFreeQueue
```
ad3154, line 112: Nice, I bet this speeds things up a bit

hardware_integration/src/can_hardware_interface.cpp
```diff
@@ -259,126 +350,89 @@ namespace isobus
...
 			{
 				// Stage 1 - Receiving messages from hardware
-				channelsLock.lock();
-				std::for_each(hardwareChannels.begin(), hardwareChannels.end(), [](const std::unique_ptr<CANHardware> &channel) {
-					std::lock_guard<std::mutex> lock(channel->receivedMessagesMutex);
-					while (!channel->receivedMessages.empty())
-					{
-						const auto &frame = channel->receivedMessages.front();
+				LOCK_GUARD(Mutex, hardwareChannelsMutex);
+				for (uint8_t i = 0; i < hardwareChannels.size(); i++)
```
ad3154, line 360: Nit: Missing `std::`, implicit cast

ad3154 (changes requested):
Looking good, I like reducing the duplication. I'd like to also run some tests myself but here's at least some initial thoughts

GwnDaan:
I pushed a separate commit for the small improvements, should make it easier to review what changes I made after the initial commit. Though it might not look clean on the main branch commit graph, so we should probably stick with for "squash and merge" for this PR :). I'm thinking it might also be more convenient for reviewing if we do it on some of the other (larger) PR's, hmm?...

ad3154 (approved):
Finally got to try this out, and it seems to be working pretty well from my testing.

Only one last note, the VT example didn't compile for me in single threaded mode on MSVC until I added `#include <string>` to `can_control_function.hpp` which seems like an include order issue that has existed for a while without notice? 

I'll open a quick PR for that since it's unrelated to your diff haha.

Thanks for your work on this!

ad3154:
> we should probably stick with for "squash and merge" for this PR :). I'm thinking it might also be more convenient for reviewing if we do it on some of the other (larger) PR's

Yep, that's also fine with me

PR #455, opened 2024-03-20
[HW]: Add Peak Ethernet Gateway DR support

ad3154:
## Describe your changes

Added a CAN hardware plugin for the [PEAK Ethernet Gateway](https://www.peak-system.com/PCAN-Ethernet-Gateway-DR.330.0.html?&L=1).

Works both on Linux and Windows so far!

This does require a small bit of config on the gateway itself. You need to create routes to your computer by IP address, and configure this setting:
image

The code is very much based on the C example from Peak, which was a little cursed admittedly/

## How has this been tested?

Ran VT example from a PC on my local network, connecting to the gateway. Object pool loaded and behaved normally.


GwnDaan:
Cool! Is this some arbitrary protocol to send CAN messages over ethernet or is this in fact also the same as the HSI protocol? 

Also have a look at the functions naming, I think we mostly used "snake_case" in this repo

ad3154:
> I think we mostly used "snake_case" in this repo

I blame work.... I spend 8 hours a day typing `PascalCase` 😅

> or is this in fact also the same as the HSI protocol?

My understanding is that HSI is still undecided on of they are going to do DDS, SOME/IP, CAN tunneling, or some other thing... so there may not be an official schema yet. I imagine this would be close-ish to whatever they were thinking for CAN tunneling, but I have no idea haha. Seems likely it's some bespoke PEAK schema.

hardware_integration/CMakeLists.txt
```diff
@@ -127,8 +137,9 @@ add_library(${PROJECT_NAME}::HardwareIntegration ALIAS HardwareIntegration)
 
 target_compile_features(HardwareIntegration PUBLIC cxx_std_11)
 set_target_properties(HardwareIntegration PROPERTIES CXX_EXTENSIONS OFF)
-target_link_libraries(HardwareIntegration PRIVATE ${PROJECT_NAME}::Utility
-                                                  ${PROJECT_NAME}::Isobus)
+target_link_libraries(
+  HardwareIntegration PRIVATE ${PROJECT_NAME}::Utility ${PROJECT_NAME}::Isobus
+                              ${SOCKET_LIB})
```
GwnDaan, line 142: thought: can this SOCKET_LIB be harmful if a platform doesn't have it?

hardware_integration/src/peak_ethernet_gateway_plugin.cpp
```diff
@@ -0,0 +1,388 @@
...
+
+			if (0 != rxSocket)
+			{
+				if (rxOpened)
+				{
+					bool txOpened = open_tx_socket();
+
+					if (txOpened)
+					{
+						rxBuffer.reset(new std::uint8_t[RX_BUFFER_SIZE_BYTES]());
```
GwnDaan, line 95: Hahah mixing a little old C stuff with C++'s smart pointers, it does it's job I guess 😆

GwnDaan (approved):
I currently don't have the hardware for testing this myself, but if you found it working I'm all good with it

PR #458, opened 2024-03-28
[Core]: Add support for CMSIS RTOS 2 Thread synchronization

ad3154:
## Describe your changes

This is an attempt to provide support for embedded RTOSes that use the CMSIS RTOS 2 abstraction, such as FreeRTOS, Keil RTX, and others.

My thinking is that this is 50 of the battle to supporting STM32 and other small ARM devices, with the other half being hardware plugins with specific HAL integrations for each micro.

In order to actually consume this, you'd need to also implement a proper hardware time-base for the stack using a hardware timer on those devices, which maybe we could also provide as part of a CAN plugin eventually. Additionally, you'd also need to start the RTOS before using the stack, which I think is acceptable (osKernelInitialize, and osKernelStart).

Either way, the point is, CMSIS RTOS2 is a common RTOS abstraction which would be nice to support.

The CMSIS RTOS 2 API is documented [here](https://www.keil.com/pack/doc/CMSIS/RTOS2/html/group__CMSIS__RTOS.html)

I added some super obvious infinite loops to catch issues between us and the kernel. I used loops because assert may not be supported on the target platforms, and since this file is included in the logger, I can't emit any log output here.

## How has this been tested?

This is a draft until I can test it. I have a cheap STM32L4 eval board coming to help test it. Will update once I have more to share.


ad3154:
Marking as ready since it's now tested. It's a bit more invasive than I expected I guess, but latching onto the CMSIS abstraction for RTOSes seems very valuable, as it should give us very easy access to:

- FreeRTOS
- Zephyr
- RTX
- ThreadX
- Azure RTOS
- Others?

utility/include/isobus/utility/thread_synchronization.hpp
```diff
@@ -88,16 +98,368 @@ class LockFreeQueue
...
+#include <atomic>
+#include <cstdint>
+#include <functional>
+#include <vector>
+
+namespace isobus
+{
+	/// @brief A wrapper around a CMSIS RTOS 2 mutex.
+	/// @details See definition at https://www.keil.com/pack/doc/CMSIS/RTOS2/html/group__CMSIS__RTOS.html
+	class Mutex
```
GwnDaan, line 114: thought: Mutex is not a template class, so we should separate the implementation and definition?

utility/include/isobus/utility/thread_synchronization.hpp
```diff
@@ -88,16 +98,368 @@ class LockFreeQueue
...
+	using recursive_mutex = isobus::RecursiveMutex;
+	using condition_variable = isobus::ConditionVariable;
+} // namespace std
+
+#define LOCK_GUARD(type, x) const LockGuard<type> x##Lock(&x)
+
+/// @brief A template class for a lock free queue.
+/// @tparam T The item type for the queue.
+template<typename T>
+class LockFreeQueue
```
GwnDaan, line 357: nit: LockFreeQueue is already in the file below. I think it makes sense if we just use a separate `#ifdef` directive there that covers both cases

utility/include/isobus/utility/thread_synchronization.hpp
```diff
@@ -88,16 +98,368 @@ class LockFreeQueue
...
+				if (osOK != osRetVal)
+				{
+					while (true)
+					{
+						// If your code is stuck in here, that means you did something
+						// very wrong, like recursively locked this mutex, or tried to
+						// lock the mutex before the OS was initialized, or called this in
+						// an interrupt service routine.
+						// osRetVal may contain more information.
+					}
```
GwnDaan, line 139: I like the way you used these for loops. It's a clever trick to help with finding where things go wrong with debugging. Though I wonder if it's always the way to go in production? I'm not sure

GwnDaan (changes requested):

PR #460, opened 2024-03-31
[TC]: Allow disabling the data dictionary to save on binary/RAM size

ad3154:
## Describe your changes

This allows you to disable the ISOBUS data dictionary if you want to save a bunch of ROM/RAM space in your binary.
This was very helpful on the STM32 platform I'm testing on.

## How has this been tested?

Unit tests which use the data dictionary interface still compile with the database disabled.

### DIFF
--- a/isobus/CMakeLists.txt
+++ b/isobus/CMakeLists.txt
@@ -100,6 +100,9 @@ prepend(ISOBUS_INCLUDE ${ISOBUS_INCLUDE_DIR} ${ISOBUS_INCLUDE})
 option(DISABLE_CAN_STACK_LOGGER
        "Compiles out all logging to minimize binary size" OFF)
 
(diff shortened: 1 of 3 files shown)


GwnDaan (approved):
Looks good, as a side note I'd say we should look into how we can generalize configuration across multiple platforms now we are expanding there. E.g. ESP-IDF doesn't use cmake, nor does Arduino I think? Ultimately there would be just one file where we specify what can be configured, but yeah I doubt doubt that is possible 🤔

ad3154:
Yes I agree, we should explore centralizing our config in a way that's more cross-platform friendly.

My initial thought is to explore one main config file with `#define`s that CMake edits on CMake builds, or can be manually edited otherwise, but needs some thought. I will create an issue for it.

PR #462, opened 2024-04-07
Daan/detached message handling

GwnDaan:
## Describe your changes

Introduced 3 classes that will remove the the dependency on the CANNetworkManager for all classes that want to receive / send messages. namely:

- **CANMessagingProvider**: a simple interface that provides the "send can message" function, currently this will always be the CANNetworkManager, but it also allows us to exchange it easily, e.g. if we want some unit test without the network manager.
- **CANMessagingConsumer**: a base class to extend from. It provides the send message function and allows for handling tx/rx messages. I.e. any class that wants to interact with can messages can extend from this one and have everything ready.
- **CANMessageHandler**: connects the CANMessagingProvider with CANMessagingConsumer's. And it's "process tx/rx message" function can be called to automatically distribute a particular message across consumers.

I think this will be a better alternative than passing the network manager around everywhere when we are refactoring away from the singleton. And it might especially be useful for unit testing, where we can isolate a class we wish to test and thereby remove interference from others. Though I'd like to hear any thoughts on this, as it does surely add an extra layer off complexity to the stack.

I see 3 ways a consumer can be set-up;
- We can pas the `CANMessageHandler` to the consumer class and let it register itself, i.e.:
```c++
void initialize(CANMessageHandler messageHandler)
{
	// This class must extends enabled_shared_from_this, unlikely and unintuitive...
	messageHandler.add_consumer(std::shared_from_this());

	// If we decide to allow raw-pointers, note that also
	// introduces the risk of forgetting the removal on cleanup
	messageHandler.add_consumer(this);
}
```
- We add an extra function to the CANNetworkManager that sets up the messaging for the passed in consumer:
```c++
void CANNetworkManager::initialize_messaging(std::shared_ptr<CANMessagingConsumer> consumer)
{
	messageHandler.add_consumer(consumer);
}
```
- Since the above is really simple, it might instead make more sense to just let the end user/application call that function of the message handler, i.e.:
```c++
auto interfaceUnderTest = std::make_shared<ShortcutButtonInterface>(internalECU, true);
CANNetworkManager::CANNetwork.get_can_message_handler().add_consumer(interfaceUnderTest);
```

I think I prefer the last one. To start off, I switched over the ShortcutButtonInterface and HeartbeatInterface to this messaging system so we can see what it's like.

## What's next

Once we find and agree on an approach, all the other classes that send/receive messages have to be adopted to this method.
Furthermore, I'd like some unit-tests that will be run first in line to make sure the system remains correctly in place and it doesn't cause other tests to fail.

### DIFF
--- a/isobus/src/can_network_manager.cpp
+++ b/isobus/src/can_network_manager.cpp
@@ -40,8 +40,14 @@ namespace isobus
 			get_next_can_message_from_tx_queue();
 		}
 
-		//! @todo: Remove this unsafe way of making a shared pointer to the network manager once the network manager is no longer a singleton
-		messageHandler.set_messaging_provider(std::shared_ptr<CANNetworkManager>(this));
+		//! @todo: Remove this unsafe way of making a shared pointer to the network manager once the network manager is no longer a singleton,
+		//! it's the cause of a double-free right now, but since it must be at the end of an application since the CANNetwork is a sngleton, it shouldn't harm too much for now
+		messagingProvider = std::shared_ptr<CANNetworkManager>(this);
+		messageHandler.set_messaging_provider(messagingProvider);
+		for (std::uint8_t i = 0; i < CAN_PORT_MAXIMUM; i++)
+		{
+			messageHandler.add_consumer(heartBeatInterfaces.at(i));
+		}
 		initialized = true;
 	}
 
@@ -506,6 +512,7 @@ namespace isobus
 		currentBusloadBitAccumulator.fill(0);
 		lastAddressClaimRequestTimestamp_ms.fill(0);
 		controlFunctionTable.fill({ nullptr });
+		heartBeatInterfaces.fill({ std::make_shared<HeartbeatInterface>() });
 
 		auto send_frame_callback = [this](std::uint32_t parameterGroupNumber,
 		                                  CANDataSpan data,
@@ -528,7 +535,6 @@ namespace isobus
 			transportProtocols.at(i).reset(new TransportProtocolManager(send_frame_callback, receive_message_callback, &configuration));
 			extendedTransportProtocols.at(i).reset(new ExtendedTransportProtocolManager(send_frame_callback, receive_message_callback, &configuration));
 			fastPacketProtocol.at(i).reset(new FastPacketProtocol(send_frame_callback));
-			heartBeatInterfaces.at(i).reset(new HeartbeatInterface(send_frame_callback));
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


isobus/src/can_network_manager.cpp
```diff
@@ -39,6 +39,15 @@ namespace isobus
 		{
 			get_next_can_message_from_tx_queue();
 		}
+
+		//! @todo: Remove this unsafe way of making a shared pointer to the network manager once the network manager is no longer a singleton,
```
ad3154, line 43: Double-free is rough... might be nicer to add a raw pointer version in the short term that we then remove once we ditch the singleton? Could make this raw pointer function badged or use a friend or something to keep it away from others if we wanted to.

ad3154 (commented):
Well, it certainly is more abstraction, as we'd now be abstracting the entire network manager from the higher level interfaces. This could be interesting for testing I suppose. It maybe has some weird implications in the app, where now you can't store some interfaces on the stack, you have to make shared_ptr's if you want to really use them, which IDK if is amazing, but I also understand... Trying to think of how someone could, in the future, maybe put a network manager on their local stack, plus all their various interfaces, seems like we'd need references or raw pointers, which I also don't love... so maybe all the shared_ptrs is fine. We've been going that route for some time now.

Adding a vtable to each of these interfaces is the same size as a reference or something to the network manager, so that's nice - for about the same class size you can abstract more stuff away.

I guess what I'm saying is I don't dislike it. I think it will get the job done. I think we do keep getting further from the ability to ever really make the stack static allocation friendly, which is maybe fine? I know some levels of functional safety forbid any dynamic memory - which would be pretty painful for us to deal with, so maybe it's not worth worrying about especially with nobody explicitly asking for it or sponsoring that kind of thing?

So I suppose I'm interested in your thoughts in return, but I think it's a fine approach.

GwnDaan:
> I think we do keep getting further from the ability to ever really make the stack static allocation friendly, which is maybe fine? I know some levels of functional safety forbid any dynamic memory - which would be pretty painful for us to deal with, so maybe it's not worth worrying about especially with nobody explicitly asking for it or sponsoring that kind of thing?

Yeah I know... Even though it would have been super nice to also be able to support those kind of strict rule sets, I think static allocation is out of our scope at the moment. Especially since we have always worked with the dynamic allocation and expanded on that. The shared pointers already help a lot to manage the lifecycle of those dynamically allocated objects. Converting the whole stack to be static allocation friendly I imagine will complicate so many things, it would be almost a rewrite...

ad3154:
Yeah... but I do think we should worry about our current stuff and releasing 1.0 before we think too hard about it the static allocation stuff I think (if ever) I was mostly just thinking out loud 😊

PR #463, opened 2024-04-17
[Core]: Refactorization of control function address claiming and factory methods 

GwnDaan:
## Describe your changes

With the goal in mind of getting rid of the singleton from the network manager, I removed a bit of dependencies of the control functions on the network manager. This should be another step closer to making it easy to flip the network manager around.

What changed:
- Moved the factory functions to the network manager:
  - `InternalControlFunction::create` -> `CANNetworkManager::create_internal_control_function`
  - `PartneredControlFunction::create` -> `CANNetworkManager::create_partnered_control_function`
- Removed the `AddressClaimStateMachine` class, and instead housed it directly into the `InternalControlFunction` class. They already had an 1:1 relation, so it basically was nothing more than an extension at this point.

## How has this been tested?

Made sure that the unit tests still correctly ran without making functional changes. Only minor thing is the removal of the check that a control function is correctly removed when it's "destroyed". I think that's something to look at later and currently gave more of a headache than it was useful. So I removed that GTEST assert.

Furthermore, I ran the seeder and VT3 object pool examples, they both still work fine. For most of the examples I removed the set preferred address, since according to the standard they have no function specifically defined for the currently configured preferred address:

> A CF claiming a preferred address in the range 0 to 127 and 248 to 253 shall perform the function defined for that preferred address and shall specify that function within its NAME.


isobus/src/can_internal_control_function.cpp
```diff
@@ -11,64 +11,378 @@
...
+				{
+					set_current_state(State::None);
+				}
+			}
+			break;
+
+			case State::SendArbitraryAddressClaim:
+			{
+				// Search the range of generally available addresses
+				for (std::uint8_t i = 128; i <= 235; i++)
```
ad3154, line 210: Technically this range depends on the NAME's industry group. We may want to enhance this to properly adjust the range at some point. For example, the construction industry group only allows for 128 to 207, with 208 and up being "reserved". We can reference the ranges on ISOBUS.net https://www.isobus.net/isobus/sourceAddress

isobus/src/can_network_manager.cpp
```diff
@@ -524,6 +519,13 @@ namespace isobus
 		}
 	}
 
+	std::shared_ptr<ControlFunction> CANNetworkManager::create_external_control_function(NAME desiredName, std::uint8_t address, std::uint8_t CANPort)
+	{
+		auto controlFunction = std::make_shared<ControlFunction>(desiredName, address, CANPort, ControlFunction::Type::External);
+		controlFunctionTable[CANPort][address] = controlFunction;
```
ad3154, line 525: Might be good to add sanity checks to these indexes before we try and use them, just in case we are somehow in an invalid state or in case the way we use this function ever changes, just to be safe

ad3154:
Looks pretty good, I think this will indeed help us move away from the singleton, which is great

isobus/include/isobus/isobus/can_message.hpp
```diff
@@ -11,6 +11,7 @@
 #ifndef CAN_MESSAGE_HPP
 #define CAN_MESSAGE_HPP
 
+#include "isobus/isobus/can_constants.hpp"
```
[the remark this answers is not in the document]
GwnDaan: In the removed `isobus/isobus/can_address_claim_state_machine.hpp` file there was this include `isobus/isobus/can_constants.hpp`. Since the address claim state machine header was included in the internal control function header, which in turn was included in the network manager. In short meaning; when one file previously included the `isobus/isobus/can_network_manager.hpp` file they also got the constants file. Since this link got removed a ton of of files were missing this header, so I assumed it was reasonable to put it here instead of in every file that was missing it... What do you think?

isobus/src/can_internal_control_function.cpp
GwnDaan: Added in https://github.com/Open-Agriculture/AgIsoStack-plus-plus/pull/463/commits/799f30292fae15e469eb85420a8630ce50c567ed

isobus/include/isobus/isobus/can_message.hpp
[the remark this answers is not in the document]
ad3154: I guess I generally prefer to minimize the scope of includes, that's OK with me

ad3154 (approved):

PR #477, opened 2024-06-16
[HW]: Add native logger for vector ascii CAN traces

ad3154:
## Describe your changes

This adds an interface to the hardware layer to natively create vector ascii log files, to eliminate the need for something like candump to log what your program is doing if you want. For devices like the PCAN probe on Windows, this eliminates the need for a second CAN device to log the bus. This is basically copied from AgIsoVirtualTerminal, but with Juce removed.

Also removed one unused parameter in the time/date interface.

It's best to instantiate this before you call "start" on the hardware layer so you get all the messages.

## How has this been tested?

Added a `VectorASCLogger` to the VT example, ran it for a while, stopped it, then compared with a file created by Peak's trc converter tool set to "version 17" of the file specification.

### DIFF
--- a/hardware_integration/src/vector_asc_logger.cpp
+++ b/hardware_integration/src/vector_asc_logger.cpp
@@ -1,0 +1,166 @@
+/*******************************************************************************
+** @file       vector_asc_logger.cpp
+** @author     Adrian Del Grosso
+** @copyright  The Open-Agriculture Developers
+*******************************************************************************/
+#include "isobus/hardware_integration/vector_asc_logger.hpp"
+
+#include "isobus/isobus/can_stack_logger.hpp"
+#include "isobus/utility/system_timing.hpp"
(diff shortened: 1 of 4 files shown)


hardware_integration/src/vector_asc_logger.cpp
```diff
@@ -0,0 +1,164 @@
...
+		openFile(filename, constructHeaderTime(currentTime));
+	}
+
+	VectorASCLogger::~VectorASCLogger()
+	{
+		if (logFileStream)
+		{
+			logFileStream.close();
+		}
+	}
```
GwnDaan, line 51: The 2 event listeners should be de-registered here. The automatic disposal has been removed some time ago because it was giving more confusion than it was worth.

ad3154: Added! 👍

GwnDaan (approved):

PR #479, opened 2024-06-18
[HW]: Fixed possibly dropping messages on Tx when Tx buffer is full

### LINKED ISSUE #476: Loading object pool to GS3 2630 display failing
Hi everyone

Very new to c++ so I appreciate your patience with the next few silly issue posts.

OS: Windows 11 10.0.22631
Compiler: Visual Studio 2022 with CMake tool
CAN driver: PCAN-USB

The bug: when running the SeederExample and VT3Example with a Topcon X35 everything works perfectly. But trying to use a John Deere 2630 display I keep running into a log error: 

[Error][VT]: Error in end of object pool message.Faulty Object 65535 Faulty Object Parent 65535 Pool error bitmask value 15

So it seems the issue is the bitmask value 15 but I just get to a dead end in the code and don't know what 15 means.

Additional context/red herring:
I have Jetter ISO-Designer and imported the VT3Example object pool, removed the image files, and using Version 3 setting exported the new .iop file to the example. It will upload successfully to the 2630, buttons and strings work correctly, but then if i add in more output strings once I reach a total of 22 strings I get the same error message previously described. Deleting one output string makes it work again. I may be going down a very wrong path here but it's the only way so far I can get the object pool to load successfully.

The 2630 has an option for ISOBUS certification mode. Turning on doesn't seem to make a difference.

Log:
Loaded object pool from VT3TestPool.iop
[Info][NM]: Partnered control function with name a0001d00042708d7 has claimed address 38 on channel 0.
[Info][VT]: No version label from the VT matched. Client will upload the pool and store it instead.
[Warn][VT]: Get Versions Response ignored!
[Error][VT]: Error in end of object pool message.Faulty Object 65535 Faulty Object Parent 65535 Pool error bitmask value 15
[Info][VT]: Resetting Failed VT Connection
[Info][VT]: No version label from the VT matched. Client will upload the pool and store it instead.
[Error][VT]: Error in end of object pool message.Faulty Object 65535 Faulty Object Parent 65535 Pool error bitmask value 15
[Info][VT]: Resetting Failed VT Connection
[Info][VT]: No version label from the VT matched. Client will upload the pool and store it instead.

I would like to take a CAN trace, if anyone can please let me know how with my hardware/system that would be great.

Thank you

ad3154:
## Describe your changes

Fixed an issue where we weren't checking the return value of the push into our Tx queue in our hardware interface. This could have caused some frames to be accidentally dropped.

## How has this been tested?

Recreated the issue discussed [here](https://github.com/Open-Agriculture/AgIsoStack-plus-plus/issues/476#issuecomment-2174568075) with AgIsoVirtualTerminal with the number of frames per DOP set to 255, and witnessed frames being dropped.

Added this check, and now object pool properly uploads!

This may fix #476

### DIFF
--- a/hardware_integration/src/can_hardware_interface.cpp
+++ b/hardware_integration/src/can_hardware_interface.cpp
@@ -314,9 +314,9 @@ namespace isobus
 			return false;
 		}
 
-		if (channel->frameHandler->get_is_valid())
+		if ((channel->frameHandler->get_is_valid()) &&
+		    (channel->messagesToBeTransmittedQueue.push(frame)))
 		{
-			channel->messagesToBeTransmittedQueue.push(frame);
 #if !defined CAN_STACK_DISABLE_THREADS && !defined ARDUINO
 			updateThreadWakeupCondition.notify_all();
 #endif


GwnDaan (approved):
Good catch! This unchecked push turned into an issue when we switched over to the fixed size ringbuffer 3 months ago🫣

PR #482, opened 2024-07-08
[VT]: Add a way to track VT float attributes

ad3154:
## Describe your changes

This lets you use the VT state helper to deal with float attributes. Previously, float values were probably not handled correctly.
I was seeing that there were warnings in the seeder example about this:
image

## How has this been tested?

- [x] Run Seeder Example and test variable scale factors

### DIFF
--- a/isobus/include/isobus/isobus/isobus_virtual_terminal_client_update_helper.hpp
+++ b/isobus/include/isobus/isobus/isobus_virtual_terminal_client_update_helper.hpp
@@ -74,6 +74,18 @@ namespace isobus
 		/// @return True if the attribute was set successfully, false otherwise.
 		bool set_attribute(std::uint16_t objectId, std::uint8_t attribute, std::uint32_t value);
 
+		/// @brief Sets the value of a float attribute of a tracked object.
+		/// @attention ONLY use this function for float attributes defined in ISO11783-6,
+		/// otherwise you will get incorrect results. Scale on output numbers, for example, is a float.
+		/// @note If the to be tracked working set consists of more than the master,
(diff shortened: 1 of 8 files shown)


isobus/include/isobus/isobus/isobus_virtual_terminal_client_update_helper.hpp
```diff
@@ -74,6 +74,18 @@ namespace isobus
...
+		/// @attention ONLY use this function for float attributes defined in ISO11783-6,
+		/// otherwise you will get incorrect results. Scale on output numbers, for example, is a float.
+		/// @note If the to be tracked working set consists of more than the master,
+		/// this function is incompatible with a VT prior to version 4. For working sets consisting
+		/// of only the master, this function is compatible with any VT version.
+		/// @param[in] objectId The object id of the attribute to set.
+		/// @param[in] attribute The attribute to set.
+		/// @param[in] value The value to set the attribute to.
+		/// @return True if the attribute was set successfully, false otherwise.
+		bool set_attribute(std::uint16_t objectId, std::uint8_t attribute, float value);
```
GwnDaan, line 87: thought: hmm, seems a lot of duplication that we might be able to avoid? With your current implementation we leave the conversion from float value to uint32_t all the way up to the VT client. What if we instead do it in this function directly? And then call the neighbor helper `set_attribute` function instead

ad3154: Hmmm good idea I think, that does sound appealing. I can rework it

GwnDaan (approved):

PR #486, opened 2024-07-27
[TC]: Add distance thresholds and handling of default process data

### LINKED ISSUE #485: [TC Client] Add a way to generically handle request for the default process data
This is related to a discussion in the discord.

Basically, if a user wants to specify a default time interval for a given process data variable, we currently don't expose a way to set one, so if a TC server doesn't command a specific time interval, the client is just out of luck.

We should have some callback of some kind where, when the default process data is requested, we loop through the DDOP to find objects that are part of the default process data set, and for each one, allow the user to enable any triggers they want or mark them as needing to have their initial value sent, or both.

**Describe alternatives you've considered**
There's probably a few different ways to implement this... a callback like the `request_value_command_callback `, or maybe some pure virtual interface that needs to be passed into the object. Lot of options. Callback might be the most consistent with how we're doing other things at the moment.

Another option would be to have the users pass in all the possible data when constructing the DDOP, but that seems less desirable, because if we accept a binary DDOP, there'd still be no way to set it.

**Additional context**
[Discord context](https://discord.com/channels/1168280592002056293/1168280827872944158/1266413784013078550)

ad3154:
## Describe your changes

- Added support for distance interval measurement commands to the TC client
    - Users can use the `set_distance` function on the TC client now to cause the client to evaluate all TC-configured distance interval measurement commands, which previously weren't really supported. Distance triggers are not often used, but at least we now support it. When the distance set to the client causes any process data variable to need to be resent, the client will now automatically query the user's value callback and send an update to the TC. In general it's very automated, but the user is required to manage the vehicle's total driven distance in the consuming application, maybe by using our speed and distance interfaces or via proprietary means.
- Added support for defining and automatically querying a consuming application for a DDOP's default process data.
    - The TC requests a special DDI (57343 - Request Default Process Data) from device element 0 when it wants the client to send values to it using the default trigger method. Previously this wasn't really supported via the stack, so this adds generic handling of it. Users can add a callback which will be called when the TC requests this. Inside that callback, the user can tell the TC client to configure any number of default triggers, and we'll error check them against the DDOP content.

Fixes #485 

## How has this been tested?

- [ ] Tested with TC
- [ ] Tested via unit tests (needs a little improvement yet)


isobus/src/isobus_task_controller_client.cpp
```diff
@@ -928,6 +1196,98 @@ namespace isobus
 		return retVal;
 	}
 
+	void TaskControllerClient::populate_any_triggers_from_settings(std::shared_ptr<task_controller_object::DeviceProcessDataObject> processDataObject, std::uint16_t elementNumber, std::uint16_t DDI, const DefaultProcessDataSettings &settings)
+	{
+		if (nullptr != processDataObject)
+		{
+			if (settings.enableChangeThresholdTrigger)
+			{
+				if (0 != (processDataObject->get_trigger_methods_bitfield() & static_cast<std::uint8_t>(task_controller_object::DeviceProcessDataObject::AvailableTriggerMethods::ThresholdLimits)))
```
GwnDaan, line 1205:
```suggestion
				if (0 != (processDataObject->get_trigger_methods_bitfield() & static_cast<std::uint8_t>(task_controller_object::DeviceProcessDataObject::AvailableTriggerMethods::OnChange)))
```

ad3154:
Tested working with a real TC

PR #499, opened 2024-10-16
[TC]: store reported by client version to use it later for DDOP parsing

oleszczyk:
It looks like version reported by the client can be quite useful for example for proper parsing of received DDOP before it will be activated. Otherwise version 4 is used for parsing by default which can be wrong assumption.

## Describe your changes

So far while DDOP is parsing some default (4) version is used. There is no direct possibility to check which version received binary DDOP stream is actually coded. On the other hand client reports its version in advanced so it can be stored and use later on for proper DDOP parsing.

This PR contains adding new public field to class `ActiveClient` which stores `reportedVersion` for later use.

## How has this been tested?

Own `TaskController` implemented on ESP32 connected to real Gaspardo corn seeder ISOBUS controller uint.


ad3154:
Will merge this for now - but would be nice if we passed the version into the appropriate callback as well (the activate callback?) so be really obvious to a consumer that they should use the version when parsing the DDOP.

ad3154 (approved):

PR #504, opened 2024-11-08
NTCAN hardware plugin

GwnDaan:
## Describe your changes

Continued #198

I will create another PR in the coming days that adds an example on how to bridge two hardware plugins together. This allows one to use a virtual bus, while connecting it to (multiple) physical busses as well.

## How has this been tested?

 I have tested it using their virtual can bus and everything is functioning as it should.

### DIFF
--- a/hardware_integration/CMakeLists.txt
+++ b/hardware_integration/CMakeLists.txt
@@ -100,6 +100,10 @@ if("SYS_TEC" IN_LIST CAN_DRIVER)
   list(APPEND HARDWARE_INTEGRATION_INCLUDE "UsbCanLs.h")
   list(APPEND HARDWARE_INTEGRATION_INCLUDE "UsbCanUp.h")
 endif()
+if("NTCAN" IN_LIST CAN_DRIVER)
+  list(APPEND HARDWARE_INTEGRATION_SRC "ntcan_plugin.cpp")
+  list(APPEND HARDWARE_INTEGRATION_INCLUDE "ntcan_plugin.hpp")
+endif()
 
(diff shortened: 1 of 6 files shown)


GwnDaan:
Hi all, any chance someone has time to review this PR for me? I'd like to continue work that depends on the contributed code

ad3154:
I think my main concern from before was just that I didn't know if we could include ntcan stuff due to license issues.

It looks like you're not adding `ntcan.h` or the `.lib` file - so I guess, what's your plan here? If the end goal is just that we want to provide a virtual bus that works cross-process, we might want to develop something home-grown for that, IDK

hardware_integration/CMakeLists.txt
```diff
@@ -304,6 +308,46 @@ if("SYS_TEC" IN_LIST CAN_DRIVER)
...
+        target_link_libraries(HardwareIntegration
+                              PRIVATE $ENV{CanSdkDir}/lib/VC/i386/ntcan.lib)
+        target_include_directories(HardwareIntegration
+                                   PUBLIC $ENV{CanSdkDir}/include)
+      else()
+        message(
+          FATAL_ERROR
+            "NTCAN Selected but no supported processor arch was detected. Only x64 and x86 are supported."
+        )
+      endif()
```
GwnDaan, line 337: @ad3154 have a look at this part. It dynamically compiles the needed header and library files into the stack based on a system variable already set by NTCan. I think it's a pretty neat approach that doesn't involve any licenses

ad3154 (approved):

PR #512, opened 2024-12-10
Use proprietary colors used by Jetter IsoDesigner instead of just white

martonmiklos:
## Describe your changes

While testing some implements with AgIsoVT I came across invisible texts on white background.
They turned out to be lightgray on an Amatron 4:
kép

They used 252 color code for this OutputNumber which is proprietary, and so far all proprietary were treated as white.

I looked up Jetter IsoDesigner (as the IOP in question was designed with it), and they color palette is the following:
kép

## How has this been tested?

Built AgIsoVT with this change and the texts got visible

### DIFF
--- a/isobus/src/isobus_virtual_terminal_objects.cpp
+++ b/isobus/src/isobus_virtual_terminal_objects.cpp
@@ -43,11 +43,31 @@ namespace isobus
 			colourTable[i] = VTColourVector((51.0f * (redCounter) / 255.0f), ((51.0f * (greenCounter)) / 255.0f), ((51.0f * blueCounter) / 255.0f));
 		}
 
-		// The rest are proprietary. Init to white for now.
-		for (std::uint16_t i = 232; i < VT_COLOUR_TABLE_SIZE; i++)
-		{
-			colourTable[i] = VTColourVector(1.0f, 1.0f, 1.0f);
-		}
+		// Proprietary colors copied it over manually from the Jetter Iso Designer color palette
+		colourTable[232] = VTColourVector(0.039f, 0.039f, 0.039f);
+		colourTable[233] = VTColourVector(0.082f, 0.082f, 0.082f);
+		colourTable[234] = VTColourVector(0.122f, 0.122f, 0.122f);
+		colourTable[235] = VTColourVector(0.165f, 0.165f, 0.165f);
+		colourTable[236] = VTColourVector(0.216f, 0.216f, 0.216f);
+		colourTable[237] = VTColourVector(0.243f, 0.243f, 0.243f);
+		colourTable[238] = VTColourVector(0.294f, 0.294f, 0.294f);
+		colourTable[239] = VTColourVector(0.333f, 0.333f, 0.333f);
+		colourTable[240] = VTColourVector(0.369f, 0.369f, 0.369f);
+		colourTable[241] = VTColourVector(0.416f, 0.416f, 0.416f);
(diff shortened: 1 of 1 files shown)


ad3154:
I'm not sure that it makes sense to use proprietary colors in our VT... ultimately it's the client's responsibility to assign colours based on the VT they are connecting to, and in this case the client is doing something explicitly discouraged by the ISO Standard. There are other more standard means for a client to alter the colour table without needing to modify the server, and choosing Jetter's proprietary behavior is pretty arbitrary I think.

```
Use of these proprietary means where the Working Set is provided by a different manufacturer than that which provides the VT is not recommended in order to provide maximum ISO compatibility. As these items are proprietary, the Working Set or VT manufacturer can change their proprietary means without disclosure.
```

ad3154:
Interested to hear if there are other opinions on this one.

martonmiklos:
Well, I think following the Jetter's palerre will provide better compatibility than the current "all white" approach.  

The standard says the following note about the prop. colors:


I am not a native English speaker but as I underatood mapping the prop. colors to grayscale is recommended which aligns with the Jetter palette.

ad3154:
Mmmm, I'm not sure that sentence is referring to the prop. range... I think it's more just saying that if your VT is monochrome you can just use grey interpolated values for all colors? And even then, since it's version 5 and before IDK if that still holds anyways for the server side unless we explicitly are configured to report compliance with version 5 or below

ad3154:
I'm just on the fence about it.... what do you think @GwnDaan ? Do you think greys for the proprietary range makes sense?

martonmiklos:
> Mmmm, I'm not sure that sentence is referring to the prop. range... I think it's more just saying that if your VT is monochrome you can just use grey interpolated values for all colors? 

That the same thing what I was wondering.
If I would be writing the similar documentation I would separate the non-correlating into distinct bullets.

> Do you think greys for the proprietary range makes sense?

We do use the problematic color code 252 in our own implement to highlight the disabled input widgets. We have tested it with least 10 type of terminals internally and I am quite sure that it would be reported by users if these widgets would became invisible (with the daylight mode white background they do disappear totally). 

I could create an IOP to show the proprietary palette and "plugfest" it with the terminals available for me to take screenshots, but it would take quite a bit effort.

I do hate pulling up the "proven in use" card in similar situations but this is something what applies here.

GwnDaan:
Interesting discussion

> Do you think greys for the proprietary range makes sense?

Yes, simply because it helps someone today (while still conforming to the standard)

**BUT:** this is definitely not a recommended practice. Hence I highly suggest @martonmiklos and his team to refactor away from using that colour on other virtual terminal servers other than their own.

martonmiklos:
> Interesting discussion
> 
> > Do you think greys for the proprietary range makes sense?
> 
> Yes, simply because it helps someone today (while still conforming to the standard)
> 
> **BUT:** this is definitely not a recommended practice. Hence I highly suggest @martonmiklos and his team to refactor away from using that colour on virtual terminal servers other than their own for VT version 3 and prior. For Version 4 and later, please start using the Colour Map object (Annex B.17)

I understand and acknowledge that we doing something which is not recommended by the standard. We would happily migrate to VT5 but our stack is limited to VT3, but anyway using a different gray from the non proprietary palette would also make sense.

ad3154:
OK, sounds good.

ad3154 (approved):

PR #517, opened 2024-12-29
Support of Change Size command to all object types where it should be supported

martonmiklos:
## Describe your changes

There were some object types where the change size command was not handled.

## How has this been tested?

A Müller implement had a OutputLinearBarGraph which got resized by button press resulting error output and invisible widget. 
If I fixed it I added the other resizeable objects too.

### DIFF
--- a/isobus/src/isobus_virtual_terminal_server.cpp
+++ b/isobus/src/isobus_virtual_terminal_server.cpp
@@ -1168,18 +1168,29 @@ namespace isobus
 											}
 											break;
 
+											case isobus::VirtualTerminalObjectType::PictureGraphic:
+											{
+												targetObject->set_width(newWidth);
+											}
+											break;
+
 											case VirtualTerminalObjectType::Animation:
 											case VirtualTerminalObjectType::Button:
 											case VirtualTerminalObjectType::Container:
+											case VirtualTerminalObjectType::InputBoolean:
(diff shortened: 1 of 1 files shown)


isobus/src/isobus_virtual_terminal_server.cpp
```diff
@@ -1168,18 +1168,29 @@ namespace isobus
...
 											case VirtualTerminalObjectType::OutputArchedBarGraph:
 											case VirtualTerminalObjectType::OutputEllipse:
 											case VirtualTerminalObjectType::OutputLine:
+											case VirtualTerminalObjectType::OutputLinearBarGraph:
 											case VirtualTerminalObjectType::OutputList:
 											case VirtualTerminalObjectType::OutputNumber:
 											case VirtualTerminalObjectType::OutputPolygon:
 											case VirtualTerminalObjectType::OutputRectangle:
 											case VirtualTerminalObjectType::OutputString:
+											case VirtualTerminalObjectType::ScaledGraphic:
```
ad3154, line 1193: Scaled graphic doesn't list change size command as an allowed command in ISO 11783-6, and isn't an output field object, so IDK if that one applies... same thing with picture graphic, if I am reading it correctly (which, maybe I'm not) that one should only allow "change attribute" and "get attribute"?

ad3154 (approved):

PR #525, opened 2025-01-06
Remove duplicated code

martonmiklos:
## Describe your changes

Removed duplicate codes to support Open-Agriculture/AgIsoVirtualTerminal#55

## How has this been tested?

Loaded Kverneland implements in AgIsoVT

### DIFF
--- a/isobus/include/isobus/isobus/isobus_virtual_terminal_objects.hpp
+++ b/isobus/include/isobus/isobus/isobus_virtual_terminal_objects.hpp
@@ -365,6 +365,24 @@ namespace isobus
 		std::uint8_t backgroundColor = 0; ///< The background color (from the VT colour table)
 	};
 
+	/// @brief Common class for VT objects having variable reference
+	class VTObjectWithVariableReference : public VTObject
+	{
+	public:
+		/// @brief Returns the object ID of a variable object that contains the value of the current object
+		/// or the null ID if the "value" attribute is used instead.
+		/// @returns The object ID of a variable object that contains the value of the current object, or the null ID
+		std::uint16_t get_variable_reference() const;
+
+		/// @brief Sets the object ID of the variable object that contains the value of the current object.
+		/// Does no error checking on the type of the supplied object.
+		/// @param[in] variableValue The object ID of the variable object that contains the value of the current object, or the null ID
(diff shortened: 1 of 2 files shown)


GwnDaan:
image
This my way of finding what's wrong :)

martonmiklos:
> image @martonmiklos This my way of finding what's wrong :)

Dauh, Ctrl+F and dynamic page load are my new enemy :D

GwnDaan (approved):
Good job! I like how you were able to reduce the duplication with a few simple base classes

GwnDaan:
@martonmiklos one nitpick, can you rephrase your commit message to be a little more descriptive?

e.g. `refactor(vt-objects)!: add several base classes to reduce code duplication`

This way, someone can scrolls through to commits get an idea of the changes made, and whether or not they are breaking to them. https://www.conventionalcommits.org/en/v1.0.0/ is pretty nice if you ask me

GwnDaan:
Also @ad3154, I would like to hear your thoughts on this and whether your initial concern still applies

> One reason they are not combined right now is because they inherit from `isobus::OutputNumber` vs `isobus::InputNumber` for example, which allows for convenient use of `isobus::VTObject::get_object_type` and other things like that. In theory I'm not opposed to combining them but there will probably be other things that need to get re-done to make that happen, or at the very least you may need to override some `virtual` functions in creative ways

_Originally posted by @ad3154 in https://github.com/Open-Agriculture/AgIsoVirtualTerminal/issues/55#issuecomment-2540211669_

martonmiklos:
> @martonmiklos one nitpick, can you rephrase your commit message to be a little more descriptive?
> 
> e.g. `refactor(vt-objects)!: replace duplicated functions with base classes`
> 
> This way, someone can scrolls through to commits to get an idea of the changes made, and whether or not they are breaking to them. https://www.conventionalcommits.org/en/v1.0.0/ is pretty nice if you ask me

I have fixed the commit message, I will try to keep myself to the mentioned conventions in the future.

ad3154:
> Also @ad3154, I would like to hear your thoughts on this and whether your initial concern still applies

I think this does a good job at reducing duplication, and it leaves `get_object_type` alone which is nice, so I'm happy.

ad3154 (approved):

PR #528, opened 2025-01-07
[HWadd VSCAN interface

GwnDaan:
requires you to set the 'VSCAN_API_DIR' environment variable to the API SDK directory that contains the `.h`, `.lib` files. Furthermore, you need to put `vs_can_api.dll` in the same directory as the executable.

## Describe your changes

Fixes # (issue)

## How has this been tested?

```bash
TODO
```


hardware_integration/include/isobus/hardware_integration/vscan_plugin.hpp
```diff
@@ -0,0 +1,74 @@
...
+		bool reconfigure(std::string channel, void *baudrate = VSCAN_SPEED_250K);
+
+	private:
+		/// @brief Parses the error from the status code
+		/// @param[in] status The status code to parse
+		/// @returns The error message
+		std::string parse_error_from_status(VSCAN_STATUS status);
+
+		std::string channel; ///< The COM port or IP address of the VSCAN device to use.
+		void *baudrate; ///< The baudrate to use for the CAN connection.
```
ad3154, line 68: Holding on to a void * pointer for baudrate? 🤔

That's a little odd

hardware_integration/include/isobus/hardware_integration/vscan_plugin.hpp
```diff
@@ -0,0 +1,74 @@
...
+/// to install their driver first, which in-turn requires you to agree to their terms and conditions.
+/// @author Daan Steenbergen
+///
+/// @copyright 2025 The Open-Agriculture Developers
+//================================================================================================
+#ifndef VSCAN_PLUGIN_HPP
+#define VSCAN_PLUGIN_HPP
+
+#include <string>
+#include "vs_can_api.h"
```
ad3154, line 15: As much as I hate to ask, I assume this includes windows.h? I always hate including windows.h in a header file if I can avoid it....

At the very least maybe consider a `WIN32_LEAN_AND_MEAN` define if they don't have one in there already maybe?

hardware_integration/src/vscan_plugin.cpp
```diff
@@ -0,0 +1,161 @@
...
+			this->channel = channel;
+			this->baudrate = baudrate;
+			retVal = true;
+		}
+		return retVal;
+	}
+
+	std::string VSCANPlugin::parse_error_from_status(VSCAN_STATUS status)
+	{
+		char errorBuffer[256] = { 0 };
```
ad3154, line 156: Big ol crunchy stack allocation there, oof, but I guess it's fine since it's PC hardware

ad3154 (approved):

hardware_integration/src/vscan_plugin.cpp
GwnDaan: Fixed with a copilot suggestion :)

hardware_integration/include/isobus/hardware_integration/vscan_plugin.hpp
GwnDaan: Yeah I hate that it includes windows.h too... But I don't think adding `WIN32_LEAN_AND_MEAN` is better:

ChatGPT:
> You should not define WIN32_LEAN_AND_MEAN inside your library header file unless you are certain that the code using your library does not need the excluded parts of windows.h. It is better to document this requirement and leave it up to the library user to control.

Meaning if we define `WIN32_LEAN_AND_MEAN`, then the end-user using our stack might face trouble with "their include" not including all the needed parts... I think it's better if we leave it to them to disable if they want to. Though idk how we can document this best? Something like a CMake warning perhaps?

hardware_integration/include/isobus/hardware_integration/vscan_plugin.hpp
GwnDaan:
```c+
#define VSCAN_SPEED_1M					(void*)8
#define VSCAN_SPEED_800K				(void*)7
#define VSCAN_SPEED_500K				(void*)6
#define VSCAN_SPEED_250K				(void*)5
#define VSCAN_SPEED_125K				(void*)4
#define VSCAN_SPEED_100K				(void*)3
#define VSCAN_SPEED_50K				(void*)2
#define VSCAN_SPEED_20K				(void*)1
```
Their baudrates are defined as `void *` for whatever reason 🤷🏽‍♂️

hardware_integration/include/isobus/hardware_integration/vscan_plugin.hpp
ad3154: Eh, IDK if that's a real concern, as a consumer could always just include the Windows headers later, and if you don't define it, Windows throws in crazy macros like `MAX` that interfere with common syntax...

But they can also just define it in their project before including the plugin too, so IDK if it really matters.

Let's leave it as is

ad3154 (approved):

PR #548, opened 2025-02-17
Add infrastructure for VT identify message handling in AgIsoVt

martonmiklos:
## Describe your changes

Added infrastructure for VT number handling and VT number identification. This includes:
- Adding setter API for the J1939 NAME field of the control funcitons
- Add virtual callback for handling the Identify  VT command 

Fixes Open-Agriculture/AgIsoVirtualTerminal/#81

## How has this been tested?

Built AgIsoVT supporting the setting of the VT number using this stack


isobus/src/can_control_function.cpp
```diff
@@ -46,6 +46,11 @@ namespace isobus
 		return controlFunctionNAME;
 	}
 
+	void ControlFunction::set_NAME(const NAME &newNAME)
+	{
+		controlFunctionNAME = newNAME;
```
ad3154, line 51: What's the intent of this setter? IDK if we would want it exposed to everything, because if it's called on an internal control function it will not restart address claiming or anything....

martonmiklos: I used this to alter the NAME of the CF here, right after it's initialization:
https://github.com/martonmiklos/AgIsoVirtualTerminal/blob/support_setting_vt_number/src/ServerMainComponent.cpp#L51

What do you think if I move all the settings handling in AgIsoVT to a separate class what I can initialize before the creation of the ServerMainComponent handle the CF Name change before the CF initialization. This way we can get rid from this change in the stack.

ad3154: I think that would be fine. Changing the internal control function's NAME properly would be a bit complicated, as we'd have to set the address claim state back to none, and mark it as invalid/no address in the network manager also for it to really work after initialization

martonmiklos: I sorted out in AgIsoVT the settings loading, so I could remove this method. 
Are you comfortable with the rest of the changes in this PR?

ad3154 (approved):

PR #553, opened 2025-03-14
[NM]: Added a method for forcing a request of the address claim PGN

ad3154:
## Describe your changes

This could be used to check for the continued existence of a control function at the network level. 

This should be used sparingly, but I can see it being desirable.

It is not to be used as a normal part of the address claim process, as normal address claiming is handled elsewhere.

RE: #542 

## How has this been tested?

Added unit test. This code was literally copied from the internal control function, so it should work the same.


martonmiklos (approved):
LGTM

PR #560, opened 2025-03-26
Fix building with CAN logger disabled

martonmiklos:
## Describe your changes

The isobus module has a DISABLE_CAN_STACK_LOGGER option.
When this option is set the build failed because the CANStackLogger methods used across the stack but not defined. 
Added dummy methods to CANStackLogger to overcome this error.

## How has this been tested?

```bash
cmake .. -DDISABLE_CAN_STACK_LOGGER=ON
make
#successful build
```

### DIFF
--- a/hardware_integration/src/vector_asc_logger.cpp
+++ b/hardware_integration/src/vector_asc_logger.cpp
@@ -114,7 +114,7 @@ namespace isobus
 		}
 		else
 		{
-			CANStackLogger::error("[ASC Logger]: Failed to open log file for writing");
+			LOG_ERROR("[ASC Logger]: Failed to open log file for writing");
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


ad3154:
Instead of doing this, the correct solution would be to replace instances of `CANStackLogger::thing` with the appropriate macros that were created to solve this initially, such as `LOG_WARNING` instead of `CANStackLogger::warn` for example. 

These are already defined in can_stack_logger.hpp

```
#ifdef DISABLE_CAN_STACK_LOGGER
/// @brief A macro which removes a "critical" log statement depending on the state of DISABLE_CAN_STACK_LOGGER
/// @param logString A log statement
#define LOG_CRITICAL(...)
/// @brief A macro which removes a "error" log statement depending on the state of DISABLE_CAN_STACK_LOGGER
/// @param logString A log statement
#define LOG_ERROR(...)
/// @brief A macro which removes a "warning" log statement depending on the state of DISABLE_CAN_STACK_LOGGER
/// @param logString A log statement
#define LOG_WARNING(...)
/// @brief A macro which removes a "info" log statement depending on the state of DISABLE_CAN_STACK_LOGGER
/// @param logString A log statement
#define LOG_INFO(...)
/// @brief A macro which removes a "debug" log statement depending on the state of DISABLE_CAN_STACK_LOGGER
/// @param logString A log statement
#define LOG_DEBUG(...)
```

ad3154 (changes requested):
See comment above about log macros

martonmiklos:
Ah I should have spotted those macros. I will replace the log calls with those macros.

ad3154 (approved):

PR #562, opened 2025-04-03
isobus vt: Add event dispatcher for signaling softkey mask change

martonmiklos:
## Describe your changes

Added event dispatcher from the stack to signal softkeymask changes
Necessary for Open-Agriculture/AgIsoVirtualTerminal#71

## How has this been tested?

Built AgIsoVT where this event handler is used.

### DIFF
--- a/isobus/src/isobus_virtual_terminal_objects.cpp
+++ b/isobus/src/isobus_virtual_terminal_objects.cpp
@@ -590,6 +590,7 @@ namespace isobus
 			set_soft_key_mask(newMaskID);
 			retVal = true;
 		}
+
 		return retVal;
 	}
 
--- a/isobus/src/isobus_virtual_terminal_server.cpp
+++ b/isobus/src/isobus_virtual_terminal_server.cpp
@@ -95,6 +95,11 @@ namespace isobus
 		return onChangeActiveMaskEventDispatcher;
 	}
 
+	EventDispatcher<std::shared_ptr<VirtualTerminalServerManagedWorkingSet>, std::uint16_t, std::uint16_t> &VirtualTerminalServer::get_on_change_active_softkey_mask_event_dispatcher()
+	{
+		return onChangeActiveSoftKeyMaskEventDispatcher;
+	}
+
 	EventDispatcher<std::shared_ptr<VirtualTerminalServerManagedWorkingSet>, std::uint16_t, bool> &VirtualTerminalServer::get_on_focus_object_event_dispatcher()
 	{
 		return onFocusObjectEventDispatcher;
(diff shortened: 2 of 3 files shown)


isobus/src/isobus_virtual_terminal_objects.cpp
```diff
@@ -7,6 +7,8 @@
 /// @copyright 2023 The Open-Agriculture Developers
 //================================================================================================
 #include "isobus/isobus/isobus_virtual_terminal_objects.hpp"
+#include <iostream>
+#include <ostream>
```
ad3154, line 11: These headers appear unneeded. Maybe you had them in there for debugging? Can be removed I think.

isobus/src/isobus_virtual_terminal_server.cpp
```diff
@@ -1338,67 +1343,69 @@ namespace isobus
...
+														parentServer->send_change_soft_key_mask_response(dataOrAlarmMaskId, newSoftKeyMaskId, 0, message.get_source_control_function());
+														parentServer->onChangeActiveSoftKeyMaskEventDispatcher.call(cf, dataOrAlarmMaskId, newSoftKeyMaskId);
+														parentServer->process_macro(targetMask, EventID::OnChangeSoftKeyMask, VirtualTerminalObjectType::DataMask, cf);
 													}
 													else
 													{
-														LOG_WARNING("[VT Server]: Client %u change soft key mask command: failed to set mask for data mask object %u to %u", cf->get_control_function()->get_address(), objectID, newObjectID);
-														parentServer->send_change_soft_key_mask_response(objectID, newObjectID, (1 << static_cast<std::uint8_t>(ChangeSoftKeyMaskErrorBit::AnyOtherError)), message.get_source_control_function());
+														CANStackLogger::warn("[VT Server]: Client %u change soft key mask command: failed to set mask for data mask object %u to %u", cf->get_control_function()->get_address(), dataOrAlarmMaskId, newSoftKeyMaskId);
```
ad3154, line 1385: This seems like it reverts the use of the LOG_WARNING macro, which was probably not what you meant to do

ad3154 (changes requested):

isobus/src/isobus_virtual_terminal_server.cpp
martonmiklos: Ah, i jumped into my own trap. I will came up with a PR  to add a hook template which checks the usage of the proper logging macros..

isobus/src/isobus_virtual_terminal_objects.cpp
martonmiklos: Yes, this is true, removed them.

isobus/src/isobus_virtual_terminal_server.cpp
ad3154, line 1385: I do still see `CANStackLogger::warn` in the latest diff

martonmiklos: Uh, sorry, I messed it up again :(
It should be OK now.

ad3154 (approved):

PR #568, opened 2025-04-12
Fixed a divide by zero issue when calculating IOP load percentage

ad3154:
## Describe your changes

This was causing an assert in the VT when iopSize was exactly zero.

This fixes that by returning 0% when the IOP size is zero.

## How has this been tested?

Verified that the assert stopped happening when uploading an object pool using a debugger.


martonmiklos (approved):
LGTM

PR #574, opened 2025-04-18
Fixed issues with PGN requests for the address claimed PGN sent to specific destination addresses

### LINKED ISSUE #573: Destination specific address claim requests are not handled by address claim logic
**Describe the bug**
Most control functions send the request PGN to the global address when requesting the address claim PGN, but the standard specifically allows for the request to be sent directly to a destination address as well in ISO 11783-5. We currently do not handle this correctly, and pass the request to the PGN request protocol rather than diverting it to the address claiming logic in the internal control function.

**Supporting Documentation**

[vt.log](https://github.com/user-attachments/files/19813948/vt.log)

RE: ISO11783-5 4.4.2.2

A user in our discord was able to create this issue using an  InCommand 800 and our VT client on ESP32.

ad3154:
## Describe your changes

- Fixed an issue where ICFs could send too many address claims if a destination specific request was made.
    - This change makes ICFs only re-claim in response to a request to do so when the destination of the request was the ICFs address or the global address. Previously they would claim for any request to any address which was incorrect.
- Don't use destination specific requests for address claim when determining online/offline CF states
    - ISO 11783-5 permits destination specific PGN requests for the address claimed PGN, and in those cases we were treating it the same as a global request when determining if a CF went offline or not. This fixes that such that only global requests will run that logic.

Fixes #573 

## How has this been tested?

Tested by simulating destination specific address claims and walking the code path in the debugger to ensure that the online/offline logic in `update_address_table` was bypassed, and ensured that ICFs filter out requests to other addresses when going to the `SendReclaimAddressOnRequest` state

### DIFF
--- a/isobus/src/can_internal_control_function.cpp
+++ b/isobus/src/can_internal_control_function.cpp
@@ -64,6 +64,9 @@ namespace isobus
 				std::uint32_t requestedPGN = message.get_uint24_at(0);
 
 				if ((static_cast<std::uint32_t>(CANLibParameterGroupNumber::AddressClaim) == requestedPGN) &&
+				    ((CANIdentifier::GLOBAL_ADDRESS == message.get_identifier().get_destination_address()) ||
+				     ((get_address_valid()) &&
+				      (get_address() == message.get_identifier().get_destination_address()))) &&
 				    (State::AddressClaimingComplete == get_current_state()))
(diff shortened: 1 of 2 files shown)


martonmiklos (approved):
LGTM, sorry for the delay

PR #576, opened 2025-04-19
Make the End Object Pool response error reporting more descriptive

### LINKED ISSUE #575: End of Object Pool response -- Clear Text Error Codes
Image

the 65535 could be better printed as "NULL"

also
The Message "Bitmask 2" does not really give Information to the User. Also it is unclear if it is a Byte 2 or Byte 7 Bitmask. 

it would be VERY COOL if the Error would be printed in clear Text: 

Byte2: 
Error Codes (0 = no errors)
Bit 0 = 1 = There are errors in the Object Pool,
refer to Bytes 3 to 8 for additional error
information
Bit 1 = 1 = VT ran out of memory during transfer
Bit 2, 3 = Reserved, set to 0
Bit 4 = 1 = any other error
Bit 5―7 = Reserved, set to 0

Byte7
Object Pool Error Codes (0 = no errors)
Bit 0 = 1 = method or attribute not supported
by the VT
Bit 1 = 1 = unknown object reference (missing
object)
Bit 2 = 1 = any other error
Bit 3 = 1 = object pool was deleted from volatile
memory
Bit 4―7 = Reserved, set to 0

martonmiklos:
Fixes #575

### DIFF
--- a/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
+++ b/isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
@@ -1594,6 +1594,11 @@ namespace isobus
 		/// @brief The worker thread will execute this function when it runs, if applicable
 		void worker_thread_function();
 
+		/// @brief A method to print the End of Object pool response message error field with LOG_ERROR
+		/// @param[in] errorCodes byte 2 from the End of Object Pool response
+		/// @param[in] objectPoolErrorCodes byte 7 from the End of Object Pool response
+		static void print_objectpool_error(uint8_t errorCodes, uint8_t objectPoolErrorCodes);
+
 		static constexpr std::uint32_t VT_STATUS_TIMEOUT_MS = 3000; ///< The max allowable time between VT status messages before its considered offline
 		static constexpr std::uint32_t WORKING_SET_MAINTENANCE_TIMEOUT_MS = 1000; ///< The delay between working set maintenance messages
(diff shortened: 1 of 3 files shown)


utility/include/isobus/utility/to_string.hpp
```diff
@@ -14,6 +14,7 @@
 
 #include <sstream>
 #include <string>
+#include "../../isobus/include/isobus/isobus/can_constants.hpp"
```
ad3154, line 17: Since this is a library dependency of the main library, probably best to not cross these streams.

`object_id_to_string` is literally only used in the VT client right now anyways, so might just want to move the changes from this file to that file for now

martonmiklos: You are right, moved this to isobus_virtual_terminal_objects.hpp as it fits there better.

isobus/include/isobus/isobus/isobus_virtual_terminal_client.hpp
```diff
@@ -1594,6 +1594,11 @@ namespace isobus
 		/// @brief The worker thread will execute this function when it runs, if applicable
 		void worker_thread_function();
 
+		/// @brief A method to print the End of Object pool response message error field with LOG_ERROR
+		/// @param[in] errorCodes byte 2 from the End of Object Pool response
+		/// @param[in] objectPoolErrorCodes byte 7 from the End of Object Pool response
+		static void print_objectpool_error(uint8_t errorCodes, uint8_t objectPoolErrorCodes);
```
ad3154, line 1600:
```suggestion
		static void print_objectpool_error(std::uint8_t errorCodes, std::uint8_t objectPoolErrorCodes);
```

ad3154 (approved):

PR #580, opened 2025-04-29
fix: Ensure PCF callbacks only trigger for matching source control functions

### LINKED ISSUE #579: Bug: PartneredControlFunction Callbacks Triggering for Unrelated Messages
**Describe the bug**
There is a bug in the CAN message routing where callbacks registered for a specific PartneredControlFunction (PCF) are being triggered for messages from other control functions that share the same PGN. This results in callbacks being executed for messages that should be filtered out based on the source control function.

## Current Behavior
- When multiple PCFs register callbacks for the same PGN (e.g., ProprietaryA), all registered callbacks are triggered regardless of the message's source control function
- Developers must manually check the source control function in their callback implementations to filter out unwanted messages
- Example of current workaround needed:
```cpp
if (message.get_source_control_function() != pcf) return;
```

## Expected Behavior
- Callbacks should only be triggered for messages that come from the specific PCF that registered the callback
- The message routing system should automatically filter messages based on both PGN and source control function
- No manual source control function checking should be required in callback implementations
- 
## Root Cause
In `CANNetworkManager::process_can_message_for_global_and_partner_callbacks`, messages are routed to PCFs based only on PGN matching, without verifying that the message's source control function matches the PCF that registered the callback.

## Proposed Fix
Add source control function validation in the message routing:
```cpp
if ((nullptr != partner) && 
   (partner->get_can_port() == message.get_can_port_index()) && 
   (partner == messageSource)) // Add this check 
{ // Process callbacks only for matching partner ... }
```

## Impact
- Improves message routing accuracy
- Reduces callback overhead by preventing unnecessary callback executions
- Eliminates need for manual source control function checking in callback implementations
- Makes callback behavior more intuitive and less error-prone

## Verification Steps
1. Create multiple PCFs registering for the same PGN
2. Send messages with this PGN from different control functions
3. Verify that callbacks are only triggered for messages from their respective PCF
4. Remove manual source control function checks from callbacks
5. Verify system still functions correctly

## Additional Notes
This issue affects any implementation using multiple PCFs that monitor the same PGN, particularly in systems with multiple similar devices on the bus.

**Environment:**
- OS: ESP32
- CAN Driver TWAI

j-c-cook:
## Describe your changes

When multiple Partnered Control Functions (PCFs) register callbacks for the same Parameter Group Number (PGN), the callbacks were being triggered for any message with a matching PGN, regardless of the source control function. This caused callbacks to execute for messages from unrelated control functions.

The fix adds source control function validation in the message routing by adding a check `(partner == messageSource)` in the PCF message processing logic. This ensures callbacks are only triggered when both the PGN matches and the message comes from the specific PCF that registered the callback.

This change:
- Prevents callbacks from being triggered by unrelated control functions
- Eliminates the need for manual source control function checking in callbacks
- Makes callback behavior more predictable and correct

closes #579 

## How has this been tested?

Test: Verified callbacks only trigger for messages from their specific PCF

### DIFF
--- a/isobus/src/can_network_manager.cpp
+++ b/isobus/src/can_network_manager.cpp
@@ -1054,9 +1054,9 @@ namespace isobus
 	void CANNetworkManager::process_can_message_for_global_and_partner_callbacks(const CANMessage &message) const
 	{
 		std::shared_ptr<ControlFunction> messageDestination = message.get_destination_control_function();
-                std::shared_ptr<ControlFunction> messageSource = message.get_source_control_function();
+		std::shared_ptr<ControlFunction> messageSource = message.get_source_control_function();
 
-                if ((nullptr == messageDestination) &&
+		if ((nullptr == messageDestination) &&
 		    ((nullptr != messageSource) ||
 		     ((static_cast<std::uint32_t>(CANLibParameterGroupNumber::ParameterGroupNumberRequest) == message.get_identifier().get_parameter_group_number()) &&
 		      (NULL_CAN_ADDRESS == message.get_identifier().get_source_address()))))
@@ -1079,7 +1079,7 @@ namespace isobus
 			{
 				if ((nullptr != partner) &&
 				    (partner->get_can_port() == message.get_can_port_index()) &&
-                                    (partner == messageSource))
+				    (partner == messageSource))
 				{
 					// Message matches CAN port for a partnered control function
 					for (std::size_t k = 0; k < partner->get_number_parameter_group_number_callbacks(); k++)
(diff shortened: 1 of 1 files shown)


j-c-cook:
I currently plan to mark this issue ready once I’ve created a minimal example that uses socketcan vcan so it can simply be recreated on other dev’s machines.

j-c-cook:
Here's a full minimal example that exposes the bug I've described: https://gist.github.com/j-c-cook/21793e9df45107c2641233420936eb6c.

ad3154 (approved):

PR #608, opened 2025-09-26
[Tests]: Update googletest

ad3154:
This just updates googletest to silence some nasty CMake deprecation warnings.


martonmiklos (approved):
Ship it!

PR #618, opened 2025-10-23
TC: fix working width discovery

gunicsba:
I noticed that the Müller Elektronik sprayer working widths weren't correct.

Now it has fallbacks according to: https://www.isobus.net/isobus/attachments/345/ISO11783-11-DDI-290-SetpointWorkState-v1.pdf

Each Section Device Element shall at least provide one type of Working Width. If
more than one type of Working Width is provided, then the Section Controller shall be
capable to use the different Working Width types with the following priority:
1. Actual Working Width (DDI 67)
2. Maximum Working Width (DDI 70)
3. Default Working Width (DDI 68)


gunicsba:
@GwnDaan  I made the changes but not sure why the Build started to fail. It doesn't make sense. previous build only had clang-format error...

gunicsba:
@ad3154 @martonmiklos  something isn't right with our tests .. for me they seem to fail randomly...
and not just for this PR but for other PRs too.

ad3154:
> @ad3154 @martonmiklos something isn't right with our tests .. for me they seem to fail randomly... and not just for this PR but for other PRs too.

Yeah, it's the timing related stuff... the tests do actual thread delays and read actual system time everywhere, which is prone to unpredictability. Ideally we just need to add an abstraction to our timing interface to allow overriding it from a test environment.  Adding this abstraction would also greatly speed up the tests....

ad3154:
> @ad3154 @martonmiklos something isn't right with our tests .. for me they seem to fail randomly... and not just for this PR but for other PRs too.

I have opened a PR that should finally completely fix the random test failures. It was not fun, but it needed to happen, haha.

https://github.com/Open-Agriculture/AgIsoStack-plus-plus/pull/682

gunicsba:
Added a few more tests and implemented the same for the booms.

isobus/src/isobus_task_controller_server.cpp
```diff
@@ -694,16 +694,21 @@ namespace isobus
...
 
-						if (1 <= numberOfWorkingSetMembers)
+						if (numberOfWorkingSetMembers == 0)
 						{
-							if (nullptr == get_active_client(rxMessage.get_source_control_function()))
-							{
-								activeClients.push_back(std::make_shared<ActiveClient>(rxMessage.get_source_control_function()));
-							}
+							LOG_ERROR("[TC Server]: Working set master reported zero members – invalid!");
+							return;
```
sujandumaru: Do we want to return from this function on here: `while (!rxMessageQueue.empty())` is still being processed?

gunicsba: Claude has fixed this.

ad3154 (dismissed):

test/tc_server_tests.cpp
```diff
@@ -1515,6 +1515,157 @@ TEST_F(TaskControllerServerTest, DDOPHelper_NoFunctions)
 	EXPECT_EQ(4000, implement.booms.at(0).sections.at(0).zOffset_mm.get());
 }
 
+TEST(TASK_CONTROLLER_SERVER_TESTS, DDOPHelper_SubBoomWidthPriority)
```
sujandumaru, line 1518: This test suite name does not match with other test suite in this file. Aren't they supposed to be `TaskControllerServerTest`? or is there another reason to keep them separate?

gunicsba: @ad3154 @martonmiklos what do you suggest? Should we align this towards Task Controller?

isobus/src/isobus_task_controller_server.cpp
```diff
@@ -761,21 +761,26 @@ namespace isobus
 					{
 						std::uint8_t numberOfWorkingSetMembers = rxData[0];
 
-						if (1 <= numberOfWorkingSetMembers)
+						if (numberOfWorkingSetMembers == 0)
```
martonmiklos, line 764: 0 == numberOfWorkingSetMembers

isobus/src/isobus_task_controller_server.cpp
```diff
@@ -761,21 +761,26 @@ namespace isobus
...
+							activeClients.push_back(std::make_shared<ActiveClient>(rxMessage.get_source_control_function()));
+							LOG_INFO("[TC Server]: New client 0x%02X detected via WorkingSetMaster. Requesting version.",
+							         rxMessage.get_source_control_function()->get_address());
+							// Proactively request version information from the new client
+							send_generic_process_data_default_payload(static_cast<std::uint8_t>(TechnicalDataCommandParameters::RequestVersion),
+							                                          rxMessage.get_source_control_function());
 						}
-						else
+
+						if (numberOfWorkingSetMembers != 1)
```
martonmiklos, line 781: 1 != numberOfWorkingSetMembers

test/tc_server_tests.cpp
martonmiklos: Yes, please use TaskControllerServerTest as test suite name.

martonmiklos (changes requested):

gunicsba:
## Working Width Discovery Fix - Summary

The PR is squashed into a single commit on top of current `main`, and the review comments are addressed. It only contains the working width fix; the Section / SubBoom deduplication will be done in a separate PR later.

### Problem
Section and sub boom widths were only read from `ActualWorkingWidth` (DDI 67). Some devices (e.g. Müller Elektronik sprayers) only provide Maximum or Default Working Width, so the width was missing.

### Width priority (ISO 11783-10, DDI 290 SetpointWorkState)
1. `ActualWorkingWidth` (DDI 67 / 0x0043)
2. `MaximumWorkingWidth` (DDI 70 / 0x0046)
3. `DefaultWorkingWidth` (DDI 68 / 0x0044)

`Section` and `SubBoom` store all three values. `get_width_with_priority()` picks the first one that is present and non-zero, so the order in the DDOP does not matter. If none qualify, an empty `ObjectPoolValue` is returned.

### Other changes
- TC server: a working set master reporting zero members logs a specific error (and uses `break` instead of `return`, so the rx message queue keeps being processed). When more than one member is reported, the client is still accepted and the unsupported member count is logged as an error.
- Comparisons now put the constant first (`0 == numberOfWorkingSetMembers`, `1 != numberOfWorkingSetMembers`).

### Tests (`test/tc_server_tests.cpp`, suite `TaskControllerServerTest`)
- `DDOPHelper_SectionWidthPriority`: Actual wins even when referenced last, Maximum wins over Default, Default only, and a zero Actual width is skipped.
- `DDOPHelper_SubBoomWidthPriority`: all three widths, Maximum + Default, and Default only.
- `DDOPHelper_SubBoomWidthPriorityWithProcessData`: values come from DeviceProperty and editable flags from DeviceProcessData.

All 33 TC server / DDOP helper tests pass locally (MSVC, Debug). clang-format has been applied.

### Backward compatibility
`width_mm` is still populated, and no existing members were renamed or removed.

martonmiklos (approved):

PR #619, opened 2025-10-28
[TC] Fix remove offset and instead clamp to allowed range

### LINKED ISSUE #598: Data Distionary Value formatting issue.
I've noticed there is an issue regarding the format_value method https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/main/isobus/src/isobus_data_dictionary.cpp#L41. Using a Muller Maxi ecu, all the values such as Actual Working Width (DDI 67) give the value directly, not using the offset which seems to comply with how isobus.net reports the value range to be 0 to 2147483647. Maybe I'm not understanding where the offset is used correctly but reading the raw value, it shows as the correct number and gets the offset incorrectly added to it.

I see there is an option to disable the dictionary but is there some flag I'm missing which dictates which values should be offset?

GwnDaan:
## Describe your changes
This PR fixes an offset being incorrectly applied. Basically in no DDI there an offset. It's just used for limiting the range. Hence we only clamp it now. 

Fixes #598 

The last two things I did:
- run the generator again and have the latest changes from isobus.net
- add pip3 and install requirements to the devcontainer

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -29,7 +29,7 @@
 	// "forwardPorts": [],
 
 	// Use 'postCreateCommand' to run commands after the container is created.
-	"postCreateCommand": "gcc -v"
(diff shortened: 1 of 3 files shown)


martonmiklos (approved):
LGTM

PR #626, opened 2025-11-22
Transport Protocol Data Transfer Optimization

j-c-cook:
This PR aims to optimize the data transfer receive path across all multi-frame transport protocols by replacing byte-by-byte copying with bulk memory operations using `memcpy()`.


j-c-cook:
I've quantified the performance improvement gained here: https://gist.github.com/j-c-cook/876a37804ca1f8cace22b1147198a243
performance_comparison

GwnDaan (approved):
LGTM! Great work

PR #628, opened 2025-11-30
Add support for parsing Scaled Graphics objects from the IOP pool

martonmiklos:
## Describe your changes

This PR adds support for parsing object pools containing scaled graphics

Fixes # (issue)

## How has this been tested?

NX9 VT6 test implement contains a scaled graphics object.


isobus/include/isobus/isobus/isobus_virtual_terminal_objects.hpp
```diff
@@ -4539,6 +4539,149 @@ namespace isobus
...
+			PositionTop = 0,
+			PositionMiddle = 1,
+			PositionBottom = 2,
+		};
+
+		/// @brief Bitmask of the horizontal justification bits in the scale type attribute of a Scaled graphics object
+		static constexpr uint8_t VERTICAL_JUSTIFICATION_MASK = 0x18;
+
+		/// @brief Bitmask of the vertical justification bits in the scale type attribute of a Scaled graphics object
+		static constexpr uint8_t HORIZONTAL_JUSTIFICATION_MASK = 0x60;
```
GwnDaan: I think it should be the other way around?

```c++
static constexpr std::uint8_t HORIZONTAL_JUSTIFICATION_MASK = 0x18; // bits 3–4
static constexpr std::uint8_t VERTICAL_JUSTIFICATION_MASK   = 0x60; // bits 5–6
```

isobus/src/isobus_virtual_terminal_working_set_base.cpp
```diff
@@ -2835,6 +2835,85 @@ namespace isobus
...
+
+					if (iopLength >= tempObject->get_minumum_object_length())
+					{
+						tempObject->set_id(decodedID);
+						tempObject->set_width((static_cast<std::uint16_t>(iopData[3]) | (static_cast<std::uint16_t>(iopData[4]) << 8)));
+						tempObject->set_height((static_cast<std::uint16_t>(iopData[5]) | (static_cast<std::uint16_t>(iopData[6]) << 8)));
+						tempObject->set_scale_type(iopData[7]);
+						tempObject->set_options(iopData[8]);
+						tempObject->set_graphic_id((static_cast<std::uint16_t>(iopData[9]) | (static_cast<std::uint16_t>(iopData[10]) << 8)));
+						const std::uint8_t numberOfMacrosToFollow = iopData[11];
```
GwnDaan: You check `if (iopLength >= tempObject->get_minumum_object_length())` which means at least 9 bytes. But you read on this line the 12th byte, so that can go wrong

isobus/src/isobus_virtual_terminal_working_set_base.cpp
```diff
@@ -2835,6 +2835,85 @@ namespace isobus
...
+									else
+									{
+										LOG_ERROR("[WS]: Macro with ID %u which is listed as part of a scaled graphics object %u has an invalid or unsupported event ID:", iopData[1], decodedID, iopData[0]);
+										retVal = false;
+										break;
+									}
+								}
+
+								iopLength -= 2;
+								iopData += 2;
```
GwnDaan: This goes wrong for ExtendedMacroReferences. They are 4 bytes

GwnDaan (changes requested):

isobus/src/isobus_virtual_terminal_objects.cpp
```diff
@@ -8623,4 +8623,166 @@ namespace isobus
 		pointerType = type;
 	}
 
+	VirtualTerminalObjectType ScaledGraphicObject::get_object_type() const
+	{
+		return VirtualTerminalObjectType::ScaledGraphic;
+	}
+
+	bool ScaledGraphicObject::get_attribute(uint8_t attributeID, uint32_t &returnedAttributeData) const
```
ad3154:
```suggestion
	bool ScaledGraphicObject::get_attribute(std::uint8_t attributeID, std::uint32_t &returnedAttributeData) const
```

Check out the other places in this file as well for this pls

PR #630, opened 2025-12-01
Fix and uniformize parsing extended macro references

martonmiklos:
Utilizing a common function to parse macro references and move the pointers correctly when parsing extended macro references.

### DIFF
--- a/isobus/include/isobus/isobus/isobus_diagnostic_protocol.hpp
+++ b/isobus/include/isobus/isobus/isobus_diagnostic_protocol.hpp
@@ -43,6 +43,7 @@
 
 #include <array>
 #include <list>
+#include <map>
 #include <memory>
 #include <string>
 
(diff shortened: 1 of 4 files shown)


isobus/src/isobus_virtual_terminal_working_set_base.cpp
```diff
@@ -112,44 +112,15 @@ namespace isobus
 								// Next, parse macro list
 								if (iopLength >= sizeOfMacros)
 								{
```
GwnDaan: I'd say we can also remove this check and just rely on the helper function instead.

```cpp
retVal = parse_object_macro_reference(tempObject, numberOfMacrosToFollow, iopData, iopLength);
if (!retVal)
{
    LOG_ERROR("[WS]: Failed to parse macro references for object " + isobus::to_string(static_cast<int>(decodedID)));
}
```

isobus/src/isobus_virtual_terminal_working_set_base.cpp
```diff
@@ -2963,12 +2684,18 @@ namespace isobus
 			// If the first byte is 255, then more bytes are used! 4.6.22.3
 			if (iopData[0] == static_cast<std::uint8_t>(EventID::UseExtendedMacroReference))
```
GwnDaan: Add an early return with `iopLength < 1` and then this function is robust to not need the check mentioned in the first comment.

PR #633, opened 2025-12-02
Add supported hardware interfaces to the README.md

martonmiklos:
I thought to be a good idea to maintain some more attractive/accessible list of the supported CAN interfaces.

It might extend the standard README quite a bit, so I am open to create a separate md file in the hardware_interfaces and just point a link to that from the main page.


ad3154:
It ended up being a very large section - I would like to keep the readme fairly short if possible. We do have [this list](https://isobus-plus-plus.readthedocs.io/en/latest/api/hardware/index.html) already with instructions. Maybe we should funnel people there? Otherwise yea maybe a sub page? I am not sure

GwnDaan (approved):

PR #641, opened 2025-12-14
[CI]: Automated build for testing no threads

### LINKED ISSUE #632: CAN_STACK_DISABLE_THREADS is not checked in all occasions
**Describe the bug**
There are a few places that uses std::mutex, std::thread, etc. without checking the disable macro first

**Supporting Documentation**
- in some cases it is probably just changing std::lock_guard to the internal macro (`LOCK_GUARD(Mutex, managedWorkingSetMutex);`) like [here](https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/3b5dd50fc3e08bc1bdcff3191f164c82e684d100/isobus/src/isobus_virtual_terminal_working_set_base.cpp#L2950)
- but for example [here](https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/3b5dd50fc3e08bc1bdcff3191f164c82e684d100/isobus/src/isobus_virtual_terminal_server_managed_working_set.cpp#L41) I'am not sure if there is an easy fix possible or if a greater effort is needed, to remove the usage of std::thread

**Environment:**

- all envs, but it leads to compile errors especially on platforms that don't have thread support e.g. gcc arm none eabi

GwnDaan:
## Describe your changes
See #638, this PR continues the work with adding a Github Actions build to verify all code in the future build fine with disabled threading.

Fixes #632

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,7 @@ project(
 
 include(GNUInstallDirs)
 include(CMakePackageConfigHelpers)
+include(CheckIncludeFileCXX)
 
 # Make CTest available and the BUILD_TESTING option (OFF by default)
 option(BUILD_TESTING "Set to ON to enable building of tests from top level" OFF)
@@ -33,6 +34,16 @@ option(
   CAN_STACK_DISABLE_THREADS
   "Set to ON to disable multi-threading, which removes the need for std::thread, std::mutex, and related libraries."
   OFF)
+
+# Disable threads if the header is not available
+check_include_file_cxx(thread HAS_STD_THREAD_HEADER)
+if(NOT HAS_STD_THREAD_HEADER)
+  message(
+    WARNING
+      "C++ <thread> header not found. Disabling CAN_STACK_DISABLE_THREADS.")
+  set(CAN_STACK_DISABLE_THREADS ON)
+endif()
+
 if(NOT CAN_STACK_DISABLE_THREADS AND NOT ARDUINO)
   # Find packages required for Threading
   set(THREADS_PREFER_PTHREAD_FLAG ON)


isobus/src/isobus_virtual_terminal_objects.cpp
```diff
@@ -3226,20 +3225,6 @@ namespace isobus
...
-		{
-			auto child = get_object_by_id(get_variable_reference(), parentWorkingSet->get_object_tree());
-
-			if ((nullptr != child) && (isobus::VirtualTerminalObjectType::StringVariable == child->get_object_type()))
-			{
-				return std::static_pointer_cast<isobus::StringVariable>(child)->get_value();
-			}
-		}
-		return get_value();
-	}
```
GwnDaan: @martonmiklos we have to remove this function in order to keep using the `isobus_virtual_terminal_objects.hpp` in the VT client code. Are you fine with changing the VirtualTerminal GUI project to make it do the logic itself?

VzdornovNA88: I apologize for interfering in the conversation, I looked at this method, well... it doesn't seem to work for VT, If you look closely at the implementations of other objects in this file, you can see that many of them also use this logic with a 'referenced value', it's just that the working set was needed only here and cutting off the 'displayed_value' just because we have a CAN_STACK_DISABLE_THREADS knife would not be a very correct solution in my opinion, and although I don't really agree that this method should be called that way (I still see it as a get_value overload for example) it would surprise me to see this logic in VT. If so much code hadn't been done for 'referenced values', I would have preferred a different way to get them - something like OutputStringRefferenced = make_ref_value_from<OutputString>(workingSetPtr ) - this interface could be located in another file that depends on the runtime of the working set but now that a lot has already been done... nevertheless, porting this code to VT looks strange, it's better to follow a more conservative path in my opinion.

CMakeLists.txt
```diff
@@ -33,6 +34,16 @@ option(
...
   OFF)
+
+# Disable threads if the header is not available
+check_include_file_cxx(thread HAS_STD_THREAD_HEADER)
+if(NOT HAS_STD_THREAD_HEADER)
+  message(
+    WARNING
+      "C++ <thread> header not found. Disabling CAN_STACK_DISABLE_THREADS.")
+  set(CAN_STACK_DISABLE_THREADS ON)
+endif()
```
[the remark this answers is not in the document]
VzdornovNA88: of course, it's a strange case when a user disables it, but he doesn't have the source code, I agree. This is just a formal nuance, I consider all combinations to be test cases.

ad3154 (approved):

PR #653, opened 2025-12-23
[Build] : fix detection of standard threading model support by the toolchain

### LINKED ISSUE #652: [Build]: Automated build for testing no threads (#641) doesn't do the job well
So, as promised, I started playing with this PR (#641 ) and immediately discovered several problems that I hadn't seen when the PR was under review.

**Problems :**
1. My suggestion was to use include(CheckIncludeFileCXX) to check support for the standard threading model, but I didn't take into account that some toolchains actually provide sources, it's just that support for a specific function is blocked at the preprocessor level (for example, gcc 10.2.1, I don't think anything will change in the future while clang does not provide such sources for arm), as for the unification of the building for different toolchains, this method is not correct and does not do the right job, I want to note here , what I don't pass on when building '-DCAN_STACK_DISABLE_THREADS=ON"' is actually what the source check was supposed to do, which was actually supposed to be named as "checking support for the standard C++ threading model".

2. And one more thing (I don't know if this is a problem), based on the semantics of what was done, that is, stubs for standard library files were added, in them we detect a compilation error using the preprocessor and say that we prohibit the inclusion of sources that relate to the threading model if it is disabled in the library and then #include<thread> was added to all the resulting targets, but at the same time, the paths to the stubs were not taken into account in the cmake files of such targets, and they were taken into account for the isobus library itself, but not for the examples.This leads to the fact that we do not get the desired error at the beginning of a long list of errors when trying to build such an example.

**Environment and build command :**

I work in vscode, so here I will provide the output data and user arguments for cmake : (The toolchain file for arm none eabi toolchains is used below, which you can find in #647 )

Ubuntu 20.04

`[build] /usr/bin/stm32-gcc-cmake-tool-chain/eabi/bin/arm-none-eabi-g++ --sysroot=/usr/bin/stm32-gcc-cmake-tool-chain/eabi/bin/../arm-none-eabi  -DCAN_PORT_MAXIMUM_VALUE=4 -DCAN_STACK_DISABLE_THREADS -DISOBUS_UNKNOWN_AVAILABLE -I../examples/diagnostic_protocol -I../isobus/include -I../hardware_integration/include -I../utility/include -specs=nosys.specs -mcpu=cortex-m7 -march=armv7e-m -mfpu=neon -mfloat-abi=hard -mthumb -ffunction-sections -fdata-sections -fno-rtti -fno-exceptions -fno-threadsafe-statics -g3 -O2 -DNDEBUG   -Wall -Wextra -MD -MT examples/diagnostic_protocol/CMakeFiles/DiagnosticProtocolExample.dir/main.cpp.obj -MF examples/diagnostic_protocol/CMakeFiles/DiagnosticProtocolExample.dir/main.cpp.obj.d -o examples/diagnostic_protocol/CMakeFiles/DiagnosticProtocolExample.dir/main.cpp.obj -c ../examples/diagnostic_protocol/main.cpp
[build] In file included from ../examples/diagnostic_protocol/main.cpp:10:`

{
  "cmake.configureArgs": [
        "-DCMAKE_TOOLCHAIN_FILE=/home/vzdornovna/yocto_work/current_projects/AgIsoStack-plus-plus/cmake/arm-none-eabi.cmake",
        "-DARM_TOOLCHAIN_PATH=/home/vzdornovna/yocto_work/current_projects/toolchains/arm-none-eabi/LLVMEmbeddedToolchainForArm-13.0.0",
        "-DARM_CPU_ARCH=cortex-m7-neon",
        // "-DCAN_STACK_DISABLE_THREADS=ON"
        "-DBUILD_EXAMPLES=ON"
        // "-DBUILD_TESTING=ON"
    ],
}

**to repeat, you can just take #647 for yourself, previously it took care of disabling the threading model itself, but now that we are seriously fixing this feature in main, I disabled this functionality in #647 and now it is completely consistent with the behavior of main.**

VzdornovNA88:
In accordance with issue #652 , I changed the method of detecting support for the standard threading model, as well as added fake sources to the target paths for examples that, in my opinion, were missed.

Fixes #652

I tested this locally with #647 (with locally installed gcc 10 for arm and clang 13 for arm) and another locally installed gcc 13 toolchain for x86_64 (ubuntu 20.04). 

It should be understood that I tested switching between toolchains and the correct state of the CAN_STACK_DISABLE_THREADS flag after configuring the build, the important point is that I do not pass this parameter to the command as a user settings, otherwise the whole check will be meaningless. As for the paths to the fake sources, here I checked for the presence of the main error in the build, which we ourselves defined in these sources.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,6 +99,12 @@ add_subdirectory("hardware_integration")
 
 option(BUILD_EXAMPLES "Set to ON to enable building of examples from top level"
        OFF)
+if(BUILD_EXAMPLES AND CAN_STACK_DISABLE_THREADS)
+  message(
+    WARNING
+      "You are trying to build examples while the isobus library is configured in single-threaded mode, examples cannot be built in this mode. The current build configuration is (BUILD_EXAMPLES == ON AND CAN_STACK_DISABLE_THREADS == ON), please change the configuration."
+  )
+endif()
 if(BUILD_EXAMPLES)
   add_subdirectory("examples/transport_layer")
   add_subdirectory("examples/diagnostic_protocol")


VzdornovNA88:
@GwnDaan, please review this changes when you have a time, thank you. Thanks for any questions and additions.

GwnDaan (approved):
LGTM! thanks

PR #660, opened 2026-01-08
Seperate speed and distance message classes into their own files

ad3154:
## Describe your changes

The `SpeedMessagesInterface` was very large. This helps to scope down each message to its own file while maintaining the same actual interface for the `SpeedMessagesInterface`, which hopefully makes it easier to add new messages and maintain old ones.


isobus/CMakeLists.txt
```diff
@@ -30,11 +30,14 @@ set(ISOBUS_SRC
     "nmea2000_fast_packet_protocol.cpp"
     "isobus_data_dictionary.cpp"
     "isobus_language_command_interface.cpp"
+    "isobus_machine_selected_speed.cpp"
```
sujandumaru, line 33: Don't we need to add these new header files in CMake?

sujandumaru (dismissed):

sujandumaru:
I am thinking to merge this PR as this should simplify huge chunk of isobus speed related code into their respective file making it easier to understand. I don't think there are any issues right now. @martonmiklos do you want to take a look into this?

sujandumaru:
Made some minor fixes on this before merging.

sujandumaru (approved):

PR #686, opened 2026-05-14
Fix CANMessage 24-bit big-endian parsing

sujandumaru:
## Describe your changes

Fixes 24-bit big-endian parsing in `CANMessage::get_uint24_at()` and `CANMessage::get_int24_at()`.

Previously, the big-endian path used `data[index + 2]` for both the most-significant and least-significant byte, skipping `data[index]`. This made big-endian 24-bit values decode incorrectly.

Also adds CAN message tests covering:
- 24-bit little-endian and big-endian reads
- 32-bit little-endian and big-endian reads

## How has this been tested?

```bash
cmake --build build
ctest --test-dir build --output-on-failure -R CAN_MESSAGE_TESTS.DataCorrectnessTest
ctest --test-dir build --output-on-failure

### DIFF
--- a/examples/virtual_terminal/esp32_platformio_object_pool/platformio.ini
+++ b/examples/virtual_terminal/esp32_platformio_object_pool/platformio.ini
@@ -11,7 +11,7 @@
 default_envs = denky32 ; Don't build the CI environment by default, to avoid unnecessary build errors for beginners who select the default environment
 
 [env:denky32]
-platform = espressif32
+platform = espressif32@6.10.0
 board = denky32
 framework = espidf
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: examples/virtual_terminal/esp32_platformio_object_pool/platformio.ini [hunk 1]
```diff
@@ -11,14 +11,15 @@
 default_envs = denky32 ; Don't build the CI environment by default, to avoid unnecessary build errors for beginners who select the default environment
 
 [env:denky32]
-platform = espressif32
+platform = espressif32@6.10.0
```
sujandumaru, line 14: This pins `espressif32` to `6.10.0` so the ESP32 example build keeps passing while `TWAIPlugin` still uses the legacy TWAI driver.

`espressif32@7.0.1` pulls in ESP-IDF 6.0.1, where `driver/twai.h` emits a legacy-driver deprecation warning. Since this project treats warnings as errors, that warning fails CI. `6.10.0` is the latest PlatformIO Espressif32 release I verified locally that builds this example successfully without requiring a full migration to the new `esp_twai.h` API.

A complete TWAI modernization should be handled separately because it requires more than an include change.

ad3154 (approved):

PR #693, opened 2026-05-31
Feat: add client version tracking and callback to TaskControllerServer

gunicsba:
Add automatic version detection and callback mechanism for Task Controller server to track connected client versions.

Changes:
- Add on_client_version_received() virtual callback for applications
- Add get_client_version() getter to retrieve stored client version
- Add request_client_version() method for proactive version queries
- Automatically request version when new client detected via WorkingSetMaster
- Add INFO/DEBUG logging for version exchange events
- Add comprehensive unit tests for version tracking functionality

The callback fires when client version info is received, supporting both V3 (SecondEditionDraft) and V4+ (SecondPublishedEdition). Version exchange is optional in V3 but automatic in the server implementation.

### DIFF
--- a/examples/task_controller_server/main.cpp
+++ b/examples/task_controller_server/main.cpp
@@ -93,6 +93,12 @@ public:
 		// This callback lets you know when a client sends a process data acknowledge (PDACK) message to you
 	}
 
+	void on_client_version_received(std::shared_ptr<isobus::ControlFunction> clientControlFunction, std::uint8_t version) override
+	{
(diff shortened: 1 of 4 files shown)


ad3154:
This looks OK to me. If you can update the example to fix the pipeline failure, then I would be OK with approving it.

gunicsba:
I tested this with a couple equipments this week and seems to work with all of them .

ad3154 (approved):

PR #694, opened 2026-06-02
Fix VT server SonarQube issues

sujandumaru:
## Describe your changes

Fixes VT server SonarQube findings related to implicit narrowing conversions and a shadowed constant.

- Reused the existing `VERSION_LABEL_LENGTH` static constant instead of declaring a local constant with the same name.
- Added explicit extraction helpers for values narrowed from `std::uint32_t` to smaller VT object value types.
- Updated VT server numeric value handling so 8-bit and 16-bit object setters receive explicitly sized values.
- Reduced SonarQube issues in `isobus_virtual_terminal_server.cpp` from 82 to 36.

The remaining SonarQube issues require larger refactoring work and completion of existing TODO items, so they are intentionally left for now.

## How has this been tested?

```bash
cmake --build build
ctest --test-dir build --output-on-failure

### DIFF
--- a/isobus/src/isobus_virtual_terminal_server.cpp
+++ b/isobus/src/isobus_virtual_terminal_server.cpp
@@ -19,6 +19,47 @@
 
 namespace isobus
 {
+	static constexpr std::uint8_t get_low_byte(std::uint16_t value) noexcept
+	{
+		return static_cast<std::uint8_t>(value & 0xFFU);
+	}
+
+	static constexpr std::uint8_t get_high_byte(std::uint16_t value) noexcept
+	{
+		return static_cast<std::uint8_t>((value >> 8) & 0xFFU);
+	}
+
+	static constexpr std::uint8_t get_bit(std::uint8_t bitIndex) noexcept
+	{
+		return (bitIndex < 8U) ? static_cast<std::uint8_t>(1U << bitIndex) : static_cast<std::uint8_t>(0U);
+	}
+
+	static constexpr std::uint8_t get_byte(std::uint32_t value, std::uint8_t byteIndex) noexcept
+	{
+		return (byteIndex < 4U) ? static_cast<std::uint8_t>((value >> (byteIndex * 8U)) & 0xFFU) : static_cast<std::uint8_t>(0U);
+	}
+
+	static constexpr std::uint16_t get_low_16_bits(std::uint32_t value) noexcept
+	{
+		return static_cast<std::uint16_t>(value & 0xFFFFU);
+	}
+
+	static std::uint16_t get_little_endian_uint16(const std::vector<std::uint8_t> &data, std::size_t index)
+	{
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: isobus/src/isobus_virtual_terminal_server.cpp [hunk 1]
```diff
@@ -19,6 +19,50 @@
 
 namespace isobus
 {
+	namespace
```
ad3154: I think we can remove this anonymous namespace? We don't use these typically

isobus/src/isobus_virtual_terminal_server.cpp
```diff
@@ -19,6 +19,50 @@
...
+		}
+
+		constexpr std::uint8_t get_high_byte(std::uint16_t value) noexcept
+		{
+			return static_cast<std::uint8_t>((value >> 8) & 0xFFU);
+		}
+
+		constexpr std::uint8_t get_bit(std::uint8_t bitIndex) noexcept
+		{
+			return static_cast<std::uint8_t>(1U << bitIndex);
```
ad3154: Functions like this make me a bit nervous when they don't have input validation, since bitIndex >=8 is undefined behavior

isobus/src/isobus_virtual_terminal_server.cpp
```diff
@@ -404,14 +437,12 @@ namespace isobus
...
 				return true;
 			}
-			break;
 
 			case Function::GetWindowMaskDataMessage:
 			{
 				send_get_window_mask_data_response(message.get_source_control_function());
 				return true;
 			}
-			break;
```
ad3154, line 414: To be honest, I'd rather remove the individual returns and go to a single return at the end of the function.

isobus/src/isobus_virtual_terminal_server.cpp
sujandumaru: Good point. I’ll add bounds checks to _get_bit_ and _get_byte_, since both use variable shift counts.

isobus/src/isobus_virtual_terminal_server.cpp
sujandumaru: I kept the anonymous namespace only to give these helpers internal linkage. I can remove the anonymous namespace and mark the helper functions static instead, keeping the same file-local visibility while matching the project style better. I don’t think we want these helper functions to be externally visible in namespace `isobus`.

isobus/src/isobus_virtual_terminal_server.cpp
sujandumaru: Sure, that makes sense. I think this one is mostly a style preference. I’ve seen both approaches used: multiple returns for readability in some cases, and a single return at the end for consistency in others. Good to know your preference here. I’ll adjust this function and other as well to use a single return.

ad3154 (approved):

PR #701, opened 2026-07-19
Fix off-by-one bounds check in get_data_custom_length

94xhn:
While looking at CANMessage::get_data_custom_length I noticed the bounds check at the top of the function doesn't quite cover the valid range correctly:

```cpp
if (endBitIndex > 8 * data.size() || length < 1 || startBitIndex >= 8 * data.size())
```

Valid bit indices for an N-byte message only go from 0 to 8*N - 1, but this check only rejects `endBitIndex` when it's *greater than* `8 * data.size()`, so a value of exactly `8 * data.size()` slips through. That happens whenever `startBitIndex + length - 1` lands exactly on that boundary, which is easy to hit in practice, e.g. asking for the last byte of an 8-byte message starting one bit too far in: `get_data_custom_length(57, 8)` on an 8-byte message gives `endBitIndex = 64 = 8*8`. The check passes, and the loop later does `data.at(byteIndex)` with `byteIndex == data.size()`, so it throws `std::out_of_range` instead of returning 0 the way every other invalid range handled by this function does.

The fix is just changing that comparison to `>=` so this boundary is treated consistently with the rest of the invalid-range handling. I added two test cases to `can_message_tests.cpp`: one confirming the last valid byte still decodes correctly, and one for the exact boundary that used to throw.

### DIFF
--- a/isobus/src/can_message.cpp
+++ b/isobus/src/can_message.cpp
@@ -340,7 +340,7 @@ namespace isobus
 		std::uint32_t startAmountOfBytes = amountOfBytesLeft;
 		std::uint8_t indexOfFinalByteBit = 7;
 
-		if (endBitIndex > 8 * data.size() || length < 1 || startBitIndex >= 8 * data.size())
+		if (endBitIndex >= 8 * data.size() || length < 1 || startBitIndex >= 8 * data.size())
 		{
 			LOG_ERROR("End bit index is greater than length or startBitIndex is wrong or startBitIndex is greater than endBitIndex");
 			return retVal;
--- a/isobus/src/can_message.cpp
+++ b/isobus/src/can_message.cpp
@@ -342,7 +342,9 @@ namespace isobus
 
 		if (endBitIndex >= 8 * data.size() || length < 1 || startBitIndex >= 8 * data.size())
 		{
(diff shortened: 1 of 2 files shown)


isobus/src/can_message.cpp
```diff
@@ -340,7 +340,7 @@ namespace isobus
 		std::uint32_t startAmountOfBytes = amountOfBytesLeft;
 		std::uint8_t indexOfFinalByteBit = 7;
 
-		if (endBitIndex > 8 * data.size() || length < 1 || startBitIndex >= 8 * data.size())
+		if (endBitIndex >= 8 * data.size() || length < 1 || startBitIndex >= 8 * data.size())
 		{
 			LOG_ERROR("End bit index is greater than length or startBitIndex is wrong or startBitIndex is greater than endBitIndex");
```
sujandumaru: You can update the error log to include "equal to" as well.

94xhn: Updated the diagnostic to use the words greater than or equal to for both inclusive boundary checks, ran the repository clang-format check, and pushed commit 2e8e6e6. Thanks for the suggestion.

sujandumaru (approved):

ad3154 (approved):

PR #705, opened 2026-07-28
Update the length of the heartbeat message to be 8 and fill the remaining bytes with 0xFF

sam-brodie:
Update the length of the heartbeat message to be 8 and fill the remaining bytes with 0xFF. The Heartbeat message has a length of 8 https://www.isobus.net/isobus/pGNAndSPN/3299?type=PGN

## Describe your changes

Pad the heartbeat message with 0xFF to a length of 8.

## How has this been tested?

Using Vector CANoe to request the heartbeat, CANoe will flag that the lengths is incorrect.

### DIFF
--- a/isobus/src/isobus_heartbeat.cpp
+++ b/isobus/src/isobus_heartbeat.cpp
@@ -137,7 +137,8 @@ namespace isobus
 	bool HeartbeatInterface::Heartbeat::send(const HeartbeatInterface &parent)
 	{
 		bool retVal = false;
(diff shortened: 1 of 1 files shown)


ad3154:
This appears to be a very recent change to the definition of the heartbeat message - and for some reason, they didn't update the PG modify date in the online database. Very odd.

Anyways, thanks for this contribution. I will approve it if you update the tests, which are currently expecting a length of 1, so that just needs to be changed to 8.

sujandumaru:
Updated the test with correct results.

ad3154 (approved):

PR #708, opened 2026-08-18
Do not reparse the full (already parsed) IOP after new segments arrive

martonmiklos:
## Describe your changes

In the case an IOP segment is being transferred the whole pool is being reparsed.

This means in the case if some object properties (like object hide/show state) are being set by commands after the reparsing these changes will be lost. This goes against the what the standard says about the visibility effect:

kép

Fixes #Open-Agriculture/AgIsoVirtualTerminal/201

## How has this been tested?

Built AgIsoVT with this stack version and checked a side effect caused by the reparsing, which is gone after this fix.

### DIFF
--- a/isobus/include/isobus/isobus/isobus_virtual_terminal_working_set_base.hpp
+++ b/isobus/include/isobus/isobus/isobus_virtual_terminal_working_set_base.hpp
@@ -104,12 +104,24 @@ namespace isobus
 		                                  std::uint8_t *&iopData,
 		                                  std::uint32_t &iopLength) const;
 
+		/// @brief A class to hold together IOP data and a processing state flag
+		class IopDataComponent
+		{
+		public:
+			///< Constructor where the IOP data can be passed
+			/// @param[in] d A pointer to the object pool data
+			IopDataComponent(const std::vector<std::uint8_t> &d) :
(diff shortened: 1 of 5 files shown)


isobus/include/isobus/isobus/isobus_virtual_terminal_working_set_base.hpp
```diff
@@ -104,12 +104,24 @@ namespace isobus
...
 
+		/// @brief A class to hold together IOP data and a processing state flag
+		class IopDataComponent
+		{
+		public:
+			///< Constructor where the IOP data can be passed
+			/// @param[in] d A pointer to the object pool data
+			IopDataComponent(const std::vector<std::uint8_t> &d) :
+			  data(d) {}
+			bool processed = false; ///< This variable is set to true after this IOP section is being parset
```
sujandumaru, line 115: Typo: parsed :)

sujandumaru:
Is it possible to add a test for this?

martonmiklos:
> Is it possible to add a test for this?

I have added an unit test to cover this feature.

sujandumaru (approved):
Would be better if we can add the "explicit" on the constructor with one argument as suggested by SonarCloud.

PR #710, opened 2026-09-04
FIX: null-check source control function in TechnicalCapabilities handling

gunicsba:
process_rx_messages() dereferenced rxMessage.get_source_control_function() unguarded in the TechnicalCapabilities case (both directly via get_address() in LOG_INFO calls, and indirectly via get_active_client(...)->reportedVersion). Since messages are queued from a separate thread via store_rx_message() and the source control function can legitimately be null (e.g. address not yet resolved during address- claim churn), a burst of claim activity could queue a TechnicalCapabilities message with a null source and crash with an access violation - confirmed via a field minidump showing a null-pointer deref inside ControlFunction::get_address() called from process_rx_messages().

Also guard get_active_client()'s return in the ParameterVersion branch, which could independently be null if a client sends its version before being registered as active, mirroring the existing guard pattern used in the DeviceDescriptor case.

### DIFF
--- a/isobus/src/isobus_task_controller_server.cpp
+++ b/isobus/src/isobus_task_controller_server.cpp
@@ -265,7 +265,11 @@ namespace isobus
 					{
 						case ProcessDataCommands::TechnicalCapabilities:
 						{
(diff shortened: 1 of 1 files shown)


martonmiklos (approved):
LGTM

PR #713, opened 2026-09-04
Dry CAN interface opening in examples

martonmiklos:
## Describe your changes

All examples had a similar CAN interface opening logic snipplet. Adding new CAN drivers could be tedious this way. 
In some examples it changed the default can0 interface to vcan0 on Linux, but I am planning to add an uniformized CAN interface argument handling as a next step.

## How has this been tested?

Built examples and tested seeder example.

### DIFF
--- a/examples/common/create_can_driver.cpp
+++ b/examples/common/create_can_driver.cpp
@@ -1,0 +1,28 @@
+#include "create_can_driver.hpp"
+
+#include "isobus/hardware_integration/available_can_drivers.hpp"
+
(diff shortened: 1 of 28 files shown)


sujandumaru (approved):
Great code dedup 👍🏻

ad3154:
This PR did not properly update the documentation used in the example sphinx docs, like in `Virtual Terminal Basics.rst`. Please go back and fix that too @martonmiklos and keep in mind those docs pls

PR #714, opened 2026-09-04
Add .cmake extension to the .gitignore

martonmiklos:
## Describe your changes

QtCreator based builds producing .cmake artifacts. 

```
mm@P15:~/Projektek/digitroll/AgIsoStack-plus-plus$ git st
HEAD detached at origin/main
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        CPackConfig.cmake
        CPackSourceConfig.cmake
        build_vcanc/
        cmake_install.cmake
        hardware_integration/cmake_install.cmake
        isobus/cmake_install.cmake
        isobusConfig.cmake
        isobusConfigVersion.cmake
        utility/cmake_install.cmake
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,3 +65,4 @@ CMakeFiles
 *.orig
 *.rej
 
+*.cmake


ad3154 (approved):

sujandumaru (approved):
I see we have 1 `.cmake` file that is currently being tracked which is on cmake/ folder. Do you think we will have similar new `.cmake` file that might need tracking? Probably not a big deal for now. Just something to consider later.

martonmiklos:
> I see we have 1 `.cmake` file that is currently being tracked which is on cmake/ folder. Do you think we will have similar new `.cmake` file that might need tracking? Probably not a big deal for now. Just something to consider later.

Ah, it makes sense. I have submitted an another PR to fix this. (I will read comments properly before merge next time).

PR #716, opened 2026-09-05
Document saving a DDOP to a .iop file

sujandumaru:
## Describe your changes

The tutorial builds `OurDDOP` in memory but never says how to get it into a `.iop` file or open that file in AgIsoDDOPGenerator. Adds both steps:

- **Saving a DDOP to a .iop File**: `generate_binary_object_pool()` into a buffer, write it with `std::ofstream` in binary mode. Notes the file step is optional (a DDOP can go straight to `TaskControllerClient` in memory) and belongs in `main()`, not the pool-building function.
- A note on opening that file in AgIsoDDOPGenerator: its version picker defaults to **Version 3**, but a default-constructed pool is version 4. Wrong pick fails with *"There were errors loading the DDOP"*, which reads like corruption but isn't.

## How has this been tested?

Compiled the snippet against the built library; the file it writes reads back at TC version 4 only, matching the note.

### DIFF
--- a/sphinx/source/Tutorials/Task Controller Basics.rst
+++ b/sphinx/source/Tutorials/Task Controller Basics.rst
@@ -166,6 +166,31 @@ In theory, you could now provide this DDOP to the AgIsoStack TC client, and it w
 
 We'll cover use of the TC client in a later tutorial, but knowing how to create a DDOP is the first step to using it.
 
+Saving a DDOP to a .iop File
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The code above builds :code:`OurDDOP` as a tree of C++ objects in memory. It does not create a file. If you want a .iop file, you have to serialize that tree into bytes yourself, and write those bytes out in binary mode:
+
+.. code-block:: c++
+
+    #include <fstream>
+
+    std::vector<std::uint8_t> binaryPool;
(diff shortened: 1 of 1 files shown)


gunicsba:
As a somewhat outsider to ISOBUS. If we have IOP for graphics and DDOP for the Device Data shouldn't we call these files .iop and .ddop ?

Or is it only me who always used .ddop and .iop ?

sujandumaru:
This is just my POV:
There is no standard file extension for this. What's important is: ISO 11783 defines the **binary layout of the pool**, not the file name. So the same bytes can be saved as `.iop`, `.ddop` or even `.bin`, and they still load fine. It depends on who is writing it. AOG-TaskController saves it as `.ddop` for DDOP.

I picked `.iop` here because the next section tells the reader to open the file in AgIsoDDOPGenerator, and that tool currently saves as `device_descriptor_object_pool.iop` by default and ships its sample as `EXAMPLE.iop`. So this way the tutorial and the tool agree.

I can add one more sentence here saying the DDOP can be save in `.iop` and `.ddop` file format as per they desired to avoid the confusion.

gunicsba:
Ah yes you're right however if we ever plan to make default file type associations then distinguishing the iop and ddop would be helpful.

martonmiklos (approved):

PR #720, opened 2026-09-09
Uniformize CAN interface argument handling in the examples

martonmiklos:
## Describe your changes

I have added --interface and --driver command line argument for each example.

This simplifies adding and using new CAN interfaces in the examples.

## How has this been tested?

Built and ran examples with the new command line switches.


examples/common/can_arguments_parser.hpp
sujandumaru, line 1: missing `#pragma once` or `ifndef/define`

examples/common/can_arguments_parser.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+	const CanParameters &parameters() const
+	{
+		return args;
+	}
+
+private:
+	void parse(int argc, char *argv[])
+	{
+		for (int i = 1; i < argc; ++i)
```
sujandumaru, line 35: Do we need to be careful here? The loop index `i` being changed inside the for-loop as well while calling `argv[++i]`. Is that expected?

sphinx/source/Tutorials/Virtual Terminal Basics.rst
```diff
@@ -55,26 +55,13 @@ Create the file `main.cpp` as shown below inside that folder with the requisite
 	int main()
```
sujandumaru: need `argc` and `argv` as argument here

sphinx/source/Tutorials/Virtual Terminal Basics.rst
sujandumaru, line 44: Gotta include our new headers here as well.

sujandumaru: Need update here as well

examples/common/can_arguments_parser.hpp
martonmiklos: Yes this is expected if the loop iteration found a --driver or --interface argument then the next argument is going to be the value. The check above ensure that there is a next argument passed. The ++i portion will jump through the value passed to the argument.

sphinx/source/Tutorials/Virtual Terminal Basics.rst
martonmiklos: Ah nice catch! I need to take more care on the docs.

examples/common/create_can_driver.cpp
```diff
@@ -1,28 +1,125 @@
...
+	{
+#if defined(ISOBUS_TOUCAN_AVAILABLE)
+		canDriver = std::make_shared<isobus::TouCANPlugin>();
+#endif
+	}
+#elif defined(__APPLE__)
+	// OSX
+	if (driver.empty() || driver == "peak")
+	{
+		const int channel = interfaceName.empty() ? PCAN_USBBUS1 : (std::stoi(interfaceName) - 1 + PCAN_USBBUS1);
```
sujandumaru, line 57: Don't we need a guard for ISOBUS_MACCANPCAN_AVAILABLE here?

sujandumaru (approved):
I think all the parts here look good. But there are still some tutorial part that we might have to go through again as some of them are pointing to old version. Not directly but in a sense. For instance, ([L502-L503](https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/a54de664d135641c8a577279db7482d085820062/sphinx/source/Tutorials/Virtual%20Terminal%20Basics.rst?plain=1#L502-L503)) still talks about the list "above" which is not there anymore. This is minor. We can go through all tutorials later and confirmed they are upto date as a lot of code seems to be changing.

PR #723, opened 2026-09-11
Make the TWAIPlugin receive and transmit timeouts configurable

patrykkrz:
`TWAIPlugin::read_frame` and `TWAIPlugin::write_frame` both block for a hard-coded
100ms. That is a fine default, but applications that need a more responsive receive
thread, or that cannot afford to block the transmit path for that long, currently
have no way to change it without patching the library.

Both timeouts are now optional constructor parameters, defaulting to the new
`TWAIPlugin::DEFAULT_TIMEOUT_MS` (100). Existing call sites, including the tutorial
and the `esp32_platformio_object_pool` example, compile unchanged and keep the same
behaviour.

```cpp
// 20ms receive timeout, 50ms transmit timeout
auto canDriver = std::make_shared<isobus::TWAIPlugin>(&twaiConfig, &twaiTiming, &twaiFilter, 20, 50);
```


martonmiklos (approved):
LGTM!

Many thanks for your contribution @patrykkrz !

PR #726, opened 2026-09-12
Add ESP32 PlatformIO seeder example with section control

gunicsba:
Ports the desktop seeder example to the ESP32-S3 (Waveshare ESP32-S3-ETH-8DI-8RO-C) as a Virtual Terminal and Task Controller client with section control.

board_config.hpp centralizes all board wiring (CAN, digital inputs, relay I2C) plus HAS_RELAY_EXPANDER and HAS_DIGITAL_INPUTS flags, so the example can be retargeted to a board without the relay expander or opto inputs by editing one header. relay_driver drives 8 relays through a TCA9554 I2C expander and probes it at init, so section commands no-op when it is absent. input_driver reads 8 debounced opto-isolated inputs as row fault feedback, so a faulted row reports not working to the Task Controller and on the VT while its commanded relay stays energized.

Also add the example to the PlatformIO CI build matrix, and fix object pool embedding in the VT object pool example and its tutorial by using board_build.embed_files (raw binary) instead of embed_txtfiles, which appends a NUL terminator that makes the uploaded pool one byte too long so the VT rejects it as corrupt.


examples/seeder_example_esp32_platformio/src/board_config.hpp
```diff
@@ -0,0 +1,62 @@
...
+///           - the TWAI (CAN) controller pins
+///           - the opto-isolated digital input pins
+///           - the I2C pins and address of the relay GPIO expander
+///          The two feature flags let the example run on boards that lack some of this hardware.
+///          When a flag is false, main.cpp skips that hardware entirely: the pins are not claimed,
+///          the inputs are not polled, and the relays are not driven. The relay expander is ALSO
+///          probed at runtime, so even with HAS_RELAY_EXPANDER true, a missing or unpowered IC just
+///          makes relay control a no-op instead of faulting the application.
+///
+/// @copyright 2024 The Open-Agriculture Developers
```
martonmiklos: 2026

gunicsba:
Andras tried it on John Deere and it seems to work well:

https://github.com/user-attachments/assets/fd77aaa8-b251-4c87-986d-6c61099e3ca4

sphinx/source/Tutorials/ESP32 PlatformIO.rst
```diff
@@ -268,7 +268,7 @@ To build and run a minimal, but interactive project that will load an ISOBUS obj
...
-    Furthermore, in the :code:`platformio.ini` file, you should specify the file under :code:`board_build.embed_txtfiles` to embed the object pool into your binary, as shown in the example `here <https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/255fd580925e1d7d9baea1b16ad4ffcedf1fc974/examples/virtual_terminal/esp32_platformio_object_pool/platformio.ini#L16>`__.
+    Furthermore, in the :code:`platformio.ini` file, you should specify the file under :code:`board_build.embed_files` to embed the object pool into your binary, as shown in the example `here <https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/255fd580925e1d7d9baea1b16ad4ffcedf1fc974/examples/virtual_terminal/esp32_platformio_object_pool/platformio.ini#L16>`__. Use :code:`embed_files` (raw binary) rather than :code:`embed_txtfiles`: the latter appends a NUL terminator, making the uploaded object pool one byte too long, which the virtual terminal then rejects as corrupt so the pool never loads.
```
martonmiklos: If this is true than we shall modify the examples/virtual_terminal/esp32_platformio_object_pool/platformio.ini (preferably in a separate PR). I would rather remove the mentioning of the embed_txtfiles and would just use the correct terms.

gunicsba: True, Lets make this a stacked PR then!
BugFix before the Feature, right?

gunicsba: Split out as #742 (requested your review there). Trimmed the fix out of this PR so #726 is feature-only now.

PR #581, opened 2025-05-21
[VT]Add infrastructure for supporting Screen Capture commands

martonmiklos:
## Describe your changes

* Added callback to the VirtualTerminalServer to being able to receive Screen capture command
* Added helper function to send Screen capture response

## How has this been tested?

Built AgIsoVT from this branch:
https://github.com/martonmiklos/AgIsoVirtualTerminal/tree/add_support_for_screen_capture
which leverages these changes and sent screen capture commands with:

```
cansend vcan0 1CE72687#C60001FFFFFFFFFF
```

PR #727, opened 2026-09-12
Add clang-tidy checks for the style rules in CONTRIBUTING.md

gunicsba:
## Why

Many review comments are about the same style rules, most of which are already listed in `CONTRIBUTING.md`. I went through the review comments on this repository to find the ones that come up repeatedly:

- Constants on the left side of `==` / `!=` (#98, #618)
- Explicit `std::` for types and functions: `std::uint8_t`, `std::size_t`, `std::find`, `std::min` (#82, #91, #235, #344, #451, #581, #672)
- Parentheses around comparisons in `&&` / `||` conditions (#91, #198)
- `snake_case` functions and `camelCase` variables (#82, #195)
- `const` member functions, `explicit` constructors, `virtual` / `override`, `std::make_shared`, redundant casts (#82, #91, #144, #254, #303, #427)
- `LOG_*` macros instead of calling `CANStackLogger` directly (#562)
- No `using namespace` (#144, #198)

clang-format cannot enforce any of these, because it only changes formatting. clang-tidy can.

## What this adds

| Rule | Check |
|---|---|
| Constant on the left of `==` / `!=` | `custom-constant-on-left-comparison` |
| `std::` qualified fixed width integers and `size_t` | `custom-std-qualified-type` |
| `std::` qualified standard library calls (e.g. `find` found through ADL) | `custom-std-qualified-function-call` |
| Parentheses around comparisons in `&&` / `||` | `custom-parenthesize-sub-conditions` |
| No `using namespace` in headers | `custom-no-using-namespace-in-headers` |
| `LOG_*` macros instead of `CANStackLogger::warn/error/...` | `custom-use-logging-macros` |
| Naming (`snake_case` functions, `camelCase` variables, `CAPITALIZED_SNAKE` global/static constants) | `readability-identifier-naming` |
| `nullptr`, `explicit`, `override`, virtual destructors, `make_shared`, `const` methods, redundant casts | built-in checks |
| Only the basic source character set | non-ASCII check on the changed lines |

Files:

- **`.clang-tidy`**: built-in checks only. clang-tidy 21 and older reject the whole file if it contains the `CustomChecks` key, so keeping them out means editors and IDE plugins can still use this file.
- **`scripts/clang_tidy/custom_checks.yaml`**: the custom checks above. They are written as clang-query matchers, which requires clang-tidy 22 or newer with `--experimental-custom-checks`.
- **`scripts/clang_tidy/run_clang_tidy.sh`**:
  - Appends the custom checks to `.clang-tidy`.
  - By default, checks only the lines changed since a base ref (including uncommitted changes); `--all` checks everything.
  - Also checks the changed lines for non-ASCII characters.
  - Works on Linux and on Windows with Git Bash.
- **New `clang-tidy style` job in the linting workflow**:
  - Installs `clang-tidy==22.1.8` from pip.
  - Generates a compilation database for the SocketCAN build.
  - Runs the script against the PR base branch, treating warnings as errors.
- **`docs/pre-commit-hook.sh`**: also rejects added lines with non-ASCII characters.
- **`CONTRIBUTING.md`**: describes how to run the check locally, and adds the parentheses rule, which reviewers ask for but was not written down.
- **`ntcan_plugin.cpp`**: uses `LOG_ERROR` instead of calling `CANStackLogger::error` directly (6 places).

## Only changed lines are checked

Running everything on the current `main` gives these findings (library / tests):

| Check | Library | Tests |
|---|---|---|
| constant-on-left-comparison | 114 | 65 |
| readability-identifier-naming | 60 | 60 |
| parenthesize-sub-conditions | 44 | 8 |
| std-qualified-function-call | 7 | 0 |
| Other built-in checks | 12 | 4 |
| std-qualified-type | 1 | 0 |

Fixing all of these in one PR would conflict with most open PRs. So CI only checks the lines a PR changes. Existing code can be cleaned up in smaller follow-up PRs, and the job can be switched to `--all` once it is clean.

## How it was tested

- **Custom checks**: ran each one against sample files with known violations and non-violations. `std::uint8_t` vs `uint8_t`, `nullptr` comparisons through `shared_ptr`'s overloaded operator, runtime `const` locals vs `constexpr` constants, and `LOG_*` macros vs direct calls are all handled.
- **Whole codebase**: 86 source files, no false positives found in the reviewed findings.
- **Diff mode**: a planted violation in a test file was the only thing reported, and the script exited with 1. Unchanged violations in the same file were not reported.
- **Older clang-tidy**: clang-tidy 21 accepts `.clang-tidy` without errors.
- **Not yet tested in CI**: the Linux CI job itself. This PR only changes one C++ file that is not part of the SocketCAN build, so its own clang-tidy job has nothing to check.

## Limitations

- Header files are not compiled on their own, so a changed header line is only checked when a changed source file includes it.
- Platform-specific plugins that are not part of the SocketCAN build (NTCAN, MacCAN, SIL Kit, ESP32, Teensy) are not checked.
- Custom checks are an experimental clang-tidy feature, so the version is pinned.

## Open questions for maintainers

1. Should the parentheses rule be enforced? Reviewers ask for it, but it was not in `CONTRIBUTING.md` until this PR.
2. Naming currently allows leading acronyms (`CANPort`) and unit suffixes (`timestamp_ms`, `speed_mm_s`), because the codebase uses both a lot. Local `constexpr` variables are not required to be `CAPITALIZED_SNAKE`, because most existing code writes them in `camelCase`. Does that match your preference?
3. Are there other review nits you would like to see added as checks?

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,36 @@
+---
+# clang-tidy configuration for the style rules described in CONTRIBUTING.md.
+# Only built-in checks are configured here, so editors and older clang-tidy versions can use this file.
+# The project specific checks (constant on the left side of comparisons, explicit std:: qualification, ...)
+# are defined in scripts/clang_tidy/custom_checks.yaml. Use scripts/clang_tidy/run_clang_tidy.sh to run both.
+Checks: >
+  -*,
+  custom-*,
+  cppcoreguidelines-virtual-class-destructor,
+  google-explicit-constructor,
+  modernize-make-shared,
+  modernize-use-nullptr,
+  modernize-use-override,
+  readability-identifier-naming,
+  readability-make-member-function-const,
+  readability-redundant-casting
+WarningsAsErrors: ''
+HeaderFilterRegex: '((isobus|hardware_integration|utility)/include/isobus|test)/.*\.hpp$'
+CheckOptions:
+  readability-identifier-naming.NamespaceCase: lower_case
+  readability-identifier-naming.ClassCase: CamelCase
+  readability-identifier-naming.StructCase: CamelCase
+  readability-identifier-naming.EnumCase: CamelCase
+  readability-identifier-naming.EnumConstantCase: CamelCase
+  readability-identifier-naming.FunctionCase: lower_case
+  readability-identifier-naming.MethodCase: lower_case
+  # camelCase variables, allowing leading acronyms (CANPort, NAMEFilters) and unit suffixes (timestamp_ms, speed_mm_s)
+  readability-identifier-naming.VariableCase: camelBack
+  readability-identifier-naming.VariableIgnoredRegexp: '^([A-Z]{2,}[A-Za-z0-9]*|[a-z][A-Za-z0-9]*(_(us|ms|s|seconds|minutes|hours|mm|cm|m|km|mm_s|m_s|km_h|deg|rad|percent|hz|bytes|bits))+)$'
+  readability-identifier-naming.ParameterCase: camelBack
+  readability-identifier-naming.ParameterIgnoredRegexp: '^([A-Z]{2,}[A-Za-z0-9]*|[a-z][A-Za-z0-9]*(_(us|ms|s|seconds|minutes|hours|mm|cm|m|km|mm_s|m_s|km_h|deg|rad|percent|hz|bytes|bits))+)$'
+  readability-identifier-naming.MemberCase: camelBack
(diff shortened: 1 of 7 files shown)


gunicsba:
@ad3154 maybe you're the best to think about this.  @martonmiklos mentioned that it would be great to enforce these rules as well. I asked Claude to do a sketch.

PR #728, opened 2026-09-16
Fix heartbeat rx sequence counter validation

sujandumaru:
Three fixes in the heartbeat receive path:

- Expected counter was `(sequenceCounter + 1) % 250`. The valid range is 0-250 and the initial value 251 is followed by 0, so both 250 -> 0 and 251 -> 0 raised `InvalidSequenceCounter` against a conformant sender.
- The received counter was stored into `trackedHeartbeats.back()` instead of the matched entry, so with more than one CF on the bus the wrong entry was updated.

Four rx tests cover init -> 0, the 250 -> 0 wrap, per-CF counters, and that a genuinely bad counter still raises.


sujandumaru:
Source: https://www.isobus.net/isobus/pGNAndSPN/3299?type=SPN

ad3154 (approved):

PR #729, opened 2026-09-17
Fix VT server working-set connection management

gunicsba:
## Summary

Per ISO 11783-6:2014 §4.6.9, fixes four related bugs in the VT server's working-set connection management (`isobus_virtual_terminal_server.cpp`/`.hpp`):

- The client maintenance timeout was only refreshed by the Working Set Maintenance message. Other ECU→VT messages (Get Memory, Get Versions, an Object Pool Transfer, etc.) didn't count, so a client that was clearly still alive could be timed out anyway.
- There was no way to recognize that a timed-out client had genuinely reconnected. Reconnection is now recognized either by a new Working Set Master (init) message, or by the same NAME reappearing under a different `ControlFunction` instance (e.g. after an address-claim re-processes it) — logged at info level.
- A working set that has timed out and hasn't re-announced itself is no longer treated as managed. This means later messages from it — including an Object Pool Transfer — are rejected outright instead of being handed to `process_connection_dependent_messages()` as if the working set were still valid (previously observed reassembling a multi-hundred-kilobyte pool that was ultimately discarded).
- `send_acknowledgement()` built the NACK's destination-address byte correctly but sent the CAN frame itself with a null destination, which the network manager resolves to the global address (0xFF). The NACK for a stale/unmanaged working set is now sent directly to the working set master, as §4.6.9 requires.

## Test plan

- Added `test/vt_server_tests.cpp` covering:
  - the maintenance timeout being refreshed by a non-maintenance message
  - a timed-out client being rejected until it re-announces
  - a timed-out client being re-accepted when the same NAME reappears under a new `ControlFunction` instance
  - the timeout NACK being sent to the working set master's address rather than broadcast
  - an Object Pool Transfer from an unregistered/timed-out client not being processed
- Full existing suite (`unit_tests`) still passes; no regressions.

Part of Open-Agriculture/AgIsoVirtualTerminal#223 (the VT server working-set connection management section of that report). The AgIsoVT application gaps and AOG-TaskController issues from that report are tracked separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gunicsba:
Closing to split into smaller, independently reviewable PRs per maintainer guidance. Replacement PRs coming shortly, each linked back to Open-Agriculture/AgIsoVirtualTerminal#223.

PR #733, opened 2026-09-18
Fix format_value() for DDI 505 and 506

sujandumaru:
## Describe your changes

`format_value()` treated DDI 505 and 506 as the same three-value enumeration. Per the ISO 11783-11 DDI database,

- https://www.isobus.net/isobus/dDEntity/1074: 505 (Supported Track Control Levels) is a bitmask of the levels the implement supports
- https://www.isobus.net/isobus/dDEntity/1075: 506 (Setpoint Track Control Level) is an enumeration where `0` means no common level.

| DDI | Value | Before | After |
|---|---|---|---|
| 505 | `0` | Level 1 | No levels supported |
| 505 | `1` | Level 2 | Level 1 |
| 505 | `3` | Reserved | Level 1, Level 2 |
| 505 | `7` | Reserved | Level 1, Level 2, Level 3 |
| 506 | `0` | Level 1 | No common level |
| 506 | `3` | Reserved | Level 3 |

`format_value()` is only used in the TC server's log output, so nothing on the bus changes.

> Note: ISO has also renamed these from Tramline to Track Control, along with 406/407, 507-511 and 515. Should I rename the enumerators to match, keeping the old names as aliases so existing code still builds?

## How has this been tested?

```bash
cmake --build build
ctest --test-dir build -R DATA_DICTIONARY_TESTS --output-on-failure


ad3154:
This file gets code-generated by `scripts/code_gen_ddi_database.py` so it's likely to get overwritten in the future. Either people will need to be very careful checking the diff when running the script in the future, or we might need to handle it in the script?

sujandumaru:
This whole file is not generated by the script. Only the `DDI_ENTRIES` on line 982 is auto-generated here, so only those gets overwritten through the script and it replaces three things: the generated date in the header, the size of DDI_ENTRIES and the array itself. This function format_value() sits above the array and it's hand-written.

I ran the script for #715 and it didn't change anything outside the array (you can check that for reference). So the 505/506 cases here should be safe for the next regeneration too.

I agree the file header is a bit misleading though, since it says "This file was generated". Should I add a line in the @brief saying only DDI_ENTRIES comes from scripts/code_gen_ddi_database.py, so nobody gets confused later?

ad3154:
Ah, I see. Even I forgot and I think I wrote it, haha, A comment might be good

sujandumaru:
Sure. I can add that. My other major concern is the **Note:** section in the PR. As you can see, the enum name currently does not match the ISO 11783 database (Current: Supported Track Control Levels, Old: Tramline Control Level) https://github.com/Open-Agriculture/AgIsoStack-plus-plus/blob/460d74396ef2e321a114a5a28a9bb13351298f6e/isobus/include/isobus/isobus/isobus_standard_data_description_indices.hpp#L524 and it is **public** API. So, I am not sure if I should change that or not to match current database. My personal preference would be to change it. What do you think about that?

ad3154:
My preference would be to match it to the database, so even if it's public I'd change it.

sujandumaru:
Sounds good, will do that in next PR as there are multiple enums that needs changing.

ad3154 (approved):

PR #740, opened 2026-09-22
Save the tutorial DDOP as .ddop, leaving .iop to VT object pools

sujandumaru:
Follow-up to #716 as mentioned by @gunicsba. AgIsoDDOPGenerator now saves as `.ddop` by default (Open-Agriculture/AgIsoDDOPGenerator#32), so the Task Controller tutorial shouldn't still tell people to write `OurDDOP.iop`.

- Renames the section and the example filename to `.ddop`.
- Adds a sentence noting the extension is only a naming convention — ISO 11783 defines the binary layout, not the filename — and that the generator opens either one, so nobody thinks existing `.iop` DDOPs stopped working.


gunicsba (approved):
Look good to me!

